PR #60, opened 2012-02-24
Patch command + Configuration class should not continually create schema

tomdmaguire:
1) Patch command added to console:

Implemented PatchCommand to execute any missing migrations up to the current version. This has been useful for us when needing to merge branches an application that uses doctrine

2) Do not continually create schema after migration table has been created to avoid long running migrate scripts:

migrationTableCreated flag needs to be set in Configuration class after we have checked that the migration table exists. This will that ensure a new schema is not created for each version that we migrate to thereafter.


mikeSimonson:
@tomdmaguire As far as I can see both feature that you propose in this PR is the way the current doctrine migration function right ?

Are there other stuff in this PR that need to be pulled in or can it be closed ?

tomdmaguire:
Yeah, I believe this got implemented

PR #67, opened 2012-04-19
Allow overriding migration template

kbond:


mikeSimonson:
@kbond Can you explain why you want to extend the generateCommand instead of the AbstractCommand ?

Thanks

PR #100, opened 2012-11-12
Migration description

jkrcma:
In our company we encountered the need to give single migration a name. Unfortunately current implementation is heavily broken, because it allows you to change the version identifier `YYYYMMDDHHMMSS`, which is imho very dangerous.

Now the description is just an extra information, which is stored only in `getDescription()` method and displayed by status command with `--show-versions` argument. Method `getName()` was removed - **BC break!** Also removed the test method, simple getters are useless to test.

Usage


mikeSimonson:
@TomasVotruba Thanks it's merged

TomasVotruba:
@mikeSimonson :+1:

PR #101, opened 2012-11-24
Ignore specified tables

samuelwilliams:
A developer, for whatever reason, may want to have tables that do not correspond to Doctrine entities. With this modification, developers are able to define in the config file which tables can be ignored. 

There are no backward compatibility issues.


mikeSimonson:
@peterrehm Thanks for the documentation

PR #102, opened 2012-11-26
Allow editor-cmd to work with terminal-based editors

arnaud-lb:
This allows `--editor-cmd=vim` to work correctly in generate commands.

shell_exec captures the output, as a result a terminal-based editor cannot display anything on the terminal.

proc_open, when called with an empty $descriptorspec argument, lets the executed command inherit its file descriptors; as a result the editor can display itself in the terminal.


arnaud-lb:
Awesome, thanks @mikeSimonson

mikeSimonson:
@arnaud-lb Thanks for your PR

PR #97, opened 2012-10-18
Simple way to generate PHAR build using the Box Project

michaelperrin:
I changed the way to build a standalone PHAR archive for Doctrine Migrations using the Box Project (http://box-project.org/).

This helps to remove the unnecessary (and currently broken) scripts to generate a phar archive.

Documentation for building the archive has been updated.

PR #111, opened 2013-02-26
package.php now succeeds in building the package (with deps installed via composer)

djfm:
It seems composer now installs stuff in directories different to those used before... I couldn't build the phar using the provided package.php file so I tweaked it and now it successfully builds!


mikeSimonson:
close in favor of #97

PR #131, opened 2013-09-17
fixed phpcs issues in migration files

acrobat:
Fixed missing docblocks before up() and down() functions in the migrations scripts. This caused php codesniffer to give coding error on these files.

PHPCS error:

| ERROR   | Missing function doc comment


mikeSimonson:
Thanks @acrobat

PR #145, opened 2013-11-27
Adjusted code style & do not create empty migration

66Ton99:


mikeSimonson:
@66Ton99 
Doing blind replacement in an sql string is asking for trouble.

I would not try to mess with sql like that as there would be no way to be sure it will not break the query.

PR #150, opened 2014-01-02
Returns empty array when processing 0 of 0 instead of exception.

Incognito:
This is consistent with other application behaviour.

Processing "0 of 1" and "1 of 1" migrations throws no exception,
but simply passes on an array(). "0 of 0" would turn out an exception
code which makes the script return an exit code to the system.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Migration.php
+++ b/lib/Doctrine/DBAL/Migrations/Migration.php
@@ -131,7 +131,7 @@ class Migration
         $direction = $from > $to ? 'down' : 'up';
         $migrationsToExecute = $this->configuration->getMigrationsToExecute($direction, $to);
 
-        if ($from === $to && empty($migrationsToExecute) && $migrations) {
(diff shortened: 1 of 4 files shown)


mikeSimonson:
@Ocramius What do you think about the exit code ?

Ocramius:
I think it is actually better to just leave the exception in place.

From an automation perspective, running an empty set of migrations seems to be a mistake and something that the user would know about. Thoughts?

mikeSimonson:
Well I guessing that if you are automating the deployment crashing because there were no migration to process makes the deploy code a little bit more complicated.
But I like that the user is warned that no migration were found.
Maybe we can change the exit code but then also show a message like warning the user, like those that warn that no sql query were run processing a migration file.

Ocramius:
@mikeSimonson or add a flag `--allow-no-migration-diff`?

mikeSimonson:
A similar option is added by #350

PR #152, opened 2014-01-04
Type mapping through configuration

pezia:
Added type mappings to the configuration options based on dyadyavasyas work.


mikeSimonson:
@pezia Type mapping should be configured trough the connection passed to doctrine migrations not in doctrine migrations itself.

PR #154, opened 2014-01-14
Change prioritise --db-configuration option in AbstractCommand.php

acclouis:
Change prioritise of db-configuration option, --db-configuration option is checked first.


mikeSimonson:
@acclouis Can you rebase your change so that it's in one commit ?

Thanks

mikeSimonson:
Thanks @acclouis

PR #159, opened 2014-03-08
Support version aliases in migrations:migrate

c960657:
Sometimes you just want to rerun the migration you are working on or run the migrations one step it a time. This involves a lot of copy-pasting of version numbers.

I suggest we allow commands like this:
* console migrations:migrate previous
* console migrations:migrate next
* console migrations:migrate first

I have introduced "first" as an alias of "0". I think this is more user-friendly (I remember that it took me a while to figure out that I the name of the initial version is "0").

I would like some comments on the naming of the aliases. Other solutions include "initial"/"prev" instead of or in addition to "first"/"previous".


stof:
Thanks

PR #166, opened 2014-04-09
Add shebang to files, so package.php and phar can be executed directly

colinfrei:


mikeSimonson:
close in favor of #97 ?

mikeSimonson:
@colinfrei Thanks but we are using box now to build the phar

PR #167, opened 2014-04-09
Allow overriding the password from the command line

colinfrei:
We have a special database user for migrations, as the regular user doesn't have permission to make schema changes. We're not able to have that users password saved in a file, so with this we can override it from the command line.


mikeSimonson:
@colinfrei Can you rebase your PR and squash the commits together ?

mikeSimonson:
I am going to side with @stof here. It is possible to do something similar with a php config file that retrieve the password from the environment.

PR #168, opened 2014-04-24
Update migration_versions table when generating SQL

rjmunro:
This means if you run the SQL generated in full, your migration_versions
table is consistent with your database, and migrations will continue
to work in future.


mikeSimonson:
@rjmunro Can you add an option to not add those queries in the generated file ?

mikeSimonson:
@rjmunro Thanks

Issue #173, opened 2014-07-04
[DX] Help the user know how to test a migration file


weaverryan:
Hi guys!

One of the trickiest parts of using migrations are testing them. That's a tough workflow (and sometimes you'll just need to pull a production schema dump to test things).

After running a `generation` or `diff`, what about adding a message to tell the user how they could run this migration independently locally to test it:

> To run just this migration for testing purposes, you can use `migrations:diff --up 2014XXXXXXX`

Thoughts in general? Overall, I want to help people into the right migrations workflow - it's tough.

allan-simon:
what is the status of this ticket ? As I see it's the only one left for the 1.0 milestone and there's already a PR opened on this

lcobucci:
I've remove this from `v1.0` because it was already released without this feature. @mikeSimonson what should we do regarding this?

mikeSimonson:
We should try to improve the documentation in general but also specifically regarding this.

PR #174, opened 2014-07-05
Help the user know how to test a migration file

agallou:
Added a message to explain how to test the migration at the end of the
diff and generate commands :

```
To run just this migration for testing purposes, you can use migrations:execute --up 2014XXXXXXX
```

See #173

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
@@ -111,6 +111,11 @@ EOT
         $path = $this->generateMigration($configuration, $input, $version, $up, $down);
 
         $output->writeln(sprintf('Generated new migration class to "<info>%s</info>" from schema differences.', $path));
+        $commandName = sprintf('%s:execute', $this->getApplication()->extractNamespace($this->getName()));
+        if (count($command = $this->getApplication()->find($commandName))) {
+          $output->writeln(sprintf('To run just this migration for testing purposes, you can use "<info>%s --up %s</info>"', $command->getName(), $version));
+        }
+
     }
 
     private function buildCodeFromSql(Configuration $configuration, array $sql)
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
@@ -93,6 +93,11 @@ EOT
         $path = $this->generateMigration($configuration, $input, $version);
 
         $output->writeln(sprintf('Generated new migration class to "<info>%s</info>"', $path));
+
+        $commandName = sprintf('%s:execute', $this->getApplication()->extractNamespace($this->getName()));
+        if (count($command = $this->getApplication()->find($commandName))) {
+          $output->writeln(sprintf('To run just this migration for testing purposes, you can use "<info>%s --up %s</info>"', $command->getName(), $version));
+        }
     }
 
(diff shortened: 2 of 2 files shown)


mikeSimonson:
@agallou I don't really like the test word. It's not really going to test it, it's going to run the migration.
I think the wording need a bit of work but the idea is good.

chrisguitarguy:
I think it might be better to just add the `--dry-run` flag to the example command -- that will actually mean that it's a test.

allan-simon:
`to run this migration in isolation for checking purposes ....` ?

stof:
@chrisguitarguy using the ``--dry-run`` flag will not let you test whether the migration can be executed on your database or if you have a bug in your SQL queries.

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
```diff
@@ -111,6 +111,11 @@ public function execute(InputInterface $input, OutputInterface $output)
         $path = $this->generateMigration($configuration, $input, $version, $up, $down);
 
         $output->writeln(sprintf('Generated new migration class to "<info>%s</info>" from schema differences.', $path));
+        $commandName = sprintf('%s:execute', $this->getApplication()->extractNamespace($this->getName()));
+        if (count($command = $this->getApplication()->find($commandName))) {
+          $output->writeln(sprintf('To run just this migration for testing purposes, you can use "<info>%s --up %s</info>"', $command->getName(), $version));
+        }
```
weaverryan: > To execute just this migration to make sure it works, you can use...

Also, the `"` at the end probably aren't needed - the `<info>` should be enough to make it stand out.

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
```diff
@@ -111,6 +111,11 @@ public function execute(InputInterface $input, OutputInterface $output)
         $path = $this->generateMigration($configuration, $input, $version, $up, $down);
 
         $output->writeln(sprintf('Generated new migration class to "<info>%s</info>" from schema differences.', $path));
+        $commandName = sprintf('%s:execute', $this->getApplication()->extractNamespace($this->getName()));
+        if (count($command = $this->getApplication()->find($commandName))) {
+          $output->writeln(sprintf('To run just this migration for testing purposes, you can use "<info>%s --up %s</info>"', $command->getName(), $version));
```
jwage: Can you change the indention here to 4 spaces instead of 2.

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
```diff
@@ -93,6 +93,11 @@ public function execute(InputInterface $input, OutputInterface $output)
         $path = $this->generateMigration($configuration, $input, $version);
 
         $output->writeln(sprintf('Generated new migration class to "<info>%s</info>"', $path));
+
+        $commandName = sprintf('%s:execute', $this->getApplication()->extractNamespace($this->getName()));
+        if (count($command = $this->getApplication()->find($commandName))) {
+          $output->writeln(sprintf('To run just this migration for testing purposes, you can use "<info>%s --up %s</info>"', $command->getName(), $version));
```
jwage: Can you change the indention here to 4 spaces instead of 2.

localheinz: Also see https://github.com/doctrine/migrations/pull/174/files#r60681298.

jwage:
This was implemented in master (2.0). Thanks!

PR #175, opened 2014-07-05
Support for non-transactional migrations.

klandaika:
Some DB engines do not support ALTER statements inside a transaction. Added ability to disable transaction on individual migration level by overriding  the isTransactional() function.

EDIT 12/6/2014 (in reply to rjmunro):
One database which experiences this issue that I ran across is Pervasive SQL: http://www.pervasive.com/database

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/AbstractMigration.php
+++ b/lib/Doctrine/DBAL/Migrations/AbstractMigration.php
@@ -86,6 +86,19 @@ abstract class AbstractMigration
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mikeSimonson:
@klandaika Do you still count to make those change in your PR ?

klandaika:
@mikeSimonson Yes, just committed the changes.

mikeSimonson:
@klandaika Thanks

mikeSimonson:
It's merged.
Thanks

PR #188, opened 2014-10-01
Add --ignore-errors option to migrations:execute

rjmunro:
If a migration fails with an error part way through, even if you fix
the error in the migration code, you usually cannot reapply the
migration because it will fail trying to redo the original commands that
succeeded.

If you try to roll back, that will fail on any commands that undo things
that weren't successfully done the first time.

This adds an --ignore-errors option to the migrations:execute command
that will run all the commands in the migration, even after earlier
commands cause errors. It is not recommended to use this option unless
you fully understand it's implications, but it can save a lot of time
fixing a broken database migration.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Version.php
+++ b/lib/Doctrine/DBAL/Migrations/Version.php
@@ -414,7 +414,7 @@ class Version
                         }
                     }  catch (DBALException $e) {
                         if ($continueAfterErrors) {
-                            $this->outputWriter->write(sprintf('         <error>- failed</error>', $query));
+                            $this->outputWriter->write(sprintf('         <error>- %s</error>', $e->getMessage()));
                         } else {
                             // Reraise exception
                             throw $e;
--- a/lib/Doctrine/DBAL/Migrations/Version.php
+++ b/lib/Doctrine/DBAL/Migrations/Version.php
@@ -19,11 +19,12 @@
 
 namespace Doctrine\DBAL\Migrations;
 
-use Doctrine\DBAL\Types\Type;
+use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Migrations\Configuration\Configuration;
 use Doctrine\DBAL\Migrations\Provider\LazySchemaDiffProvider;
 use Doctrine\DBAL\Migrations\Provider\SchemaDiffProvider;
 use Doctrine\DBAL\Migrations\Provider\SchemaDiffProviderInterface;
+use Doctrine\DBAL\Types\Type;
 
 /**
  * Class which wraps a migration version and allows execution of the
@@ -260,7 +261,7 @@ class Version
      *
      * @throws \Exception when migration fails
      */
-    public function execute($direction, $dryRun = false, $timeAllQueries = false)
+    public function execute($direction, $dryRun = false, $timeAllQueries = false, $continueAfterErrors = false)
     {
         $this->sql = [];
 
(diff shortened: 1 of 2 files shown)


mikeSimonson:
@rjmunro Not sure about that one.

It's convenient that's for sure.
What do you think about it @Ocramius ?

stof:
The first commit is totally unrelated to this PR and should not be part of it

mikeSimonson:
@rjmunro You will need to put that first commit in an other PR if you want this one accepted.

@stof What do you think about the feature ?

rjmunro:
I've put the first commit in https://github.com/doctrine/migrations/pull/205. Once that's merged, I can rebase this and we can look at it again.

stof:
@rjmunro can you rebase ? The other PR has been merged

rjmunro:
@stof: Rebased now.

lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -256,12 +257,21 @@ public function execute($direction, $dryRun = false)
...
+                            if ( ! isset($this->params[$key])) {
+                                $this->outputWriter->write('     <comment>-></comment> ' . $query);
+                                $this->connection->executeQuery($query);
+                            } else {
+                                $this->outputWriter->write(sprintf('    <comment>-</comment> %s (with parameters)', $query));
+                                $this->connection->executeQuery($query, $this->params[$key], $this->types[$key]);
+                            }
+                        } catch (DBALException $e) {
+                            if ($continueAfterErrors) {
+                                $this->outputWriter->write(sprintf('         <error>- failed</error>', $query));
```
stof: the output message should include the exception message.

and your sprintf with the SQL is useless because there is not placeholder in your string (and there is no reason to put the SQL again here because it has already been displayed before running the query)

dextervip:
+1 It would be great to fix database easily.

titolendable:
This is handy to have in place, any plans to merge? 
In my use case, we want to insert some data as part of the migration, since it crucial to go live at the same time as the migration, but data is only valid for production, in all the other environments it will fail due to foreign keys constraints.

samsonasik:
@rjmunro need rebase again

rjmunro:
@samsonasik @stof I've rebased this again. Can it be merged?

Ocramius:
@mikeSimonson just read through this, and I can't figure out a scenario where this makes sense. Do we have any real-world scenarios where it is actually useful?

A few other things are also missing:

 * tests
 * wrapping "retry" code in at least a private method

mikeSimonson:
@Ocramius I can imagine some scenarios were it could make sense but I am not sure that the added complexity is worth it.

And I am not merging something that touch that part of the code without tests.

samsonasik:
@Ocramius I've sample for failure alter table for MySQL if column exists https://gist.github.com/samsonasik/6272d635836832809321e8f103c3a2ed 

In real life, there are devs that just messing with DB directly :(

Ocramius:
> In real life, there are devs that just messing with DB directly :(

The fact that someone fucks up gloriously doesn't make the problem part of the toolchain, IMO.

luispabon:
This is a consequence of the --dry-run command not actually doing anything useful. For instance, you'll get these errors when applying a migration to a database where the schema updater has already been run. Dry run doesn't run anything on the database so even if it's successful, the migration execution will fail later on.

I'd say effort will be better spent trying to improve --dry-run to a point where it can be used to do safe, full automatic database migrations on deployment.

rjmunro:
@luispabon My use case was a consequence of migrations not happening in transactions. Where a migration fails part way through, the database is left in a half-migrated state. Running the migration again causes it to error at the first statement. Reverting doesn't work because that just tries to revert the previous migration, not the one that has just failed.

Currently the easiest solution is to fix the problem line in the migration, comment out all the lines before it, re-run the migration, then uncomment the lines before you commit the fixed migration.

luispabon:
Aye, doing the same thing here on my case as well. To be fair the case I put forward is a consequence of devs being naughty and not using migrations to move the db forward, but I do believe dry-run needs beefing up to catch accidental schema updates that break migrations.

jwage:
I think I am :-1: on this change too unless someone can make a strong case for it that makes sense. I don't quite understand the points that have been made around `--dry-run` not behaving correctly. If someone wants to reopen and make a case for this, please do.

PR #192, opened 2014-12-13
Switch cli-stub from using deprecated DialogHelper to QuestionHelper

ackimwilliams:
Symfony 2.5 deprecates DialogHelper , use \Symfony\Component\Console\Helper\QuestionHelper instead

### DIFF
--- a/phar-cli-stub.php
+++ b/phar-cli-stub.php
@@ -52,7 +52,7 @@ if (file_exists($configFile)) {
 }
 
 $helperSet = ($helperSet) ?: new \Symfony\Component\Console\Helper\HelperSet();
-$helperSet->set(new \Symfony\Component\Console\Helper\DialogHelper(), 'dialog');
+$helperSet->set(new \Symfony\Component\Console\Helper\QuestionHelper, 'question');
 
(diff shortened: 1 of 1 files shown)


mikeSimonson:
need to be merged just before #97

stof:
@mikeSimonson this alone is not enough (the code and the initialization should be changed in the same PR).
and this change would force to drop the support for the Symfony LTS, which looks wrong to me

mikeSimonson:
I put this PR on hold until the symfony LTS  (2.3 right ?) end of life then.
Around May of 2016 ?

stof:
@mikeSimonson it is possible to write code compatible with both 2.3 and new APIs, but it requires a little bit more work than replacing the old way with the new one.
and in any case, this PR is broken given that it only performs a small part of the change (commands are not updated here)

mikeSimonson:
But you confirm that it's not really an issue to wait until 2.3 EOL ?

stof:
@mikeSimonson it will be an issue because 3.0 will be released before that (and it removed the deprecated API), and people using the new 2.7 LTS will complain if our code triggers deprecation warnings and will ask us to update it.
However, supporting both APIs in parallel is not very hard

mikeSimonson:
@stof Can you link to an example ?

PR #193, opened 2014-12-13
Switch MigrationCommand from using deprecated DialogHelper to QuestionHelper

ackimwilliams:
A recent commit of migration tool uses `Symfony\Component\Console\Helper\QuestionHelper` instead of the soon to be deprecated `Symfony\Component\Console\Helper\DialogHelper`.  Composer update results in the following error:

` The helper "dialog" is not defined.`

The changes above switches the MigrateCommand from DialogHelper to QuestionHelper.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php
@@ -24,6 +24,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
 
 /**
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php
```diff
@@ -24,6 +24,7 @@
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
```
sdwrage: I think this will be needed in the execute command as well.

ddelrio1986:
@mikeSimonson why was this closed?

mikeSimonson:
@ddelrio1986 As far as I understand, we need to support both as long as symfony 2.4 is supported.

Is that roght @stof ?

snapshotpl:
Good question @ddelrio1986 . We need to fix it

mikeSimonson:
Also this pull request #198 make the change everywhere at once

PR #195, opened 2014-12-13
Switch from using deprecated DialogHelper to QuestionHelper

ackimwilliams:
DialogHelper deprecated in Symfony 2.5


stof:
This would break the support of the Symfony LTS, which seems a bad idea. Can you check the availability of the QuestionHelper and fallback to the DialogHelper for older versions ?

stof:
and please make consolidate the different similar changes in a single PR

PR #198, opened 2015-01-06
Replacing the deprecated class DialogHelper to QuestionHelper.

### LINKED ISSUE #361: migrations:status --show-version shows all executed migrations under "executed unavailable"
**Run:**
   doctrine:migrations:status --show-version

**Result:**
"Available Migration Versions" and "Previously Executed Unavailable Migration Versions" lists contain same migration numbers. Available migration versions are shown under both lists.

**Expected result:**
"Previously Executed Unavailable Migration Versions" should contain only migrations that are executed but no longer available.

**Fix:** https://github.com/doctrine/migrations/pull/360

nastron:
This fixes #361.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/ExecuteCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/ExecuteCommand.php
@@ -23,6 +23,7 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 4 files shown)


kusmierz:
Any chances to merge it?

mikeSimonson:
@kusmierz As soon as the 1.0 is out wich is soon too.

stof:
-1 for the current approach. We should not drop support for the Symfony 2.3 LTS while supporting both the old API and the new API in parallel is not hard.

mikeSimonson:
@stof Can you link to an example ?

bizmate:
Is this in anyway linked to a connection error? I could not figure out why the db connection is failing despite correct configuration being passed
http://pastie.org/10182707

mikeSimonson:
@bizmate No, can you open another issue

PR #205, opened 2015-02-16
Simplify handling of --no-interaction and confirmation

rjmunro:
If no interaction is set, pre-set confirmation to true, then only call execute from one place.

Reduces the number of places you have to update the code if you want to add more parameters to the execute method, e.g. --ignore-errors (https://github.com/doctrine/migrations/pull/188)

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/ExecuteCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/ExecuteCommand.php
@@ -82,14 +82,14 @@ EOT
             $path = is_bool($path) ? getcwd() : $path;
             $version->writeSqlFile($path, $direction);
         } else {
-            $noInteraction = $input->getOption('no-interaction') ? true : false;
-            if ($noInteraction === true) {
+            if ($input->getOption('no-interaction')) {
                 $confirmation = true;
             } else {
                 $confirmation = $this->getHelper('dialog')->askConfirmation($output, '<question>WARNING! You are about to execute a database migration that could result in schema changes and data lost. Are you sure you wish to continue? (y/n)</question>', false);
             }
-            if ($confirmation === true) {
-                $version->execute($direction, $input->getOption('dry-run') ? true : false);
+
+            if ($confirmation) {
+                $version->execute($direction, (boolean) $input->getOption('dry-run'));
             } else {
                 $output->writeln('<error>Migration cancelled!</error>');
             }
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/ExecuteCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/ExecuteCommand.php
@@ -83,12 +83,12 @@ EOT
             $version->writeSqlFile($path, $direction);
         } else {
             if ($input->getOption('no-interaction')) {
-                $confirmation = true;
+                $execute = true;
             } else {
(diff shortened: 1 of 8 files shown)


lib/Doctrine/DBAL/Migrations/Tools/Console/Command/ExecuteCommand.php
```diff
@@ -84,14 +84,14 @@ public function execute(InputInterface $input, OutputInterface $output)
         } else {
             $noInteraction = $input->getOption('no-interaction') ? true : false;
             if ($noInteraction === true) {
-                $version->execute($direction, $input->getOption('dry-run') ? true : false);
+                $confirmation = true;
```
mikeSimonson: Maybe changing the name of the variable to $execute make it a bit more understandable ?

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/ExecuteCommand.php
```diff
@@ -84,14 +84,14 @@ public function execute(InputInterface $input, OutputInterface $output)
...
             } else {
                 $confirmation = $this->getHelper('dialog')->askConfirmation($output, '<question>WARNING! You are about to execute a database migration that could result in schema changes and data lost. Are you sure you wish to continue? (y/n)</question>', false);
-                if ($confirmation === true) {
-                    $version->execute($direction, $input->getOption('dry-run') ? true : false);
-                } else {
-                    $output->writeln('<error>Migration cancelled!</error>');
-                }
+            }
+            if ($confirmation === true) {
+                $version->execute($direction, $input->getOption('dry-run') ? true : false);
```
stof: can you remove the ``? true : false`` ternary operator too ? It makes no sense to turn a boolean into another boolean.

mikeSimonson: Isn't getOption a return mixed function ?

stof: it depends how the option is configured. For option without value (which is the case here), it always return a boolean. For options with a value, it returns a string (the CLI only allows to pass strings)

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/ExecuteCommand.php
```diff
@@ -82,16 +82,16 @@ public function execute(InputInterface $input, OutputInterface $output)
             $path = is_bool($path) ? getcwd() : $path;
             $version->writeSqlFile($path, $direction);
         } else {
-            $noInteraction = $input->getOption('no-interaction') ? true : false;
-            if ($noInteraction === true) {
-                $version->execute($direction, $input->getOption('dry-run') ? true : false);
+            if ($input->getOption('no-interaction')) {
```
stof: ``$input->isInteractive()`` would be even cleaner actually (the ``no-interaction`` option is an Application-level option, but the Application has other ways to guess that the input cannot be interactive in some case).

rjmunro: I'm a bit nervous as this changes functionality slightly - I've seen people use constructs like `echo y | app/console doctrine:migrations:migrate` because they didn't know about the `--no-interaction` option - that kind of thing might break. I'd rather put that in a separate PR from this, which is just intended to tidy some things a bit.

rjmunro:
@stof Is this ready to be merged now, or do I need to do anything else?

mikeSimonson:
@rjmunro I am just waiting for the input of others on other PR before to merge this one but I feel likes it's ready to be merged.

mikeSimonson:
@rjmunro Thanks

PR #207, opened 2015-02-16
Add missing doctrine2 dependency.

exi:
lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
references
Doctrine\ORM\Tools\SchemaTool
which is included in doctrine/orm

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,8 @@
     ],
     "require": {
         "php": ">=5.3.2",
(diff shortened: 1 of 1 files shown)


stof:
This is only an optional dependency (the ORM is only needed when diffing from the ORM, but you could generate migrations in other ways)

PR #209, opened 2015-02-18
Remove wrong dependencies set up instruction with the correct composer u...

bizmate:
...pdate command, also make it more obvious in the Build section that a PHAR is ready for download

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -31,7 +31,8 @@ This should give you the flexibility you need for customizing your input/output 
 
 ### Building Your Phar
 
-Simply run `php package.php`, which will create the file: `build/doctrine-migrations.phar` for you.
+The PHAR is ready for [download here](http://github.com/downloads/ericclemmons/migrations/doctrine-migrations.phar), but if you still would like to build your PHAR
+run `php package.php`, which will create the file: `build/doctrine-migrations.phar` for you.
 Done! :)  This is a bit simpler than getting Phing/Ant going and running `phing build-migrations` and
 hoping the rest of the build dependencies work.
 
@@ -48,10 +49,9 @@ This can be fixed by setting the following in your php.ini:
 
 ### Installing Dependencies
 
-To install dependencies issue the following commands:
+To install dependencies run a composer update:
 
-    git submodule init
-    git submodule update
+    composer update
 
 ## Official Documentation
 


TomasVotruba:
:+1:

README.markdown
```diff
@@ -31,7 +31,8 @@ This should give you the flexibility you need for customizing your input/output
 
 ### Building Your Phar
 
-Simply run `php package.php`, which will create the file: `build/doctrine-migrations.phar` for you.
+The PHAR is ready for [download here](http://github.com/downloads/ericclemmons/migrations/doctrine-migrations.phar), but if you still would like to build your PHAR
```
stof: Given that the URL you are using is referencing the github downloads feature which has been shutdown more than 1 year ago, I'm quite sure the phar is totally outdated. So this looks like a bad idea to use it

TomasVotruba: Taking it further, this package doesn't seem like PHAR based. It requires regular updates starting with active repository.

Personally prefer composer dependecy, with CLI intergration that's same and available for whole project.

bizmate: @stof had no idea the URL was not autoupdate. If the project is run through automated tests, i.s. travis, then the PHAR can be autoupdated somewhere on a fixed URL when builds pass

TomasVotruba:
@bizmate The automatization seems like a good idea, but what about versioning? It would make sense to publish phar only for stable tags.

README.markdown
TomasVotruba: @bizmate The automatization seems like a good idea, but what about versioning? It would make sense to publish phar only for stable tags.

stof: The URL used here is for a legacy system. It cannot be updated (existing downloads are kept by github for BC reasons but they don't allow adding new ones: https://github.com/blog/1302-goodbye-uploads

stof: and I'm not even sure why phars are provided for this package actually. I always use it integrated in my project, not as a standalone CLI tool

bizmate: @TomasVotruba Indeed I would always focus on stable versions and if anyone uses alphas, overnight builds etc let them generated their own PHAR. I am a bit new to doctrine but I might create a pull request if I work out how to do these PHAR builds through travis. A *after_success* hook/step looks like the right place on travis.

bizmate: @stof About the legacy system, not sure what to replace it with but there must be another place where these PHARs could be hosted on. About the need for PHARs, i see them useful if you are introducing components in your project but are not ready to use a framework console yet, i.e. no symfony console exists

mikeSimonson: It's possible to upload the phar to github but only for tag release.
Travis can do that automatically.

https://github.com/blog/1547-release-your-software

If someone want to look it up.

TomasVotruba:
@bizmate It's worth asking for the need of PHAR first, before putting so much effort to generating it. 

Usually it's integrated with another package:

- Symfony - https://github.com/doctrine/DoctrineMigrationsBundle
- Nette - https://github.com/Zenify/DoctrineMigrations
- Laravel - https://github.com/atrauzzi/laravel-doctrine
...

README.markdown
TomasVotruba: @bizmate It's worth asking for the need of PHAR first, before putting so much effort to generating it. 

Usually it's integrated with another package:

- Symfony - https://github.com/doctrine/DoctrineMigrationsBundle
- Nette - https://github.com/Zenify/DoctrineMigrations
- Laravel - https://github.com/atrauzzi/laravel-doctrine
...

PR #210, opened 2015-02-22
adding build status

mikeSimonson:


stof:
I updated the URLs in https://github.com/doctrine/migrations/commit/2c9c9482e9062ba3a53eea3babb96f4cccbdb9f3 to use uptodate ones (SVG rather than PNG and not using the old ``secure.travis-ci.org`` subdomain)

PR #211, opened 2015-02-22
Timing per query

mikeSimonson:


stof:
@mikeSimonson this needs a rebase to fix conflicts

PR #212, opened 2015-02-22
Allow symfony 3.0 components

nicolas-grekas:
Tests should tell if any deprecated interfaces of Symfony are used. If not, then the bundle is defacto compatible with 3.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,8 +14,8 @@
         "doctrine/dbal": "~2.0"
     },
     "require-dev": {
-        "symfony/console": "2.*",
-        "symfony/yaml": "2.*"
+        "symfony/console": "~2.3|~3.0",
+        "symfony/yaml": "~2.3|~3.0"
     },
     "suggest": {
         "symfony/console": "to run the migration from the console"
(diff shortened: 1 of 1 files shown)


xabbuh:
@nicolas-grekas It would make sense to update the Travis configuration as well to run dedicated tests with Symfony 3.0 and, for example, PHP 5.6, doesn't it?

mikeSimonson:
I am going to put that on hold until symfony 3.0 gets out.

composer.json
```diff
@@ -25,6 +25,7 @@
             "Doctrine\\DBAL\\Migrations": "lib"
         }
     },
+    "minimum-stability": "dev",
```
Ener-Getick: Not sure this is a good idea to always use this ...
What about using [this](https://github.com/KnpLabs/KnpMenuBundle/blob/master/.travis.yml#L30) ?

mikeSimonson:
I ended up merging a very similar PR. https://github.com/doctrine/migrations/commit/4e909f2e661a8414a3e04ce987a09c9e849cd13f

Thanks @nicolas-grekas

PR #214, opened 2015-03-02
travis: PHP 7.0 nightly added; --dev is by default

TomasVotruba:
Fixed spaces to 2, as in other parts of file

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,13 @@ php:
   - 5.4
   - 5.5
   - 5.6
+  - 7.0
   - hhvm
  
 matrix:
-    allow_failures:
-        - php: hhvm
+  allow_failures:
+    - php: 7.0
+    - php: hhvm
 
 install:
-  - "composer install --dev"
+  - composer install


stof:
thanks.

@mikeSimonson what do you think about disallowing failures on HHVM given that the testsuite passes (I think it is too early for PHP 7 though, even though it is passing here)

TomasVotruba:
Thank you

mikeSimonson:
Well if the tests already passes why not ?

stof:
@mikeSimonson what if a new commit in php-src breaks something temporarily ? Do we want our builds to be subject to the instability of the PHP-7 branch ?
However, I'm not against removing it from allow_failures as well. I'm just unsure about the best strategy for now

mikeSimonson:
sorry I meant disallowing failure for hhvm, for php 7 it's a little bit too early I think

stof:
@mikeSimonson done in https://github.com/doctrine/migrations/commit/8a7783409d6e41f8891e2ce997d39cc9145ec167

PR #215, opened 2015-03-02
updated deprecated use of Yaml::parse

itised:
Parsing a filename has been deprecated. The preffered method is to pass the file's contents to Yaml::parse


stof:
thanks

PR #217, opened 2015-03-03
migrationsTableName should be case insensitive

vershnik:
Oracle returns all table names UPPERCASE, so this check fails and migrationsTable is included to every migration.


mikeSimonson:
@stof @beberlei I believe it's a valid change. Can you confirm ?

beberlei:
yes

mikeSimonson:
@vershnik Thanks

PR #218, opened 2015-03-05
Version number

mikeSimonson:
@stof Any comment ?

### DIFF
--- a/bin/doctrine-migrations.php
+++ b/bin/doctrine-migrations.php
@@ -43,7 +43,7 @@ if (file_exists($configFile)) {
 $helperSet = ($helperSet) ?: new \Symfony\Component\Console\Helper\HelperSet();
 $helperSet->set(new \Symfony\Component\Console\Helper\DialogHelper(), 'dialog');
 
-$cli = new \Symfony\Component\Console\Application('Doctrine Migrations', \Doctrine\DBAL\Migrations\MigrationsVersion::VERSION);
+$cli = new \Symfony\Component\Console\Application('Doctrine Migrations', \Doctrine\DBAL\Migrations\MigrationsVersion::VERSION());
 $cli->setCatchExceptions(true);
 $cli->setHelperSet($helperSet);
 $cli->addCommands(array(
--- a/box.json
+++ b/box.json
@@ -8,5 +8,6 @@
             "in": ["lib", "vendor"]
         }
     ],
+    "git-version": "git-version",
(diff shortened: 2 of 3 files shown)


bin/doctrine-migrations.php
```diff
@@ -17,6 +17,10 @@
  * <http://www.doctrine-project.org>.
  */
 
+
+/**
+ * @version     @git_version@
```
stof: why this ?

lib/Doctrine/DBAL/Migrations/MigrationsVersion.php
```diff
@@ -1,25 +0,0 @@
...
- * This software consists of voluntary contributions made by many individuals
- * and is licensed under the LGPL. For more information, see
- * <http://www.doctrine-project.org>.
- */
-
-namespace Doctrine\DBAL\Migrations;
-
-class MigrationsVersion
-{
-    const VERSION = '2.0.0-DEV';
```
stof: Removing this constant is a BC break. Some people might rely on it.

and it will not be available anymore at all when installing from composer

box.json
```diff
@@ -8,5 +8,10 @@
             "in": ["lib", "vendor"]
         }
     ],
+    "files": [
+        "LICENSE"
```
stof: :+1:

bin/doctrine-migrations.php
mikeSimonson: No really useul. More for clarity.

bin/doctrine-migrations.php
```diff
@@ -43,7 +47,7 @@
 $helperSet = ($helperSet) ?: new \Symfony\Component\Console\Helper\HelperSet();
 $helperSet->set(new \Symfony\Component\Console\Helper\DialogHelper(), 'dialog');
 
-$cli = new \Symfony\Component\Console\Application('Doctrine Migrations', \Doctrine\DBAL\Migrations\MigrationsVersion::VERSION);
+$cli = new \Symfony\Component\Console\Application('Doctrine Migrations', '@git-version@');
```
KingCrunch: This only works for the PHAR package, right? I assume this will break, if I use this library as dependency.

mikeSimonson: Yeah.
I was hoping to find a way to make it work but then I need to make sure that the git repo is there and that can't be guaranteed.

PR #156, opened 2014-01-30
Remove unused AbstractMigration::$configuration and fix some CS

rybakit:

PR #219, opened 2015-03-05
Prefer configuration from options

pCoLaSD:
The configuration options from the command line should override the configuration property of all commands. Otherwise the specified options will have no effect on the command.

The background is, that we define our migrations and entities in different packages with different name spaces and must migrate with individual configuration files for every subset of affected entities or tables.

### DIFF
--- a/tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/AbstractCommandTest.php
+++ b/tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/AbstractCommandTest.php
@@ -95,14 +95,6 @@ class AbstractCommandTest extends MigrationTestCase
      */
     public function testMigrationConfigurationReturnsConnectionFromInputOption()
     {
-        $content = <<<EOF
-<?php
-
-return array('driver' => 'pdo_sqlite', 'memory' => true);
-EOF;
-        $filename = \tempnam(\sys_get_temp_dir(), 'migrations_test');
-        file_put_contents($filename, $content);
-
         $input = $this->getMockBuilder('Symfony\Component\Console\Input\ArrayInput')
             ->setConstructorArgs(array(array()))
             ->setMethods(array('getOption'))
@@ -110,20 +102,14 @@ EOF;
 
         $input->expects($this->any())
             ->method('getOption')
-            ->with($this->logicalOr($this->equalTo('db-configuration'), $this->equalTo('configuration')))
-            ->will($this->returnCallback(function ($name) use ($filename) {
-                if ('db-configuration' === $name) {
-                    return $filename;
-                }
-                return null;
-            }));
+            ->will($this->returnValueMap(array(
+                array('db-configuration', __DIR__ . '/_files/db-config.php')
+            )));
 
         $actualConfiguration = $this->invokeMigrationConfigurationGetter($input);
 
         $this->assertInstanceOf('Doctrine\DBAL\Migrations\Configuration\Configuration', $actualConfiguration);
         $this->assertEquals($this->getSqliteConnection(), $actualConfiguration->getConnection());
-
-        unlink($filename);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


mikeSimonson:
@pCoLaSD Can you provide a test ?

pCoLaSD:
Here we go! I have made a little change to the abstract command. Otherwise I must mock the application too. Never wrote a test for a protected method in an abstract class before. Learning never stops.

mikeSimonson:
Thanks a lot.

stof:
why are you removing the code caching the configuration in a property to avoid reloading it multiple times ?

mikeSimonson:
@stof The configuration property was removed by this PR #156 .
I taught that you saw it ?

stof:
@mikeSimonson not in the same class

pCoLaSD:
This PR does not refer to AbstractMigration, it refers to AbstractCommand.
The reason why I removed code caching is, that otherwise any input option passed to migration commands would have no effect on the affected one. This is no preferred behaviour in my opinion.

mikeSimonson:
@stof Sorry

mikeSimonson:
@stof do you agree ?

tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/AbstractCommandTest.php
```diff
@@ -0,0 +1,166 @@
...
+     */
+    public function testMigrationConfigurationReturnsConfigurationFileOption()
+    {
+        $content = <<<EOF
+name: "name"
+table_name: "migrations_table_name"
+migrations_namespace: "migrations_namespace"
+EOF;
+        $filename = \sys_get_temp_dir() . '/' . uniqid('migrations') . '.yml';
+        file_put_contents($filename, $content);
```
BreiteSeite: It would be better to commit that file and reference it here... instead of dynamically create it in the test.

tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/AbstractCommandTest.php
```diff
@@ -0,0 +1,166 @@
...
+
+        $input = $this->getMockBuilder('Symfony\Component\Console\Input\ArrayInput')
+            ->setConstructorArgs(array(array()))
+            ->setMethods(array('getOption'))
+            ->getMock();
+
+        $input->expects($this->any())
+            ->method('getOption')
+            ->with($this->logicalOr($this->equalTo('db-configuration'), $this->equalTo('configuration')))
+            ->will($this->returnCallback(function ($name) use ($filename) {
```
BreiteSeite: Should use [returnValueMap()](https://phpunit.de/manual/current/en/test-doubles.html#test-doubles.stubs.examples.StubTest5.php)

pCoLaSD:
@stof I hope this is better now.

mikeSimonson:
@pCoLaSD Thanks

Can you explain me why any input option would have no effect on the command if cached in the private property ?
Do you mean that if people call command with specific options ( like --configuration= ) their options wouldn't have any effect on the command extending the AbstractCommand class ?
If so I don't really see the problem because the configuration is build with those options.
What am I missing ?

pCoLaSD:
@mikeSimonson You miss the use case when the configuration was injected via setMigrationConfiguration() _before_ the configuration options were read. We use Zend Framework 2 with the [Doctrine ORM Module](https://github.com/doctrine/DoctrineORMModule). The [MigrationsCommandFactory](https://github.com/doctrine/DoctrineORMModule/blob/master/src/DoctrineORMModule/Service/MigrationsCommandFactory.php) injects the configuration in the application bootstrapping process. So you have no chance to switch between multiple configurations via the `---configuration` option (or any other migrations option). The options provided in the terminal will have no effect on the command.

mikeSimonson:
So now we have a migrationConfiguration property and a configuration property.
The migrationConfiguration property contains the configuration created taking into account the command line options.
The configuration property only contains the configuration injected by the setter.

When any (config) command line option is passed to the migration the migrationConfiguration is set with the new generated configuration.
If no (config) option is passed the migrationConfiguration property is set to the value of the configuration one (if any).

The  migrationConfiguration is always returned.
@pCoLaSD Right ?

pCoLaSD:
@mikeSimonson Correct.

seyfer:
https://github.com/doctrine/migrations/issues/228

Please fix it back!

PR #222, opened 2015-03-20
Small fix for prevent php notice.

r0b1n:


mikeSimonson:
@r0b1n Thanks

lib/Doctrine/DBAL/Migrations/MigrationsVersion.php
```diff
@@ -39,7 +39,7 @@ public static function VERSION(){
      * Check if doctrine migration is installed by composer or
      * in a modified (not tagged) phar version.
      */
-    private static function isACustomPharBuild($gitVersion) {
+    private static function isACustomPharBuild($gitversion) {
```
xabbuh: If you want to be completely consistent, changing the variable in the method body would have probably better. Or just change the argument name in the docblock now. :)

mikeSimonson: @xabbuh Thanks

PR #223, opened 2015-03-23
Adds empty string return in AbstractMigration::getDescriopion

rtreffler:
Just being strict on return type.
IDE (phpStorm) can use this to add return statement into method.


mikeSimonson:
@rtreffler Thanks

rtreffler:
Happy to help improving library I'm using :)

Issue #204, opened 2015-02-04
Be able to register migrations from multiple folders


jjanvier:
Hello Doctrine team,

At Akeneo, our migrations are stored in different folders (yes, we really need that). Would you be interested a PR that allows this feature ? Basically it seems that only the class `Doctrine\DBAL\Migrations\Configuration\Configuration` should be modified.

NigelGreenway:
+1 for this.

I have been looking at the best way of doing this myself.

@jjanvier Out of interest, how do you separate yours? At the moment I have been looking at `connection/release/versionxxxxxxxx.php`

chrisguitarguy:
@mikeSimonson I can help with this! Just need to know how you would want this to work? Just glob multiple directories for `Version*.php`? Or does there need to be separate namespaces?

I would think the namespace wouldn't matter. Just recursively search the directory, `require_once` every file, and look for all the classes from the directory filtering out the ones that don't extend `AbstractMigration`. Similar to what the [doctrine persistence annotation driver does](https://github.com/doctrine/common/blob/5add480dcb5da8cd8e4e9c4b5055a2c9bbd05af3/lib/Doctrine/Common/Persistence/Mapping/Driver/AnnotationDriver.php#L193).

mikeSimonson:
I would like to know why @jjanvier and @NigelGreenway need that feature to find the best way to do it.

If it's to have a folder to keep older migration, it make sense to me.
We wouldn't need to change the namespace and a simple recursive glob would work.

Is it what they are looking for ?

chrisguitarguy:
I can think of a good reason for multiple directories and removing the namespace: integration with framework modules or bundles. Each one could hook it and register their own migrations.

NigelGreenway:
The way I would like to use it is to separate them by feature, this way it would only be an extra level of directories.

Sorry I can do anything myself at the moment.

mikeSimonson:
Separating by feature looks like a valid use case too but the multi integration does not. The migration are going to be different anyway, it makes more sense to have different folder and different config for different integration IMO.

chrisguitarguy:
Would this require recursively searching a single directory? Or being able to register multiple directories in the config? Maybe both?

mikeSimonson:
I think that we can probably get away just searching recursively the folder. The question is should we add an option to keep the old behavior ?

chrisguitarguy:
[...]

PR #227, opened 2015-04-04
Make the Diff Command Independent of the ORM

### LINKED ISSUE #226: Make the Diff Command Independent of the ORM
I use the [DBAL schema](https://github.com/doctrine/dbal/tree/master/lib/Doctrine/DBAL/Schema) stuff independent of the ORM. It's a great way to maintain platform independent schema. There's probably other folks out there who do the same.

To make that usage work well with migrations, specifically the diff command, I had to subclass [`DiffCommand`](https://github.com/doctrine/migrations/blob/master/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php) and pull out the entity manager specific parts.

It would be excellent if the way [`$toSchema`](https://github.com/doctrine/migrations/blob/65978aa4e9ffca3bb632225ad8c6320077d80d85/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php#L88) was generated was abstracted away and easily replaced. The default would, of course, be the entity manager way.

I see a couple ways forward, if you folks like this idea: centralizing the `$toSchema` creation in a single method so subclassing can be done easily or creating a new interface (something like `CurrentSchemaProvider`) and implementation to generate the current schema. Separate classes would probably be more testable and nice.

Thoughts? I can help implement this if it sounds like a good idea.

chrisguitarguy:
Closes #226 

Feel free to close or discard. Sending a PR seemed like the easiest way to do this: https://github.com/doctrine/migrations/issues/226#issuecomment-89656128


mikeSimonson:
Well, what is your goal exactly ?

Do you want to provide something similar to that PR #79 ?

chrisguitarguy:
Right now the diff command relies on the ORM to generate the schema to which the database will be migrated. See:

- https://github.com/doctrine/migrations/blob/39d6223d919baa35883270199c7e935acc440257/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php#L68
- https://github.com/doctrine/migrations/blob/39d6223d919baa35883270199c7e935acc440257/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php#L88

The goal is to be able to get the schema to which the database will be migrated from anywhere. Not just the ORM metadata. So if someone is using the DBAL Schema stuff outside of the ORM (I do), they can use the `migrations:diff` command to generate migrations.

#79 sees more about code generation *after* the `$toSchema` is generated from the ORM metadata. This is about changing the way the the `$toSchema` objects are generated.

mikeSimonson:
I fail to understand in which context it might be useful ?
Can you give an example ?

chrisguitarguy:
> I fail to understand in which context it might be useful ?

The [schema library](https://github.com/doctrine/dbal/tree/master/lib/Doctrine/DBAL/Schema) is not coupled to the ORM. So sometimes it gets used outside of the ORM. Here's an example [from the Symfony component](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Acl/Dbal/Schema.php).

For case where the schema component is used outside of the ORM, it would nice to be able to use the diff command to generate migrations.

Take #204, if that goes through and Symfony starts providing migrations for its ACL component, they might want to be able to generate those automatically using the diff command.

To give some code, I basically want to be able to do this:

```php
use Doctrine\DBAL\Migrations\Provider\SchemaProvider;
use Doctrine\DBAL\Migrations\Tools\Console as MigrationsConsole;

class StubSchemaProvider implements SchemaProvider
{
   private $schema;

   public function __construct(Schema $schema)
   {
       $this->schema = $schema;
   }

   public function createSchema()
   {
       return $this->schema;
   }
}

/** @var Schema $toSchema */
$toSchema = createASchemaObjectOutsideOfTheOrm();

/** @var Symfony\Component\Console\Application $app */
$app->addCommands([
    // add all the console commands for migrations here

    // Customize the diff command with another schema provider not tied to the ORM
    new MigrationsConsole\DiffComamnd(new StubSchemaProvider($toSchema)),
]);

$app->run();
```

Is this a common use case? No. That's why the default is still to use the ORM. This just pulls out the schema generation to a separate object and hides it behind an interface, providing an extension point if necessary and coupling the diff command less tightly to the ORM.

chrisguitarguy:
https://github.com/bolt/bolt is another [good example](https://github.com/bolt/bolt/blob/9612b0a5f19ac17a9a0577fb5ed672398bfeb6e8/src/Database/IntegrityChecker.php#L334-L343) of an application that uses the schema component directly, outside of the ORM.

ddelrio1986:
I think this PR is a great idea. Would be very useful for people wanting to use Doctrine Migrations without making use of Doctrine ORM.

mikeSimonson:
Ok, I see the point now.
The problem is that I pretty much never use the schema manager.
We just need to write the documentation about that feature https://github.com/doctrine/migrations-documentation
and some tests.

chrisguitarguy:
> We just need to write the documentation about that feature and some tests.

I can work on both of those. Still trying to wrap my head around some of the test suite.

Is it okay to add `doctrine/orm` to `require-dev` so I can write some more integration-like tests for the `OrmSchemaProvider`?

chrisguitarguy:
Not sure if I'm on the right track with b14320d3d4b805022683e4587443f3936cbf350d -- none of the commands appeared to have a lot of tests. I added some end-to-end tests that run things through a symfony console application instance. The status, generate, diff, version, and migrate commands are all touched now at least.

Added a `StubSchemaProvider` implementation to the library itself, since it seemed useful in general as well as for testing.

chrisguitarguy:
Failure is on HHVM, maybe due to the ORM having PHPUnit in its `require-dev` dependencies?

mikeSimonson:
Composer is not going to pull a the dev dependencies from a required library.

Seems to be a problem with the pre-build phpunit of travis.

Apparently right now the only way to go around it is to require phpunit in the dev dependencies.
And add those changes : https://github.com/thephpleague/commonmark/pull/92/files

chrisguitarguy:
Do you want me to add phpunit on this branch? Or would you rather do it elsewhere?

mikeSimonson:
no adding it on this branch is perfect

chrisguitarguy:
Looks like it worked! Need anything else from me here?

mikeSimonson:
@chrisguitarguy Thanks a lot.

The only reason why we need the orm is because of the tests right ?

chrisguitarguy:
> The only reason why we need the orm is because of the tests right?

Yep. There's a few typehints for ORM classes in `OrmSchemaProvider`, but that's it.

PR #230, opened 2015-04-07
Reversed the conditional in isACustomPharBuild()

### LINKED ISSUE #229: MigrationsVersion::VERSION returns @git-version@
I may have misunderstood this but when installing doctrine/migrations as a standalone tool, @git-version@ is being returned as the version string rather than the static $version property (currently v1.0.0-alpha3).

liquorvicar:
Fixes #229

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/MigrationsVersion.php
+++ b/lib/Doctrine/DBAL/Migrations/MigrationsVersion.php
@@ -40,6 +40,6 @@ class MigrationsVersion
      * in a modified (not tagged) phar version.
      */
     private static function isACustomPharBuild($gitversion) {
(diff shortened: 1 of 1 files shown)


mikeSimonson:
Thanks @liquorvicar

stof:
The fix is wrong. the function name now does the opposite of what the name describes. Either the *usage* of the function should have changed, or the function should also be renamed

mikeSimonson:
This is so confusing.

If '@git-version@' is passed to the function, it is not a custom phar build and thus the function should return false.
If something else is passed to the function that means that the '@git-version@' string has been replaced by box and thus that it is a custom phar build (return true).

Isn't that what is going on here ?

stof:
ah sorry, indeed

PR #232, opened 2015-04-25
Implement "--range-from/--range-to" and add tests for VersionCommand

### LINKED ISSUE #142: Allow multiple versions in doctrine:migrations:version command
In some cases it is useful to mark all migrations upto a certain point as completed. I'd suggest the following syntax for this:

    doctrine:migrations:version --add version1..version2

This would mark all versions between (and including) version1 and version2 as added. Note that version1 and version2 do not need to exist; only at least one version should be between those values (and an error may be thrown otherwise). (That would allow version1 to be 0, or version2 to be 99999999999999).

Similar syntax could be used for other commands where it makes sense.

aivus:
Fix #142

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/VersionCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/VersionCommand.php
@@ -115,33 +115,28 @@ EOT
     {
         $affectedVersion = $input->getArgument('version');
 
-        $allOption = $input->getOption('all');
+        $allOption   = $input->getOption('all');
         $rangeOption = $input->getOption('range');
 
         if ($allOption && $rangeOption) {
             throw new \InvalidArgumentException('Options --all and --range both used. You should use only one of them.');
         }
 
-        switch (true) {
-            case $allOption === true:
-                $availableVersions = $this->configuration->getAvailableVersions();
-                foreach ($availableVersions as $version) {
+        if ($allOption) {
+            $availableVersions = $this->configuration->getAvailableVersions();
+            foreach ($availableVersions as $version) {
+                $this->mark($version, true);
+            }
+        } elseif ($rangeOption) {
+            $availableVersions = $this->configuration->getAvailableVersions();
+            $versionEnd        = $input->getArgument('version_end');
+            foreach ($availableVersions as $version) {
+                if ($version >= $affectedVersion && $version <= $versionEnd) {
                     $this->mark($version, true);
                 }
-                break;
-
-            case $rangeOption === true:
-                $availableVersions = $this->configuration->getAvailableVersions();
-                $versionEnd        = $input->getArgument('version_end');
-                foreach ($availableVersions as $version) {
-                    if ($version >= $affectedVersion && $version <= $versionEnd) {
-                        $this->mark($version, true);
-                    }
-                }
-                break;
-
-            default:
-                $this->mark($affectedVersion);
+            }
+        } else {
+            $this->mark($affectedVersion);
         }
     }
 
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/VersionCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/VersionCommand.php
@@ -55,9 +55,11 @@ class VersionCommand extends AbstractCommand
             ->setName('migrations:version')
             ->setDescription('Manually add and delete migration versions from the version table.')
             ->addArgument('version', InputArgument::OPTIONAL, 'The version to add or delete.', null)
+            ->addArgument('version_end', InputArgument::OPTIONAL, 'The version to add or delete.', null)
             ->addOption('add', null, InputOption::VALUE_NONE, 'Add the specified version.')
             ->addOption('delete', null, InputOption::VALUE_NONE, 'Delete the specified version.')
             ->addOption('all', null, InputOption::VALUE_NONE, 'Apply to all the versions.')
+            ->addOption('range', null, InputOption::VALUE_NONE, 'Apply to versions range.')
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Migrations/Tools/Console/Command/VersionCommand.php
```diff
@@ -85,61 +92,81 @@ public function execute(InputInterface $input, OutputInterface $output)
...
-        } else {
-            $this->mark($version);
+        $allOption = $input->getOption('all');
+        $rangeOption = $input->getOption('range');
+
+        if ($allOption && $rangeOption) {
+            throw new \InvalidArgumentException('Options --all and --range both used. You should use only one of them.');
+        }
+
+        switch (true) {
```
mikeSimonson: Why a switch statement here ?

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/VersionCommand.php
```diff
@@ -85,61 +92,81 @@ public function execute(InputInterface $input, OutputInterface $output)
...
+            case $allOption === true:
+                $availableVersions = $this->configuration->getAvailableVersions();
+                foreach ($availableVersions as $version) {
+                    $this->mark($version, true);
+                }
+                break;
+
+            case $rangeOption === true:
+                $availableVersions = $this->configuration->getAvailableVersions();
+                $versionEnd        = $input->getArgument('version_end');
```
mikeSimonson: What happens when version_end is not specified ?

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/VersionCommand.php
```diff
@@ -55,9 +55,11 @@ protected function configure()
             ->setName('migrations:version')
             ->setDescription('Manually add and delete migration versions from the version table.')
             ->addArgument('version', InputArgument::OPTIONAL, 'The version to add or delete.', null)
+            ->addArgument('version_end', InputArgument::OPTIONAL, 'The version to add or delete.', null)
```
mikeSimonson: the version_end is only used in combination with range right ?
If so it would be better just after the range option IMO.
Also the comment is not right on that one.

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/VersionCommand.php
aivus: I use `switch(true)` instead of
```
if (...) {
    ...
} elseif (...) {
    ...
} else {
    ...
}
```

For example like here: https://github.com/symfony/symfony/blob/a8e2c743c23fe17d360868868ec1fb35529040c9/src/Symfony/Component/Console/Helper/SymfonyQuestionHelper.php#L58

This makes the code easy to read

mikeSimonson:
@aivus Thanks a lot.
I commented a bit on some question I had and some stuff that need to be improved but your contribution is really appreciated.
In general try to avoid changing the style of some line ( you are removing a lot of spaces before exclamation mark ) as it makes your PR harder to read.

Thanks again for the tests :+1: :sunglasses:

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/VersionCommand.php
mikeSimonson: When when I read it, I find a if, elseif much more telling then a switch true.
I would be really against that, it's super hacky.

aivus: Ok, I'll change it on `if`

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/VersionCommand.php
aivus: Nothing to be migrated

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/VersionCommand.php
aivus: `Command` component set order of options and arguments.
It set options for beginning and arguments after that:

`migrations:version [--add] [--delete] [--all] [--range] [--configuration[="..."]] [--db-configuration[="..."]] [version] [version_end]`

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/VersionCommand.php
```diff
@@ -55,9 +55,11 @@ protected function configure()
             ->setName('migrations:version')
             ->setDescription('Manually add and delete migration versions from the version table.')
             ->addArgument('version', InputArgument::OPTIONAL, 'The version to add or delete.', null)
+            ->addArgument('version_end', InputArgument::OPTIONAL, 'The end version to add or delete. Works only with --range option.', null)
```
mikeSimonson: Wouldn't that make more sense then to make a option of that argument ?
I have a hard time knowing when to use one or the other.

aivus: What do you think about:
```--range-from=XXX --range-to=YYY```?

Or maybe you have other suggestions?

mikeSimonson: I like the range-from, range-to.
Seems more clear to me.

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/VersionCommand.php
aivus: Changed

mikeSimonson:
Tanks again @aivus

PR #234, opened 2015-04-30
Don't Use Normalized Table Names in Schema Filters

### LINKED ISSUE #179: Case Sensitivity in DiffCommand.php when using schema_filter(Oracle)
Basic setup Symfony 2.5 Project, Using Doctrine to access a Share database.

Consider a Oracle schema (DBSCHEMA) with multiple tables, of which we want to manage some using Doctrine Migrations. 

The tables we want to Manage are prefixed with a string eg. MIGTABLE1, MIGTABLE2.
There ofcourse are other tables are called diffrently XYSOMETABLE, ABSOMETABLE2

Do only allow access to the Migrationtabes, one would set a schema_filter like '~^MIG~' (note uppercase) which will properly filter all tables starting with MIG, and make all other tables inaccessible as intended in symfony.

Now, when Using the DiffCommand shell, the schema will Ignore ALL tables, and produce migrations removing MIGTABLE1 and MIGTABLE2.

Turns out this is an issue with the loop applying the schema_Filter rules in 
[DiffCommand.php Line 93](https://github.com/doctrine/migrations/blob/master/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php#L93)

the variable $tableName contains tables as lowercased string (e.g. dbschema.migtable1, dbschema.xysometable)  which is shortened to migtable1, xysometable (still lowercase).

Line 95 will apply the filter defined in schema_filter, which (usually) will look/Match for an Uppercase tablename.

Since "migtable1" will not match "~^MIG~", the tables are discarded from $toSchema.

Simply switching the schema_filter to lowercase "~^mig~" will make the migrations work, but also filter the tables when trying to access them inside the Symfony application.

The current Workaround is to use a regular extression that will match both upper and lowercase tablenames, which is quite unintuitive and (afaik.) not documented.

chrisguitarguy:
Closes #179 

If a user enters an uppercased table name into a schema (via the schema API itself or entity metadata) that value is used in the SQL dumped by the diff command. The lowercased table names that are used internally in doctrine are an implementation detail of the library itself.

See the docblock for [`AbstractAsset#getFullQualifiedName`](https://github.com/doctrine/dbal/blob/41e718eab32980697c876d217ff4413cf86a4d2b/lib/Doctrine/DBAL/Schema/AbstractAsset.php#L117-L120).

This changes the filter expression stuff to use the table name entered by the user, rather than the normalized version. Also adds a test case for the schema filters in general and a test case to reproduce the bug.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
@@ -92,10 +92,9 @@ EOT
 
         //Not using value from options, because filters can be set from config.yml
         if ( ! $isDbalOld && $filterExpr = $conn->getConfiguration()->getFilterSchemaAssetsExpression()) {
-            $tableNames = $toSchema->getTableNames();
-            foreach ($tableNames as $tableName) {
-                $tableName = substr($tableName, strpos($tableName, '.') + 1);
-                if ( ! preg_match($filterExpr, $tableName)) {
+            foreach ($toSchema->getTables() as $table) {
+                $tableName = $table->getName();
+                if ( ! preg_match($filterExpr, $this->resolveTableName($tableName))) {
                     $toSchema->dropTable($tableName);
                 }
             }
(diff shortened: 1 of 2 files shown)


mikeSimonson:
@chrisguitarguy Thanks a lot

mikeSimonson:
just waiting for the tests to pass

chrisguitarguy:
> just waiting for the tests to pass

I'm waiting for the tests to tell me I used short array syntax and broke the PHP 5.3 build :)

mikeSimonson:
I am going to deprecate 5.3 anyway/

mikeSimonson:
@chrisguitarguy Thanks

PR #235, opened 2015-04-30
Prefer the Injected Migrations Configuration Over Fallback Files

### LINKED ISSUE #228: One of the latest commits breaks migrations with ZF2 and DoctrineModule!
I'm have updated migrations dev-master branch with composer

```
Updating doctrine/migrations dev-master (058a463 => 65978aa)
```

After that I'm always have error with migrations in console like this:

```
php public/index.php migrations:status
PHP Fatal error:  Class 'Application\Version20140623142428' not found in /var/www/domain.com/vendor/doctrine/migrations/lib/Doctrine/DBAL/Migrations/Version.php on line 113
PHP Stack trace:

```

But I have migrations configured with another namespace. Here is my doctrine config for migrations

```
'migrations_configuration' => [
            'orm_default' => [
                'directory' => 'data/DoctrineORMModule/Migrations/',
                'namespace' => 'DoctrineORMModule\Migrations',
                'table'     => 'migrations',
            ],
        ],
```

And part of existed migration code for example:

```
namespace DoctrineORMModule\Migrations;

use Doctrine\DBAL\Migrations\AbstractMigration;

/**
 * Auto-generated Migration: Please modify to your needs!
 */
class Version20140623142428 extends AbstractMigration
{
```

When I downgrade migrations commit with composer then all works again!

```
- Updating doctrine/migrations dev-master (65978aa => 058a463)
```

So, one of the latest changes breaks migrations with ZF2!

chrisguitarguy:
Closes #228 

The `configuration` command line option still trumps everything, but if a configuration was injected the commands will use that first.

This makes the most sense to me as a user: if the command line option is there, prefer that, otherwise use what was injected OR fallback to the `migrations.{xml,yml,yaml}` files if no config was injected.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/AbstractCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/AbstractCommand.php
@@ -105,6 +105,8 @@ abstract class AbstractCommand extends Command
                 $class = $info['extension'] === 'xml' ? 'Doctrine\DBAL\Migrations\Configuration\XmlConfiguration' : 'Doctrine\DBAL\Migrations\Configuration\YamlConfiguration';
                 $configuration = new $class($this->getConnection($input), $this->getOutputWriter($output));
                 $configuration->load($input->getOption('configuration'));
+            } elseif ($this->configuration) {
+                $configuration = $this->configuration;
             } elseif (file_exists('migrations.xml')) {
                 $configuration = new XmlConfiguration($this->getConnection($input), $this->getOutputWriter($output));
                 $configuration->load('migrations.xml');
@@ -114,8 +116,6 @@ abstract class AbstractCommand extends Command
             } elseif (file_exists('migrations.yaml')) {
                 $configuration = new YamlConfiguration($this->getConnection($input), $this->getOutputWriter($output));
                 $configuration->load('migrations.yaml');
-            } elseif ($this->configuration) {
-                $configuration = $this->configuration;
             } else {
                 $configuration = new Configuration($this->getConnection($input), $this->getOutputWriter($output));
             }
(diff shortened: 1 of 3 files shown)


mikeSimonson:
@chrisguitarguy Completely agree with the idea. Makes more sense to me too but it looks like the code does the opposite of what you describe.
I will need to look at it a bit more.

chrisguitarguy:
It does the config file passed into the CLI: https://github.com/chrisguitarguy/migrations/blob/e40b333856fc0949dbf57989c795370c5d05f83d/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/AbstractCommand.php#L103

Then checks for the injected config: https://github.com/chrisguitarguy/migrations/blob/e40b333856fc0949dbf57989c795370c5d05f83d/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/AbstractCommand.php#L108

Then falls back on the defaults: https://github.com/chrisguitarguy/migrations/blob/e40b333856fc0949dbf57989c795370c5d05f83d/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/AbstractCommand.php#L110-L121

[There's a test](https://github.com/chrisguitarguy/migrations/blob/949622d3270815a8373339796fbd6ccae4e08558/tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/AbstractCommandTest.php#L171) to verify this. You can checkout 949622d3270815a8373339796fbd6ccae4e08558 to watch it fail.

The weirdness is that there's two configuration properties: `configuration` and `migrationConfiguration`. There's as PR someplace that had an explanation of why, but I can't find it now. Something about justing using one property meant that the command line option was never considered.

mikeSimonson:
Yeah sorry I read the diff the wrong way.

seyfer:
:+1:

PR #236, opened 2015-05-02
Check for Both EntityManagerInterface and EntityManager

chrisguitarguy:
Prevents some fatal errors when folks are using an old ORM version.

See the comments here: https://github.com/doctrine/migrations/commit/e9ef6385d0b5ae7d8177fc9194cf9b59e58c460a

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Provider/OrmSchemaProvider.php
+++ b/lib/Doctrine/DBAL/Migrations/Provider/OrmSchemaProvider.php
@@ -19,6 +19,7 @@
 
 namespace Doctrine\DBAL\Migrations\Provider;
 
+use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Tools\SchemaTool;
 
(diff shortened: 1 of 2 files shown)


Ocramius:
Note: I'd rather lock onto ORM 2.4+ instead.

zeroedin-bill:
Agree - given 2.3 is EOL, and 2.4 is in support, it makes sense to release this with 2.4+

chrisguitarguy:
Migrations doesn't require the ORM, would it be better to just check the ORM version and throw an exception explaining that ORM 2.4+ is required here?

mikeSimonson:
@Ocramius and @zeroedin-bill Well the problem is that it would break all the symfony 2.3 app that will get updated because there is no stable tag yet. Wouldn't it ?

Ocramius:
@chrisguitarguy `"conflict": {"doctrine/orm": "<2.4"}`

Ocramius:
> that will get updated because there is no stable tag yet. Wouldn't it ?

The line above prevents the upgrade, pinpointing to the latest tag that allows ORM 2.3

chrisguitarguy:
@Ocramius TIL!

mikeSimonson:
@chrisguitarguy I am thinking that we should have a 1.0 with your previous fix that checked for both objects, and then for 1.1 have the solution we have now were we lock on 2.4.

What do you think ?

My problem is that now all the people that use symfony 2.3 will have a broken doctrine migration and it's still suported until may 2016.

chrisguitarguy:
Symfony Standard 2.3 supports [Doctrine ORM 2.4](https://github.com/symfony/symfony-standard/blob/v2.3.28/composer.json#L12). So does the Symfony [doctrine bridge](https://github.com/symfony/symfony/blob/v2.3.28/src/Symfony/Bridge/Doctrine/composer.json#L34).

I think it's pretty safe to just require folks to upgrade from the ORM 2.2 or 2.3 to at least 2.4. The upgrade doesn't look [too bad](https://github.com/doctrine/doctrine2/blob/master/UPGRADE.md#upgrade-to-24).

mikeSimonson:
@chrisguitarguy okay I just saw the original composer.json that marked doctrine <2.4@dev.

PR #238, opened 2015-05-02
Allow Migrations to Live in Multiple Directories

### LINKED ISSUE #204: Be able to register migrations from multiple folders

chrisguitarguy:
Right now migrations only supports a single directory. This changes that behavior so there can be migrations inside subdirectories.

The same logic is still used to name and version files as before: `VersionYYYYMMDDHHMMSS.php` is still required.

Closes #204

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
@@ -23,8 +23,8 @@ use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Migrations\MigrationException;
 use Doctrine\DBAL\Migrations\OutputWriter;
 use Doctrine\DBAL\Migrations\Version;
-use Doctrine\DBAL\Migrations\Finder\GlobFinder;
 use Doctrine\DBAL\Migrations\Finder\MigrationFinder;
+use Doctrine\DBAL\Migrations\Finder\RecursiveRegexFinder;
 use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Types\Type;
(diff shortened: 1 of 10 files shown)


mikeSimonson:
@chrisguitarguy nice work thanks

mikeSimonson:
@chrisguitarguy There is one little problem with your last commit.

You are assuming that the version name can only be on the form Version"Datetime".php but in fact it can be anything right now and some people use that possibility so we can't really remove it now.
(btw this cause other ordering problems with the prev and next functionnality but I am working on a patch for those).

I think that your regex should be change to /Version.{1,255}\.php$/i, as the 255 limit is the only one that is currently enforced by the field size in the database.

Thanks for the PR

chrisguitarguy:
@mikeSimonson Checkout 17f7f4a5a34548aee2e687d6b642f97eb6bf0a41

I think it makes sense -- had to change the way the `RegexIterator` worked.

mikeSimonson:
We need to update the documentation too.

chrisguitarguy:
@mikeSimonson I can do that tonight or tomorrow if you'd like. Just open an issue over there and @ reply me (or assign me) so I remember.

TomasVotruba:
Awesome work! Thanks

TomasVotruba:
This should be documented somewhere, took me while to find this feature.

RoestVrijStaal:
How to use this? It's not documented: https://www.doctrine-project.org/projects/doctrine-migrations/en/latest/reference/configuration.html

jwage:
No new configuration is required. This PR just allows you to have sub directories inside your configured directory. Previously it would not look for migrations recursively.

precious-coder:
@jwage Ok, I suspect now that full directories trees are scanned for Migrations.

I've migrations in my `<project-dir>src/Migrations` and in `<project-dir>/precious/very-barebone/src/Migrations`. 
Does this PR would work in my situation?

I prefer to whitelist directories which I want to be scanned. Adding my complete project folder to scan for Migrations is not efficient.

jwage:
No. Currently you can only provide 1 directory to start from when looking for migrations. #805 would allow that.

PR #240, opened 2015-05-04
Don't Allow ORM Verions Older than 2.4

chrisguitarguy:
Prevents some fatal errors in `OrmSchemaProvider`.

See https://github.com/doctrine/migrations/pull/236#issuecomment-98735529


TomasVotruba:
:+1:

Issue #61, opened 2012-02-28
migrations:migrate doesn't cross-check all file migration versions with DB entries


Gamesh:
the problem arises when multiple developers work with migrations and in one case migrations could be skipped:

1. Developer A creates migration 20120228123443 and runs it against local DB. 
2. In the mean time unaware of this migration developer B creates  migration 20120228114838 runs it against local DB and pushes it to the central repository.
3. Developer A pulls changes, receives migration 20120228114838, runs doctrine.php migrations:migrate.

The 20120228114838 is skipped because last migration in DB is newer then the one received, and since a full cross-ref with DB is not run, the migration does not get executed.

mikeSimonson:
It's also not super consistent with the behaviour of doctrine migration when the migration file name is not a datetime. When it's not a datetime the migration gets executed even if it come earlier in the ordering.

chrisguitarguy:
I spent some time trying to reproduce this tonight without much luck. Here's a test case that does what the issue describes: https://github.com/chrisguitarguy/migrations/commit/a68b6017d861895bc9152f0fa363e88bea22cb80

When you add a migration older than the current version in the database, it gets run and has [since the first commit](https://github.com/doctrine/migrations/commit/8f198ef623051a5eeb21d93bfc7a8a49efbe9d63#diff-dbe4e88a1dedfee1368f036832ec5c7dR338).

I also...

1. Spun up an entirely new application with just migrations and symfony console + yaml as dependencies
2. Added a migration and ran `migrations:migrate`
3. Added a migration with an older timestamp and ran `migrations:migrate` again

Both runs of `migrations:migrate` ran one migration as expected.

chrisguitarguy:
Here's a repo with the example: https://github.com/chrisguitarguy/doctrine-migrations-issue-61

```
composer install
git checkout HEAD^
./console migrations:migrate
git checkout master
./console migrations:migrate
```

The last command will execute an older migration despite the newer migration being in the database.

mikeSimonson:
@chrisguitarguy Can you make a PR with that test case ?

mikeSimonson:
@Gamesh Can you confirm that you still have or the problem or that it's fixed ?
Thanks

Gamesh:
@mikeSimonson I'll take a look, but it's a really old issue and i moved to Liquibase since then

TomasVotruba:
@chrisguitarguy Great job! 

@Gamesh Thanks for update.

I think this cane be closed then.

[...]

PR #244, opened 2015-05-11
Add a Test Case to Verify that Older Migrations are Run When a Newer Version has Already Been Run

chrisguitarguy:
See #61


mikeSimonson:
Thanks @chrisguitarguy

PR #253, opened 2015-05-20
use question helper if available, otherwise fallback to dialog helper

BreiteSeite:
This PR fixes doctrine/DoctrineORMModule#361 in a backwards compatible way.

Related PRs: #193 #198

This PR has a dependency to [PR #405 of doctrine/DoctrineORMModule](https://github.com/doctrine/DoctrineORMModule/pull/405).

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/ExecuteCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/ExecuteCommand.php
@@ -86,7 +86,13 @@ EOT
             $version->writeSqlFile($path, $direction);
         } else {
             if ($input->isInteractive()) {
-                $execute = $this->getHelper('dialog')->askConfirmation($output, '<question>WARNING! You are about to execute a database migration that could result in schema changes and data lost. Are you sure you wish to continue? (y/n)</question>', false);
+                $questionText = 'WARNING! You are about to execute a database migration that could result in schema changes and data lost. Are you sure you wish to continue? (y/n)';
+                if ($this->getHelperSet()->has('question')) {
+                    $question = new \Symfony\Component\Console\Question\ConfirmationQuestion($questionText);
+                    $execute = $this->getHelper('question')->ask($input, $output, $question);
+                } else {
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Migrations/Tools/Console/Command/ExecuteCommand.php
```diff
@@ -86,7 +86,13 @@ public function execute(InputInterface $input, OutputInterface $output)
             $version->writeSqlFile($path, $direction);
         } else {
             if ($input->isInteractive()) {
-                $execute = $this->getHelper('dialog')->askConfirmation($output, '<question>WARNING! You are about to execute a database migration that could result in schema changes and data lost. Are you sure you wish to continue? (y/n)</question>', false);
+                $questionText = 'WARNING! You are about to execute a database migration that could result in schema changes and data lost. Are you sure you wish to continue? (y/n)';
+                if ($this->getHelperSet()->has('question')) {
+                    $question = new \Symfony\Component\Console\Question\ConfirmationQuestion($questionText);
```
stof: please add a use statement for that

BreiteSeite: I didn't because this would break symfony 2.3 (QuestionHelper is not available there) based on your comment here: https://github.com/doctrine/migrations/pull/198#issuecomment-99458078

mikeSimonson:
Thanks @BreiteSeite 

@stof 
Is it possible to test that case by adding an environnement in travis with something like that ?
```yml
matrix:
  include:
    - php: 5.3
      env: deps=low

install:
  - if [ "$deps" = "no" ]; then composer install; fi;
  - if [ "$deps" = "low" ]; then composer --prefer-lowest --prefer-stable update; fi;
```

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/ExecuteCommand.php
stof: no it would not. Use statements don't trigger anything. they are resolved at compile time anyway.

PR #256, opened 2015-05-21
 Finishing the support of the questionHelper

mikeSimonson:
Symfony/console has been bumped because of a problem with the helperSet

Every question has been refactored in a method of the AbsctractCommand

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php
@@ -117,13 +117,8 @@ EOT
             }
 
             if (! $noInteraction) {
-                $questionText = 'Are you sure you wish to continue? (y/n)';
-                if ($this->getHelperSet()->has('question')) {
-                    $question = new \Symfony\Component\Console\Question\ConfirmationQuestion($questionText);
-                    $confirmation = $this->getHelper('question')->ask($input, $output, $question);
-                } else {
-                    $confirmation = $this->getHelper('dialog')->askConfirmation($output, '<question>' .  $questionText . '</question>', false);
-                }
+                $question = 'Are you sure you wish to continue? (y/n)';
+                $confirmation = $this->askConfirmation($question, $input, $output);
 
                 if (! $confirmation) {
                     $output->writeln('<error>Migration cancelled!</error>');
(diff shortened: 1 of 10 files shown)


lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php
```diff
@@ -117,7 +117,9 @@ public function execute(InputInterface $input, OutputInterface $output)
             }
 
             if (! $noInteraction) {
-                $confirmation = $this->getHelper('dialog')->askConfirmation($output, '<question>Are you sure you wish to continue? (y/n)</question>', false);
+                $question = 'Are you sure you wish to continue? (y/n)';
```
stof: IMO, you don't need the ``$question`` variable here

tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/AbstractCommandTest.php
```diff
@@ -192,6 +194,69 @@ public function testInjectedConfigurationIsPreferedOverConfigFileIsCurrentWorkin
...
+        $this->assertEquals(false, $this->invokeAbstractCommandConfirmation($input, $helper, "n"));
+
+        if (class_exists("Symfony\\Component\\Console\\Helper\\QuestionHelper")) {
+            $helper = new QuestionHelper();
+            $this->assertEquals(true, $this->invokeAbstractCommandConfirmation($input, $helper));
+            $this->assertEquals(false, $this->invokeAbstractCommandConfirmation($input, $helper, "n"));
+        }
+    }
+
+    protected function getInputStream($input)
```
stof: please make it private

stof:
@mikeSimonson Are you the coveralls admin for this repo ? If yes, please change the config so that it does not post a comment each time they make a build (we already have the PR status displaying the info in a way avoiding to spam us)

mikeSimonson:
@stof I think I fixed it

mikeSimonson:
@stof Do you know why it failed with the deps low while the package symfony/yaml has been installed.

Says ```PHP Fatal error:  Class 'Symfony\Component\Yaml\Yaml' not found```

stof:
@mikeSimonson you should not allow 2.0 versions for the Symfony composer deps IMO. The 2.0 releases were for the pre-composer area, and the autoloading metadata were wrong for several 2.0.x releases until we got Composer right.

TomasVotruba:
@stof What version do you suggest?

chrisguitarguy:
2.3+ I would say, since that's the most recent LTS release and still has support.

stof:
Yeah, either use ``~2.1`` or ``~2.3``. I suggest using ``~2.3`` as nobody should be using 2.1 or 2.2 anymore

TomasVotruba:
`2.3` for me as well

stof:
@mikeSimonson looks like the coveralls issue is not fixed

stof:
and looks like one of the test requires a newer version of Console

stof:
BufferedOutput was added in Console 2.4

mikeSimonson:
@stof 
everything is unchecked now.

Is it possible to require 2.3 and require-dev 2.4 ?

stof:
@mikeSimonson still not good for coveralls...

Regarding the dev dependency on a different version, it is a bad idea. Maybe you could change the tests to use a StreamOutput on a memory stream instead.

mikeSimonson:
@stof Thanks for the info on the StreamOutput.

I deleted the coveralls stuff as the configuration didn't seemed to work.

PR #267, opened 2015-06-14
small fixes

mikeSimonson:
Changing the name of some interface so that they have interface in their name.
Adding a new line at the end of some file that where missing it.
Removing some unused use.


stof:
@mikeSimonson Doctrine generally does not use the Interface suffix (it does for EntityManagerInterface because the interface was introduced only years after the EntityManager name was taken for the class)

mikeSimonson:
@stof I guess, I need a new major before to change that no ?

stof:
@mikeSimonson yeah, so it is not worth changing it for this interface (too bad I haven't reviewed this PR when it was done). and btw, it was probably the reason of the original naming of this interface.
But you should keep this in mind for future interfaces added in the library.

PR #279, opened 2015-06-23
use 24-hour format for migration datetime displaying

smatyas:
The current datetime formatting of migration versions is misleading, we should either display am/pm or just use the 24-hour format (which this PR solves).

I.e. the output now looks like this:
```
>> 2015-06-22 03:55:27 (20150622155527)
```
while it sohuld look like this:
```
>> 2015-06-22 15:55:27 (20150622155527)
```

This PR solves this formatting.


mikeSimonson:
Thanks @smatyas

PR #282, opened 2015-06-24
Version : use exec instead of executeQuery when no parameter is bound

ecolinet:
Is there a reason for using prepared statements instead of executing the raw query when there is no parameters bound ?

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Version.php
+++ b/lib/Doctrine/DBAL/Migrations/Version.php
@@ -268,7 +268,7 @@ class Version
                         $queryStart = microtime(true);
                         if ( ! isset($this->params[$key])) {
                             $this->outputWriter->write('     <comment>-></comment> ' . $query);
-                            $this->connection->executeQuery($query);
+                            $this->connection->exec($query);
                         } else {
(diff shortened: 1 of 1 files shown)


mikeSimonson:
Not that I know about. Just that exec seems to be absent from the documentation.

c960657:
If you add a statement that returns some result rows, the migration will die with the following error:

> Migration 20151208110928 failed during Post-Checks. Error An exception occurred while executing 'INSERT INTO doctrine_migration_versions (version) VALUES (?)' with params ["20151208110928"]:
>
> SQLSTATE[HY000]: General error: 2014 Cannot execute queries while other unbuffered queries are active.  Consider using PDOStatement::fetchAll().  Alternatively, if your code is only ever going to run against mysql, you may enable query buffering by setting the PDO::MYSQL_ATTR_USE_BUFFERED_QUERY attribute.

An example is `OPTIMIZE TABLE foo`. It returns some result rows with info about the optimization.

You could argue that such statements do not belong in a migration at all, but if we can support them with very little effort, I think we should allow this flexibility.

mikeSimonson:
I also have had problems on some setup were that method would not throw an exception in case of issue (specificaly with mysqli). But I have not been able to write a failing test for this case.

c960657:
Reverted in #388.

PR #300, opened 2015-06-27
Making sure that we only require the orm 2.4 for now

mikeSimonson:


lib/Doctrine/DBAL/Migrations/Provider/OrmSchemaProvider.php
```diff
@@ -53,4 +61,22 @@ public function createSchema()
...
+     * library allows those older version we need to be able to check for those
+     * old ORM versions. Hence the helper method.
+     *
+     * No need to check to see if EntityManagerInterface exists first here, PHP
+     * doesn't care.
+     *
+     * @param   mixed $manager Hopefully an entity manager, but it may be anything
+     * @return  boolean
+     */
+    private static function isEntityManager($manager)
```
stof: why making it static ? You call it not statically

mikeSimonson: It slipped though, How come that it even worked like that ?

PR #303, opened 2015-06-29
Move exception classes into separate directory

neeckeloo:


stof:
This is a BC break (it will break any existing generated migration using them, and any code catching them).

So -1 for this.

neeckeloo:
The change below left me thinking we could afford to break compatibility in alpha releases.

https://github.com/doctrine/migrations/commit/0e60856a10e8b510daa612fe25f6245aece77e68

stof:
Well, the change in https://github.com/doctrine/migrations/commit/0e60856a10e8b510daa612fe25f6245aece77e68 only impacts people using an unmaintained PHP version.

This change will impact everyone using the library. and people are using it in production since years, even though we never released it in non-alpha. So IMO, we should already care about BC (and we should release it as stable IMO)

neeckeloo:
I understand and this is not an essential change either.

Issue #308, opened 2015-07-13
Support of PHP >=5.3.2


robertfausk:
As of https://github.com/doctrine/migrations/commit/0e60856a10e8b510daa612fe25f6245aece77e68 PHP 5.3.2 support was skipped.

It seems it's only skipped because of the array definition in some test classes.

What about not dropping support for PHP 5.3 (until some features of PHP >=5.4 are really needed)?
I think many people still depends on it.

mikeSimonson:
@robertfausk Why should I bother supporting an outdated and unsupported version of PHP ?

robertfausk:
@mikeSimonson Yeah. I know it's outdated and not supported anymore. But here are some reasons why you should bother:

1. Because many sites still depends on PHP 5.3 and they won't be able to use doctrine/migrations in future.
see: http://w3techs.com/technologies/details/pl-php/5/all
"Version 5.3 is used by 40.2% of all the websites who use PHP version 5." (W3Techs.com, 13 July 2015)

2. There is no feature in this library which really relies on features of PHP 5.4 or higher.
3. You have very low impact to ensure compatibility.

mikeSimonson:
Well I would use that article
<http://blog.ircmaxell.com/2014/12/on-php-version-requirements.html> from
Antony Ferrara to not depend on 5.3.
The problem with your logic is that then eveyone has the impression that
it's okay to stay with outdated and unpatched version of PHP.
It's not.

robertfausk:
Yes. According to the article you posted you're right for large frameworks like wordpress, codeigniter, zend, symfony (e.g. symfony 2.7 which depends on PHP 5.3 is supported until 2018 http://symfony.com/doc/current/contributing/community/releases.html#schedule) etc.

I'm totally with you to upgrade the minimum supported PHP version IF there are PHP features you're in need of.
As for myself I wouldn't upgrade the minimum supported PHP version of my lib if just for the use of short array syntax.

Unfortunately there are companies out there which still relies on PHP 5.3 like the one I'm working at ;) and I think we are not alone out there.
We plan to upgrade to a new PHP version in Q3 2015 but thats not now...
Furthermore we are using this lib for our CI and deployment process and there is no possibility to use a PHP 5.3 compatible version than using an own fork.

What about adding a 5.3 compatitiblity branch?

If you still don't agree feel free to close this issue and according pr. I will pragmatically use my own fork then until we upgrade our PHP version and then probably switch back to this repository.

[...]

PR #309, opened 2015-07-13
readd PHP 5.3.2 support

robertfausk:
see #308 for discussion


mikeSimonson:
Sorry but as I said I don't want the burden of old and unsupported version of php

PR #310, opened 2015-07-16
Adding new configuration-option "organize-migrations"

SchulteMarkus:
This PR introduces an new configuration-option "organize-migrations", which can be set to "year" or "year_and_month". If set, the migration-classes (files) will be in appropriate sub-directories relative to "migrations-directory". For example, if you set "organize-migrations" to "year" in your doctrine-migration configuration, and generate a new migration, the new file will be placed in "migrations-directory"/2015/ (if the actual year is 2015).

This change is use-case-driven. When using doctrine-migration, having every version (98 by now) in the "migrations-directory" became ugly.


SchulteMarkus:
fyi: I will be on (long) holidays from this Thursday, 2015-07-23.

TheRatG:
:+1:

dmkuznetsov:
Good job :+1:

PR #311, opened 2015-07-21
Put doctrine-migrations into including project's vendor/bin directory

dotEvan:
Enables a project to include doctrine/migrations and then to run any migration commands by running

```
./vendor/bin/doctrine-migrations
```

from the main project's directory.


mikeSimonson:
Thanks @dotEvan .

Don't we need to set as executable only the doctrine-migrations file and not the php one ?

composer.json
```diff
@@ -35,5 +35,9 @@
         "branch-alias": {
             "dev-master": "1.0.x-dev"
         }
-    }
+    },
+    "bin": [
+        "bin/doctrine-migrations",
+        "bin/doctrine-migrations.php"
```
stof: `` bin/doctrine-migrations.php`` should not be there. Only the actual bin is needed

stof:
@mikeSimonson all files listed in the ``bin`` section of composer must be executable (otherwise Composer will make them executable and then the next update will tell you that the package has some changes).
But the mistake is listing the PHP file in this section in the first place

composer.json
dotEvan: @stof To be honest, I only use the version without the .php. However, I was being consistent with the way this was done in https://github.com/doctrine/doctrine2/blob/master/composer.json

I didn't want to assume that people didn't also want to be able to do
```
php ./vendor/bin/doctrine-migrations.php
```
if they wanted to.

n-couet:
+1

composer.json
stof: And this is a mistake in the ORM (it was reverted in DBAL but I missed the fact that the ORM included it too)

mikeSimonson:
@dotEvan Can you make that change in your PR ?

Revert the change of permission on bin/doctrine-migrations.php and remove bin/doctrine-migrations.php from the bin section of the composer.json ?

dotEvan:
@mikeSimonson Done.

mikeSimonson:
@dotEvan Thanks

PR #314, opened 2015-07-31
Ability to set configuration via ConfigurationHelper in HelperSet

DavidHavl:
Added ability to set configuration via ConfigurationHelper in HelperSet.
I.e. from cli-config.php as such: 

```
$configFile = '/settings/config.yml';
$configuration = new \Doctrine\DBAL\Migrations\Configuration\YamlConfiguration($dbConnection);
$configuration->load($configFile);
$configHelper = new \Doctrine\DBAL\Migrations\Tools\Console\Helper\ConfigurationHelper($dbConnection, $configuration);
$helperSet->set($configHelper, 'configuration');
```

The same idea (but for connection) can be seen here: http://davidhavl.com/devnotes/2015/07/setting-up-standalone-doctrine-migrations/

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/AbstractCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/AbstractCommand.php
@@ -100,7 +100,11 @@ abstract class AbstractCommand extends Command
     protected function getMigrationConfiguration(InputInterface $input, OutputInterface $output)
     {
         if (!$this->migrationConfiguration) {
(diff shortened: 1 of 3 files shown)


mikeSimonson:
@DavidHavl Can you add some tests for that new use case ?

DavidHavl:
@mikeSimonson Sure, no problem, will do within next few days.

DavidHavl:
@mikeSimonson The test for this use case has now been added.

mikeSimonson:
@DavidHavl Thanks

PR #321, opened 2015-08-02
Avoid uploading code coverage to Scrutinizer for HHVM and PHP 7

stof:
PHP 7 and HHVM are not generating the code coverage data. Submitting to Scrutinizer in such case would notify it that there is no data, which would cancel the build in case it happens before another job succeeds and uploads the coverage.

this will avoid having all builds being cancelled on Scrutinizer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ php:
   - 5.6
   - 7.0
   - hhvm
- 
+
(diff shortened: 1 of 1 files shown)


mikeSimonson:
@stof Thanks, I was wondering were this issue was coming from.

PR #322, opened 2015-08-02
Improve the Travis configuration by persisting the composer cache

stof:
This keeps using dists when the github token is not available, relying on the fact that dists will already be cached in most cases (because of builds for the master branch) and so would still be usable, and that Composer will fallback to source install automatically in case the dist downloading fails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ php:
   - 5.6
   - 7.0
   - hhvm
- 
+
 matrix:
   allow_failures:
     - php: 7.0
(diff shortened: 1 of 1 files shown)


TomasVotruba:
:+1:

mikeSimonson:
@stof Are you sure that this would work with the PR.
I read somewhere in the doc that it was better to put --prefer-source in the PR runs because the token won't be there.
Note that this might not be the case when the cache is activated.

stof:
@mikeSimonson when the doc was written (years ago), composer was not able to fallback from source when the API rate limit was reached and it could not ask for authentication (and so was failing hard, breaking builds). But this has been implemented in Composer since a long time.
This means that a build which must download packages (because it does not have them in the cache) and cannot authenticate will simply be a bit slower by making an HTTP call first and then running ``git clone`` instead of running ``git clone`` directly. but for the majority of builds will not need to download archives because they will already be cached by previous builds (it would only happen when a package has a new release, and only for this package, not for all of them)

stof:
note that this is indeed related to the fact that we can hit the cache in most cases. Without persisting the cache, we would have cache misses all the time of course.

PR #323, opened 2015-08-02
Add a way to avoid reading the existing schema before each migration

stof:
Most migrations don't actually perform changes on the Schema object but rather add SQL queries directly in the migration, because it is much more powerful (it is the only way to migrate data for instance, as this must generally happen in the middle of schema changes) and because the
diff commands generates migrations this way.
For such migrations, the Schema object will never be altered, but the migration tool will still need to read it at the beginning of each migration to pass it to the Migration (which does not care about it). To
improve performance, migrations can now implement an optional marker interface. In such cases, the migration will receive an empty Schema object and will not be allowed to perform changes on it. This avoids the need to read the database schema all the time for nothing.

To prove the performance improvement, look at these Blackfire profiles. They are profiling the execution of ``app/console doctrine:migrations:migrate -n`` on my work project. The project currently has 55 migrations going from an empty database to the current schema (for as much as it matters, this is running on the PostgreSQL platform, but similar results would be visible with other platforms).

- before the change (running with the librart at 1c7d88ae519361761fad42c03707ca8e8e37cd53): https://blackfire.io/profiles/c5b7813b-305d-4b8f-b1fa-ea63ad013adc/graph
- after the change of this PR (and implementing the marker interface on all 55 migration classes): https://blackfire.io/profiles/ab773a44-1e02-4873-9939-8021c6c253de/graph
- comparison between them: https://blackfire.io/profiles/compare/efe08c00-7a85-4ddc-a7b3-f62f8a08426d/graph

note how we are spending 72% of the time loading schemas from the DB and 22% of the time comparing them to find no changes. These nodes account for 94% of the runtime for something we don't need at all.

TODOs:

- [x] change the DiffCommand to implement the marker interface by default as these migrations don't rely on the Schema object (the GenerateCommand will not add it IMO as we don't know how the code in them will look like)
- [ ] Settle on the name of the interface in case someone has a better name
- [ ] Write some tests for this behavior
- [ ] Write some doc about this feature

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
@@ -109,7 +109,7 @@ EOT
         }
 
         $version = date('YmdHis');
-        $path = $this->generateMigration($configuration, $input, $version, $up, $down);
+        $path = $this->generateMigration($configuration, $input, $version, $up, $down, true);
 
         $output->writeln(sprintf('Generated new migration class to "<info>%s</info>" from schema differences.', $path));
     }
--- a/lib/Doctrine/DBAL/Migrations/SqlMigration.php
+++ b/lib/Doctrine/DBAL/Migrations/SqlMigration.php
@@ -1,0 +1,31 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the LGPL. For more information, see
+ * <http://www.doctrine-project.org>.
+*/
+
+namespace Doctrine\DBAL\Migrations;
+
+/**
+ * Marker interface implemented by migrations which provide SQL queries rather than using the Schema
+ *
+ * When the migration implements this interface, the runner will pass an empty Schema object to the migration
+ * instead of reading the current state of the database, and it will forbid making changes on the Schema itself.
+ * This improves performance when the migration does not rely on the schema
+ */
+interface SqlMigration
+{
+}
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
@@ -42,12 +42,13 @@ class GenerateCommand extends AbstractCommand
 namespace <namespace>;
 
 use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Migrations\SqlMigration;
 use Doctrine\DBAL\Schema\Schema;
 
 /**
  * Auto-generated Migration: Please modify to your needs!
  */
-class Version<version> extends AbstractMigration
+class Version<version> extends AbstractMigration<interfaces>
 {
     /**
      * @param Schema $schema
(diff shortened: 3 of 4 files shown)


stof:
this is an optimization I have in mind since several months actually. Given it took me approximately 5 minutes to implement, it is a shame I haven't done it previously. I would already have saved much more than that just for migration runs on my work project (and even more time in Travis builds which always run migrations from 0 to the current one while I only run a single migration locally in general) :smile: 

@mikeSimonson tell me if you are interested in merging this feature (I guess I now the answer though) so that I finish it.

stof:
regarding the scrutinizer red status, it will be solved by #321

jwage:
That is great! :+1:

johnkary:
Just wanted to say our team only does SQL migrations. Running only SQL migrations as the default strategy will work great for us.

stof:
@johnkary given that the diff command generates SQL migrations, and that they are the only way to migrate data, I guess almost everyone uses them. This is why I changed the behavior of the diff command to implement the interface.
I kept this behavior opt-in though, for BC reasons.

beberlei:
We had the plan to actually remove the schema from the signature for a very long time now, but it would the worst possible BC break.

I am unsure about the marker interface, it is a bit of a hack, how about we introduce a Proxy object for ``Schema`` instead that loads the actual schema lazily in the background?

stof:
@beberlei this is not possible, because we have to clone the Schema object before calling the method, to be able to keep the old schema and compare it to the version modified by the user.

And I agree that removing the Schema would be better, but it breaks BC badly and some people actually prefer using it in the DB migration, even though it leads to fragile migrations

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
```diff
@@ -109,7 +109,7 @@ public function execute(InputInterface $input, OutputInterface $output)
         }
 
         $version = date('YmdHis');
-        $path = $this->generateMigration($configuration, $input, $version, $up, $down);
+        $path = $this->generateMigration($configuration, $input, $version, $up, $down, true);
```
texdc: Please avoid [flag arguments](http://martinfowler.com/bliki/FlagArgument.html).

> "Flag arguments are ugly. Passing a boolean into a function is a truly terrible practice. It immediately complicates the signature of the method, loudly proclaiming that this function does more than one thing. It does one thing if the flag is true and another if the flag is false!"
>    -- Robert C. Martin, [Clean Code](http://amzn.to/1lPJsXv)

mikeSimonson:
@stof Thanks for the work.

beberlei:
@stof we could handle the proxies specially in migrations code so that users don't have to care.

stof:
@beberlei I don't understand what you mean

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
stof: @texdc generating a migration should rather be extracted out of the command themselves to allow cleaner logic about configuring the generated migration. But this is best done separately.
for now, this is acceptable (we have worse things in the code of the library btw, and this is an internal function)

texdc: I understand.  However, it'd be better to not add to the problems and instead use best-practices.  The concerns Fowler has about this style are valid regardless of the current state of the code.

chrisguitarguy: The concerns of Martin and Fowler apply when a routine does completely different things based on a flag:

```php
function doStuff($flag)
{
  if ($flag) {
    doSomething();
  } else {
    doSomethingCompleteDifferent();
  }
}
```

The method in question does exactly one thing (generate migrations) and the flag is appropriate for configuration. It can be made more reasonable with a constant if it's really a concern.

```php
$this->generateMigration(..., self::SQL_ONLY);
```

texdc: @chrisguitarguy see Version::execute @ L252 (below)

chrisguitarguy: Totally separate issue from this method, but again it's doing exactly one thing (executing a migration). It would be nicer, IMO, to just give the migration itself a method that turned the migration into executable SQL statements. But that's a pretty major refactor that doesn't make a lot of sense with the current system and would cause a lot of BC breaks.

Idealism is great, but not practical in this case. It's not an idea solution, but it is a solution that would provide a benefit for a lot of people.

stof: @texdc executing migrations is totally different than generating them

bretrzaun:
I was trying to use Doctrine Migrations (instead of legacy selfmade migration tool) within an application with a medium-to-large schema (~800 schema objects).

Unfortunately applying just one migration takes very long time - too long to be usable.
So I am hoping for this PR to bring a solution.

PS: We are using SQL-only migrations.

mikeSimonson:
@bretrzaun A very quick fix for your speed issue that might work (need to check) is that in the connection object that you pass to doctrine/migrations you set a schema filter that exclude all tables.
It worked for me when I had to work with tables that are not mappable by doctrine because of the lack of primary key.

stof:
@mikeSimonson this might be hard if the connection used is the same connection is used by the ORM as that would break the diff command

stof:
hmm, it would even break it all the time. The existing DB schema is loaded by the connection passed to migrations

mikeSimonson:
To make it work I just created another connection with the name migration that had all the same configuration except for that filter. Obviously it breaks the diff command for that connection. I was generating the diff with the "normal" slow connection and then running the migration with the connection that excluded everything.

bretrzaun:
@mikeSimonson Great tip - thank you very much - works perfectly !

Maff-:
We are having the same performance issue with a growing Symfony application. Currently executing the ~180 migrations during CI takes about six minutes.

It would be great if there was a solution for this. This PR seems a nice way, is it still in development and/or under consideration? I would be happy to help if that's useful.

@mikeSimonson @bretrzaun: Does either of you have an example config for a Symfony project to get this workaround your are talking about working. I'm not having any luck with this.

stof:
@Maff- we are investigating an alternate solution based on proxies. If it does not work, I will update this PR.

stof:
Closing this as #407 has been merged instead, being even better.

PR #324, opened 2015-08-03
Added ability to load config from php array or json files.

DavidHavl:
Added ability to load config from php array or json files.


mikeSimonson:
@DavidHavl Thanks

PR #328, opened 2015-08-04
Switch unexisting <warning> tag to <comment>

bamarni:
It's outputting the tag as such, switching to <comment> turns it to yellow.


stof:
:+1: ``<warning>`` is not an existing style by default in the Symfony Console

PR #333, opened 2015-08-07
add the ability to auto create migrations in a folder by month and year

mikeSimonson:


lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
```diff
@@ -119,10 +146,29 @@ public function __construct(Connection $connection, OutputWriter $outputWriter =
...
+        if ($finder === null) {
+            $finder = new RecursiveRegexFinder();
+        }
         $this->migrationFinder = $finder;
     }
 
     /**
+     * @return bool
+     */
+    public function areMigrationsOrganizedByYear()
```
mikeSimonson: @stof What do you think about the naming ? Shouldn't that be is ?

mikeSimonson: @Ocramius Do you think the naming is an issue here ?

Ocramius: Naming seems clear to me here

PR #334, opened 2015-08-10
Fix: Version should probably be 1.0.0

localheinz:
This PR

* [x] fixes a thing where the version should be `v1.0.0` instead of `v1.0.0-alpha3`

:information_desk_person: I heard we're stable now, aren't we`


stof:
thanks

PR #339, opened 2015-08-18
Fix: Undefined class

localheinz:
This PR

* [x] prepends a slash to fix a thing where `InvalidArgumentException` is an undefined class in the current namespace


mikeSimonson:
@localheinz thanks

localheinz:
@mikeSimonson 

:+1:  Sure, no problem!

PR #340, opened 2015-08-18
Fix: Type-hint against OutputInterface

localheinz:
This PR

* [x] type-hints against `OutputInterface` as the private method `showVersions()` is always invoked with an instance of `OutputInterface`


mikeSimonson:
@localheinz thanks

PR #341, opened 2015-08-18
Fix: Return early

localheinz:
This PR

* [x] returns early, and makes the happy path more obvious


mikeSimonson:
@localheinz thanks, all those little PR where super easy to review and very helpfull.

localheinz:
@mikeSimonson 

I think everyone likes small PRs, no? :wink:

PR #342, opened 2015-08-18
Enhancement: Allow to set OutputWriter

localheinz:
This PR

* [x] asserts that an instance of `OutputWriter` can be injected via constructor
* [x] asserts that an instance of `OutputWriter` is created otherwise
* [x] asserts that an instance of `OutputWriter` can be injected via mutator

### Use Case

When fetching a migration configuration from a container, you don't have an `OutputWriter` yet, you don't even have an `Output` yet (you may not even need it). So, wiring up both `OutputWriter` and `Output` in the container seems like overkill.

Being able to set the `OutputWriter`, however, when you need it, would be nice.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
@@ -204,6 +204,16 @@ class Configuration
     }
 
     /**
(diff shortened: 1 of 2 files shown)


localheinz:
Thank you, @mikeSimonson!

PR #343, opened 2015-08-18
Scrutinizer Auto-Fixes

scrutinizer-auto-fixer:
@mikeSimonson requested this pull request.

It consists of patches automatically generated for this project on Scrutinizer:
https://scrutinizer-ci.com/g/doctrine/migrations/

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
@@ -43,7 +43,7 @@ class DiffCommand extends GenerateCommand
      */
     protected $schemaProvider;
 
-    public function __construct(SchemaProviderInterface $schemaProvider=null)
+    public function __construct(SchemaProviderInterface $schemaProvider = null)
     {
         $this->schemaProvider = $schemaProvider;
         parent::__construct();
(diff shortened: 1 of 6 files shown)


lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
```diff
@@ -90,10 +90,10 @@ public function execute(InputInterface $input, OutputInterface $output)
         $toSchema = $this->getSchemaProvider()->createSchema();
 
         //Not using value from options, because filters can be set from config.yml
-        if ( ! $isDbalOld && $filterExpr = $conn->getConfiguration()->getFilterSchemaAssetsExpression()) {
+        if (!$isDbalOld && $filterExpr = $conn->getConfiguration()->getFilterSchemaAssetsExpression()) {
```
stof: this change does not match the Doctrine coding standards.

I would be totally in favor of using PSR-2 compatible CS rules, but this should be an educated decision.

mikeSimonson: @stof Where should we talk about that ?

stof: Given you are the lead of the project, I think the decision is yours, but you can probably discuss it with other doctrine core devs:

- either follow the PSR-2 rule
- or follow @guilhermeblanco's preference (or maybe @beberlei's one, I don't remember) which is part of the Doctrine coding standards (this is the only place where they are incompatible with PSR-2).

We just need to be consistent inside the package.

For me, it is a :+1: to be PSR-2 compatible (I've been asking for this for years, approximately since PSR-2 went out).

PR #345, opened 2015-08-18
Enhancement: Extract constants for directions in which Versions can be executed

localheinz:
This PR

* [x] asserts the values of constants for `up` and `down`
* [x] creates these constants in `Version`
* [x] makes use of these constants

:information_desk_person: Magic strings and numbers, we can avoid them, can't we?


mikeSimonson:
@localheinz thanks

PR #347, opened 2015-08-24
Fix for migrations with no number

liquorvicar:
The issue here was the type juggling in the getRelativeVersion() method. Originally, it was not doing a strict search and thus a non-numeric string was matching the zero value. When I changed it to a strict search, I had to rebuild the versions array as numeric strings are converted to ints when used as array keys. This now works and all the tests pass (including the new one I added for the use case that was failing). I've fixed one little missing docBlock and also added a test for this use case to check it loads the Migrations OK (it worked out of the box and didn't require any code changes).

Let me know if there's anything else I need to do to get the PR merged.


mikeSimonson:
@liquorvicar Thanks, looks perfect at first glance.
But given the sensibility of that part of the code I want to wait when I have enough time to review it before to merge it.

liquorvicar:
No worries. All the existing tests still pass, but that does assume the existing tests cover all possible use cases (which is probably unlikely!).

mikeSimonson:
#354 Also try to fix partly the same issue.

Right now I am kinda leaning to change the documentation and drop the 0 feature as we have the first alias but I still need to check that they are equivalent.

dotEvan:
@mikeSimonson FWIW, I don't think that "first" implies the same as zero. When I migrate to 0, I expect that all migrations are rolled back to the starting point. If only migrations were used to bootstrap the database, I should end up with an empty database with only the doctrine_migrations table that is empty as well.

Conversely, when I think I am migrating to the first migration, I feel like that would actually be the state where all migrations are rolled back *except* the first one. In this case, I would end up with whatever changes exist in the up() migration in the earliest migration that exists.

Semantics, yes. Probably something that could be explained with documentation. Nevertheless, it is certainly something that is quite opaque right now.

liquorvicar:
@mikeSimonson Allowing Versions without a number in the name was pretty much the whole point of the PR. Our use case is to have incrementing Version numbers with a human-readable description in the name, e.g. Version001InitialSchema, Version002NewFeature, Version003PaypalIntegration etc and then to have versions with no numbers for WIP migrations (i.e. during dev). A git hook would then either check the Migrations on merge and fail if there are non-numeric Versions, or perhaps better rename them and give them the next number in sequence. I don't think this causes a problem with ordering as Versions without a numeric part will always come after those that do have a numeric element and in the workflow described it would be rare to have more than one WIP migration. If there were two migrations without a numeric part then they would be ordered alphabetically. This could cause an issue if the user wasn't aware of this but as this would be advanced usage, this could be covered with documentation.

You say "Right now doctrine migrations count on the fact that there will be a part of the name of the migration that will be interpretable as a number" but that's not quite true, in that doctrine migrations allow you to add a migration without a number currently and it all works fine apart from this issue I've raised. If you're going to disallow migrations without a numeric part, then surely that is potentially breaking BC. Better to fix this issue and document it (and possibly even deprecate that behaviour if you really don't want to go down that route), surely?

We haven't set our workflow in stone yet so going down the route of not allowing a non-numeric version name wouldn't necessarily be a huge issue for us.

Let me know if you have any further thoughts.

mikeSimonson:
@liquorvicar I am closing this PR as I don't think that having migration without number is the right way forward. It only add complication if there are more than one of those migration. It's much more easier to have people agree upfront on the organisation of the numbering of the migration.

Thanks for your time

PR #353, opened 2015-09-03
Conditional internal migrations steps

Voronenko:
Possibility to have internal migration steps that depend on database state

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/AbstractMigration.php
+++ b/lib/Doctrine/DBAL/Migrations/AbstractMigration.php
@@ -173,6 +173,11 @@ abstract class AbstractMigration
         $this->version->addSql($sql, $params, $types);
     }
 
+    protected function checkResultExists($sql, array $params = array(), array $types = array())
+    {
+      return $this->version->checkResultExists($sql, $params, $types);
+    }        
+
     protected function write($message)
     {
         $this->outputWriter->write($message);
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
@@ -703,6 +703,13 @@ class Configuration
         return $versions;
     }
 
+
+    public function setMappingTypes($mapping_types) {
+        foreach($mapping_types as $field=>$type) {
+            $this->connection->getDatabasePlatform()->registerDoctrineTypeMapping($field, $type);
+        }
+    }    
+
     /**
      * Find all the migrations in a given directory.
      *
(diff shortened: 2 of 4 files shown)


lib/Doctrine/DBAL/Migrations/AbstractMigration.php
```diff
@@ -173,6 +173,11 @@ protected function addSql($sql, array $params = [], array $types = [])
         $this->version->addSql($sql, $params, $types);
     }
 
+    protected function checkResultExists($sql, array $params = array(), array $types = array())
+    {
+      return $this->version->checkResultExists($sql, $params, $types);
```
stof: wrong indentation

lib/Doctrine/DBAL/Migrations/AbstractMigration.php
```diff
@@ -173,6 +173,11 @@ protected function addSql($sql, array $params = [], array $types = [])
         $this->version->addSql($sql, $params, $types);
     }
 
+    protected function checkResultExists($sql, array $params = array(), array $types = array())
```
stof: phpdoc should be added

lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
```diff
@@ -703,6 +703,13 @@ public function getMigrationsToExecute($direction, $to)
         return $versions;
     }
 
+
+    public function setMappingTypes($mapping_types) {
+        foreach($mapping_types as $field=>$type) {
+            $this->connection->getDatabasePlatform()->registerDoctrineTypeMapping($field, $type);
```
stof: changing the configuration of the connection looks wrong to me. the connection is passed as a constructor argument to this class, which means that configuring it is meant to be done outside this class

stof:
@Voronenko you could just check the Schema object passed as argument

lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
stof: and btw, this seems unrelated to the feature you are trying to add here

mikeSimonson:
@Voronenko 

There are 2 unrelated stuff in your migration.
A new method and the type mapping.
Please make 2 different pull request.

The build folder has to stay there for the tests.

I don't think that the new method checkResultExists is interesting enough so that it should be there by default. You can add it in your version of the AbstractMigration without touching the version.

If you want me to merge the type mapping pull request, you need to add tests.

lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
Voronenko: Agree. I will split into two and address comments

stof:
@mikeSimonson IMO, the type mapping stuff is wrong here. The connection is meant to be configured externally.
and if they really want to do such weird config change inside their migration, they can still do it (the connection is accessible). But we should not encourage it by adding a dedicated API to do it

PR #354, opened 2015-09-03
Allow 0 to be used as a version number.

aolalpha:
While using "first" as an alias will work, I expect that I can type in a version number as the parameter for the migration. This only came up with "Unknown version: 0".

I initially implemented this in `resolveVersionAlias()` as another alias for "first", but 0 really isn't an alias for a version - it is an actual version. I thought it made more sense to return true in `hasVersion()`. Let me know if you feel this implementation should be changed.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
@@ -439,7 +439,7 @@ class Configuration
      */
     public function hasVersion($version)
     {
-        return isset($this->migrations[$version]);
+        return isset($this->migrations[$version]) || $version === '0';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


KingCrunch:
As far as I know `0` is indeed a valid version. I can imagine, that this leads to unwanted behaviour, when someone _really_ uses a version `0` as regular version.

dotEvan:
@KingCrunch If someone created class Migration0, then that would be read into the migrations array and the first part of the hasVersion() method would return true. Wouldn't that class get executed at that point?

As it currently stands, when someone types 'first', it is aliased to 0.

I think there are also semantic differences between migrating to "first" and migrating to "0", see my thoughts in #347.

mikeSimonson:
As @KingCrunch noted this will lead to problems as 0 is used internaly by doctrine migration to be before the first migration.

@aolalpha Thanks

PR #356, opened 2015-09-15
Keep the license file in archives

stof:
Shipping the code without the license would violate the license.


.gitattributes
```diff
@@ -1,6 +1,5 @@
 build/*                 export-ignore
 tests/*                 export-ignore
-composer.lock           export-ignore
```
stof: this is duplicated

mikeSimonson:
@stof Thanks

PR #360, opened 2015-10-15
fix "all migrated versions shown as unavailable executed" issue (part…

flarelt:
…ial revert of d81cf0dd13a2fd9a835d5639b0b64a85856edf1e)


mikeSimonson:
Thanks @flarelt, can you add a test case for that fix ?

PR #362, opened 2015-10-16
support addSql() calls in postUp/postDown methods

farzadghanei:
In migration version classes It's possible to call `$this->addSql()` method (will be proxied to `Doctrine\DBAL\Migrations\Version::addSql` in preUp/preDown and `up`/`down` methods. But calling this method in `postUp`/`postDown` will have no effect right now since all the SQL commands queued to be executed are executed before the `postUp`/`postDown` methods.

This branch allows calling `addSql` in `postUp`/`postDown` by extracting the logic of executing the SQL queue from the `execute` method body into a private method, and call it both `up`/`down` as well as `postUp`/`postDown`.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Version.php
+++ b/lib/Doctrine/DBAL/Migrations/Version.php
@@ -264,42 +264,19 @@ class Version
             $this->migration->$direction($toSchema);
             $this->addSql($fromSchema->getMigrateToSql($toSchema, $this->platform));
 
+            $this->executeRegisteredSql();
+
+            $this->state = self::STATE_POST;
+            $this->migration->{'post' . ucfirst($direction)}($toSchema);
+            $this->executeRegisteredSql();
+
             if (! $dryRun) {
-                if (!empty($this->sql)) {
-                    foreach ($this->sql as $key => $query) {
-                        $queryStart = microtime(true);
-
-                        if ( ! isset($this->params[$key])) {
-                            $this->outputWriter->write('     <comment>-></comment> ' . $query);
-                            $this->connection->exec($query);
-                        } else {
-                            $this->outputWriter->write(sprintf('    <comment>-</comment> %s (with parameters)', $query));
-                            $this->connection->executeQuery($query, $this->params[$key], $this->types[$key]);
-                        }
-
-                        $this->outputQueryTime($queryStart, $timeAllQueries);
-                    }
-                } else {
-                    $this->outputWriter->write(sprintf(
-                        '<error>Migration %s was executed but did not result in any SQL statements.</error>',
-                        $this->version
-                    ));
-                }
-
                 if ($direction === self::DIRECTION_UP) {
                     $this->markMigrated();
                 } else {
                     $this->markNotMigrated();
                 }
-
-            } else {
-                foreach ($this->sql as $query) {
-                    $this->outputWriter->write('     <comment>-></comment> ' . $query);
-                }
             }
-
-            $this->state = self::STATE_POST;
-            $this->migration->{'post' . ucfirst($direction)}($toSchema);
 
             $migrationEnd = microtime(true);
             $this->time = round($migrationEnd - $migrationStart, 2);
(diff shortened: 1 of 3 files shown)


mikeSimonson:
@farzadghanei Thanks a lot. Do you think that you would have time to add a little test for this.

farzadghanei:
@mikeSimonson Sure, I took a look at current tests and added a functional test to modify the schema in `up`, and then insert some values in `postUp` using `addSql` (This is the same scenario that I wanted to have in the first place, to modify the schema in `up`, and modify the data in `postUp`).

I could have added the same test for `down` and `postDown`, however the flow does not seems natural, because most of times modifying the data should happen before migrating the schema down.

Thanks mike.

stof:
This is the cause of #367, because it now expects post* methods to register queries, which is wrong

stof:
> and modify the data in postUp

in most cases, you need to modify the data in up anyway, because it often needs to be done in the middle of the schema changes (before adding NOT NULL constraints and foreign keys for instance) rather than after

farzadghanei:
> in most cases, you need to modify the data in up anyway, because it often needs to be done in the middle of the schema changes (before adding NOT NULL constraints and foreign keys for instance) rather than after

It depends on the situation. Being able to modify the data in `up` (everywhere stated `up` should apply for `down` as well), does not mean one is not meant to modify data in `postUp`.

Before this merge request, there were 2 ways to modify the schema and the data in a migration class:

1. Use `addSql` in `up` for everything
2. Modify the schema instance in `up`, and then use `connection` in `postUp`

This merge request introduces the feature to call `addSql` in `postUp` as a 3rd option to the above.

What was the main concern for this merge request, was the fact the calling `addSql` in `postUp` had no effect at all, but it was not being disallowed.

stof:
For anyone coming there, this PR was reverted in 282f2dbdedaaf7731c308f9540a3c201141cae12

msrumon:
Isn't it even possible to use query builder to add some production data along with the migration?

```php
public function postUp(Schema $schema): void
{
    $data = ['foo', 'bar', 'baz'];

    $qb = $this->connection->createQueryBuilder();

    foreach ($data as $datum) {
        $qb->insert('table')->values(['column' => $datum]);
    }
}
```

But it doesn't seem to work. I see no data whatsoever in the table. I also tried:

```php
foreach ($data as $datum) {
    $this->connection->insert('table', ['column' => $datum]);
}
```

But it fails to set IDs automatically, throwing `SQLSTATE[23502]: Not null violation: 7 ERROR:  null value in column "id" of relation "table" violates not-null constraint` error.

> I'm using PostgreSQL database.

stof:
@msrumon your never execute the query that you build

msrumon:
Thanks. Makes sense. But what about `$this->connection->insert()`? Why does it throw not-null constraint violation error? `$qb->insert()` does the same `INSERT INTO` stuff behind the scene, right? Or is there something else happening besides that?

stof:
`$qb->insert()` does **nothing** if you don't execute the query, which is probably why you don't get any error.

regarding the error you get with `$this->connection->insert()`, I cannot help you without knowing your schema structure. your code would be valid only if the `id` field has a default value generating the id in the database when you insert without a value. My guess is that it is not how your id field is configured.

msrumon:
This is the `up()` method:

```php
public function up(Schema $schema) : void
{
    // this up() migration is auto-generated, please modify it to your needs
    $this->addSql('CREATE SEQUENCE articles_id_seq INCREMENT BY 1 MINVALUE 1 START 1');
    $this->addSql('CREATE SEQUENCE categories_id_seq INCREMENT BY 1 MINVALUE 1 START 1');
    $this->addSql('CREATE SEQUENCE status_id_seq INCREMENT BY 1 MINVALUE 1 START 1');
    $this->addSql('CREATE TABLE articles (id INT NOT NULL, category_id INT NOT NULL, author_id INT NOT NULL, status_id INT NOT NULL, title VARCHAR(255) NOT NULL, content TEXT NOT NULL, hits INT NOT NULL, created_at TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL, updated_at TIMESTAMP(0) WITHOUT TIME ZONE NOT NULL, PRIMARY KEY(id))');
    $this->addSql('CREATE INDEX IDX_BFDD316812469DE2 ON articles (category_id)');
    $this->addSql('CREATE INDEX IDX_BFDD3168F675F31B ON articles (author_id)');
    $this->addSql('CREATE INDEX IDX_BFDD31686BF700BD ON articles (status_id)');
    $this->addSql('CREATE TABLE categories (id INT NOT NULL, name VARCHAR(64) NOT NULL, PRIMARY KEY(id))');
    $this->addSql('CREATE TABLE status (id INT NOT NULL, label VARCHAR(32) NOT NULL, stat VARCHAR(16) NOT NULL, PRIMARY KEY(id))');
    $this->addSql('ALTER TABLE articles ADD CONSTRAINT FK_BFDD316812469DE2 FOREIGN KEY (category_id) REFERENCES categories (id) NOT DEFERRABLE INITIALLY IMMEDIATE');
    $this->addSql('ALTER TABLE articles ADD CONSTRAINT FK_BFDD3168F675F31B FOREIGN KEY (author_id) REFERENCES users (id) NOT DEFERRABLE INITIALLY IMMEDIATE');
    $this->addSql('ALTER TABLE articles ADD CONSTRAINT FK_BFDD31686BF700BD FOREIGN KEY (status_id) REFERENCES status (id) NOT DEFERRABLE INITIALLY IMMEDIATE');
}
```

`$this->addSql()` works. Like if I call `$this->addSql('INSERT INTO categories (name) VALUES (\'foo\')')`, then everything works. It's particularly with `$this->connection->insert()` which doesn't seem to work. I looked into the actual source code, and based on that code it should also work.

PR #366, opened 2015-10-19
WIP

mikeSimonson:


mikeSimonson:
It execute this sql stmt https://github.com/doctrine/migrations/blob/master/tests/Doctrine/DBAL/Migrations/Tests/Functional/CliTest.php#L316
And it fails at the next step when it try to regenerate the schema object with this error message.

Can cause this error Doctrine\DBAL\Migrations\Tests\Functional\CliTest::testMigrationLifecycleFromCommandLine
Doctrine\DBAL\DBALException: Unknown database type integer auto_increment requested, Doctrine\DBAL\Platforms\SqlitePlatform may not support it

This PR is a WIP refactor so that all the method that have the same signature behave the same way.
Aka preUp(Schema), up(Schema), postUp(Schema), preDown(Schema), preDown(Schema), postDown(Schema) .

stof:
your refactoring means that a migrations runs in 3 different transactions rather than a single one, which is a no-go. It makes transactional mode much less useful

mikeSimonson:
@stof Yeah, I realized that the whole idea was going in the wrong direction.

Issue #367, opened 2015-10-20
"Migration XYZ was executed but did not result in any SQL statements." is displayed


boteeka:
"Migration XYZ was executed but did not result in any SQL statements." is displayed after each migration, right after printing the SQL statements that actually have been executed.

stof:
@boteeka are you using ``->addSql()`` in your SQL migration, or are you using the connection directly ?

boteeka:
I am using `->addSql()`

stof:
Can you give the exact version of the library you are using ? Use ``composer show -i doctrine/migrations`` to get the info if you are using composer

boteeka:
That I can do only tomorrow, but I am on dev-master, so I suspect it is the head of master.

stof:
well, it depends when you locked it

stof:
the issue is a bug in #362

boteeka:
Indeed, but `composer update` has been run today, so should be the latest. But there is no point speculating on it, I'll be able to provide exact info tomorrow.

stof:
Well, as I identified the cause of the issue, I can be sure it is the latest, as the last commit is the merge of the PR

slavafomin:
Hello!

I have had the same issue today:

```
$ composer show -i doctrine/migrations
name     : doctrine/migrations
descrip. : Database Schema migrations using Doctrine DBAL
keywords : database, migrations
versions : * dev-master, 1.1.x-dev
type     : library
license  : GNU Lesser General Public License v2.1 only (LGPL-2.1) (OSI approved) https://spdx.org/licenses/LGPL-2.1.html#licenseText
source   : [git] https://github.com/doctrine/migrations.git 68cc7bf629a7c1fc0696e583cc8ed46f897d0afb
dist     : [zip] https://api.github.com/repos/doctrine/migrations/zipball/68cc7bf629a7c1fc0696e583cc8ed46f897d0afb 68cc7bf629a7c1fc0696e583cc8ed46f897d0afb
names    : doctrine/migrations

autoload
psr-4
Doctrine\DBAL\Migrations\ => lib/Doctrine/DBAL/Migrations

requires
doctrine/dbal ~2.2
php >=5.4.0
symfony/console ~2.3
symfony/yaml ~2.3

requires (dev)
doctrine/coding-standard dev-master
doctrine/orm 2.*
johnkary/phpunit-speedtrap ~1.0@dev
mockery/mockery ^0.9.4
phpunit/phpunit ~4.7
[...]

PR #368, opened 2015-10-21
Fix issue of false warning for no queries after migrations

farzadghanei:
Migrations shows a warning if not actual SQL statement is run after
migrations. After merging "add-sql" branch
(Ydbbcee1906524f4667c9db06ee9afef476fa0)
an issue is introduced for the warning kept showing.

This is because the logic of executing registered SQL statements was
moved to a private function which is called once after the
migration and once after post migration method. This logic
(`executeRegisteredSql` mehtod) used to show the warning for no SQL
statements, and when called after post migration methods it is
quite possible to show the warnings, since they might not register
any SQL statements at all.

Fixed this issue by removing showing the warning from the
`executeRegisteredSql` method to the caller (`execute` method).

Also the previous behavior of `execute` method was to return the list of
SQL statements registered, but "add-sql" branch affected this, since
we cleared out the `sql` property of the `Version` class after executing
them. Fixed this issue by keeping a local copy of the SQL statements
and return that list instead.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Version.php
+++ b/lib/Doctrine/DBAL/Migrations/Version.php
@@ -264,17 +264,27 @@ class Version
             $this->migration->$direction($toSchema);
             $this->addSql($fromSchema->getMigrateToSql($toSchema, $this->platform));
 
-            $this->executeRegisteredSql($dryRun, $timeAllQueries);
+            $allSqlRegistered = $this->sql;
+            $countSqlExecuted = $this->executeRegisteredSql($dryRun, $timeAllQueries);
 
             $this->state = self::STATE_POST;
             $this->migration->{'post' . ucfirst($direction)}($toSchema);
-            $this->executeRegisteredSql($dryRun, $timeAllQueries);
+
+            $allSqlRegistered = array_merge($allSqlRegistered, $this->sql);
+            $countSqlExecuted += $this->executeRegisteredSql($dryRun, $timeAllQueries);
 
             if (! $dryRun) {
                 if ($direction === self::DIRECTION_UP) {
                     $this->markMigrated();
                 } else {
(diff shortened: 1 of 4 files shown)


farzadghanei:
This pull request fixes issue #367

lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -372,26 +382,21 @@ public function __toString()
 
     private function executeRegisteredSql($dryRun = false, $timeAllQueries = false)
     {
+        $executedQueries = 0;
```
stof: no need for this variable. It is simply ``count($this->sql)``

farzadghanei: I thought of `executedRegisterSql` to return the number of actual SQL statements executed (considering dry run). That's the reason for this variable. However this can be implemented either by incrementing the variable in the loop (as is now), or getting the count after the loop using `count($this->sql)`. I don't have a strong opinion on which one would be better.

stof: before or after the loop does not change anything. The loop does not mutate the array

stof:
@farzadghanei I suggest you to read #369 (and the initial discussion which I linked in it). I think it might be better to have a decision there before finishing this PR as it will be impacted

farzadghanei:
@mikeSimonson As issue #369 is still open, I'm guessing no final decision has been made regarding running or preventing SQL commands in post migration steps.
Yet the issue of false warnings (issue #367) is still showing up on master. This merge request is to fix that issue only, and is independent of core contributors decision about SQL commands in post migration steps.

mikeSimonson:
@farzadghanei Thanks for your PR but I choose to revert my change as they weren't going in the right direction.

PR #371, opened 2015-10-22
Cleanup

TomasVotruba:
- few fixes of unused code
- phpunit.xml added to .gitignore (existence of `phpunit.xml.dist` would make no sense without it) - allows local phpunit configuration, that doesn't require coverage and is much much faster to run => faster and more pleasant development

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ tests/phpunit.xml
 vendor/
 composer.lock
 box.phar
+phpunit.xml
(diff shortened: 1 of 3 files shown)


mikeSimonson:
@TomasVotruba Thanks. Can you try to put completely unrelated change like thoses 3 ones in 3 different PR for the future ?

TomasVotruba:
@mikeSimonson  Yes, I'm ok with that. Thanks for explaining.

PR #373, opened 2015-10-27
Confinguration: check if migration class exists added

TomasVotruba:
This should prevent issue that happened to me and to users of package I maintain.

Migration class may not found due to typo in it's namespace, name or the namespace set in configuration.

This should tell user before it fail [while trying to create non-existing class in Version](https://github.com/doctrine/migrations/blob/68cc7bf629a7c1fc0696e583cc8ed46f897d0afb/lib/Doctrine/DBAL/Migrations/Version.php#L116). Took me while to figure out, what it really meant and where the failure was (naming namespace `Migrations` vs `migrations` etc.).

Test provided.

What do you think?


mikeSimonson:
@TomasVotruba I like it.
Can you move your exception has a method into the migrationException class.
https://github.com/doctrine/migrations/blob/master/lib/Doctrine/DBAL/Migrations/MigrationException.php

TomasVotruba:
Thanks for feedback.

Do you mean something like `MigrationException::classNotFound()`?

mikeSimonson:
More something like MigrationException::migrationClassNotFound()

TomasVotruba:
@mikeSimonson Done.

mikeSimonson:
@TomasVotruba Thanks

TomasVotruba:
Thank you

PR #376, opened 2015-11-11
RecursiveDirectoryIterator don't obtain some order of the file list.

drefixs:
RecursiveDirectoryIterator shouldn't return a sorted list and haven't to do it be direct. But migration script logic requires sorted list of the Version*.php files for correct working.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Finder/AbstractFinder.php
+++ b/lib/Doctrine/DBAL/Migrations/Finder/AbstractFinder.php
@@ -53,6 +53,9 @@ abstract class AbstractFinder implements MigrationFinderInterface
     protected function loadMigrations($files, $namespace)
     {
         $migrations = [];
+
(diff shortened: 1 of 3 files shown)


mikeSimonson:
Can you make a test for this case ?
@drefixs 

Thanks

drefixs:
This depend on the System OS and file system driver. This problem can be simple reproduce into docker containers env into CentOS 7. What test do you mean. If it need unit test to prevent this problem, i will make unit for this case. 
@mikeSimonson

Thanks

mikeSimonson:
@drefixs Ok this explain how you got a bug but was it already a test failling or is there no test to check that yet.

drefixs:
There no test to check that yet.

drefixs:
done. review pls.
@mikeSimonson 

thx.  :)

mikeSimonson:
@drefixs Sorry I missed your updated PR.

PR #377, opened 2015-11-17
Finished in Xs

Nyholm:
Added a little 's' after the number so the user knows that it finished in X seconds.


mikeSimonson:
@Nyholm Thanks

Nyholm:
No worries. Thank you for merging.

PR #378, opened 2015-11-19
Made composer.json compatible with SF 3.0

robinvdvleuten:
Hi there, 

It seems that the DoctrineMigrationsBundle is ]already made compatible](https://github.com/doctrine/DoctrineMigrationsBundle/blob/master/composer.json#L24) with SF 3.0 but you can't install it because this underlaying package still requires only ~2.3. This is a fix for that. Cheers!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,8 +13,8 @@
     "require": {
         "php": ">=5.4.0",
         "doctrine/dbal": "~2.2",
-        "symfony/yaml": "~2.3",
-        "symfony/console": "~2.3"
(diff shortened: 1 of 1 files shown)


mikeSimonson:
@robinvdvleuten thanks

patrickbrouwers:
@mikeSimonson can this be tagged as 1.1.1 ? 
Seems to be failing for people: https://github.com/laravel-doctrine/migrations/issues/10#issuecomment-162353049

mikeSimonson:
I will try to tag that this week but I need to fix master before to do it.

PR #400, opened 2015-12-22
Fix the write-sql functionnality and add a test for it

mikeSimonson:


stof:
Are we sure that ``executeRegisteredSql`` is called only once per migration ?

mikeSimonson:
@stof yes unless there is a bug that I missed.

PR #403, opened 2015-12-23
deleting the composer.lock file to make sure that we get the last dep

mikeSimonson:


thorstenk:
But how do you track library dependencies now?

One result is that Release 1.5.0 comes with `ocramius/proxy-manager` 2.0.4 now which requires PHP 7.0 (return type declarations were added) while `composer.json` still states compatibility with PHP 5.5.

Please consider to add the `composer.lock` file to the repo *and* revert this change of the build script as suggested in the [Composer documentation](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control).

alcaeus:
Hint: the composer.lock file is not used when running composer install in a project that has a dependency on this lib. Thus, it would not have prevented the scenario you are describing. If you find yourself with version issues after a composer update, please see this blog post: http://doctrine-project.org/2017/07/25/php-7.1-requirement-and-composer.html

thorstenk:
Installing Doctrine Migrations via Composer would have prevented that issue: we track the dependency locks in our projects and and so would not have updated `ocramius/proxy-manager` because it would have been tracked in the `composer.lock` of the respective project.

The issue I see is that the build script for the `.phar` packages always installs the latest allowed dependencies. If these aren't nailed down to distinct versions subtle changes may happen with every build and nobody notices them.

So that was what happened to us: we tried to switch from version 1.4.1 to 1.5.0 and found out that the latter version failed on our PHP 5.6 environments without prior notice. There's just no hint (neither in code nor in release notes) that Doctrine Migrations 1.5.0 don't work any more on PHP 5.6.

If the lock file was tracked in the Git repo the library would not have been updated accidentally. Yes, it's manual work to switch to the latest dependency versions then but that's how it is: dependency changes can't be automated IMO.

alcaeus:
> we track the dependency locks in our projects and so would not have updated `ocramius/proxy-manager` because it would have been tracked in the composer.lock of the respective project

😳 I don't even know what to say to that.

> The issue I see is that the build script for the .phar packages always installs the latest allowed dependencies. [...] and found out that the latter version failed on our PHP 5.6 environments without prior notice.

The issue I see is that you ran `composer update` on PHP 7.1 and went on to deploy the changes to a server running 5.6. If it makes you feel any better, you're not the first to do this, and I'm afraid you also won't be the last.

> Yes, it's manual work to switch to the latest dependency versions then but that's how it is: dependency changes can't be automated IMO.

Yes, they can. That's what SemVer is for, that's what the SAT solver in composer is for. If you tell composer to update your dependencies for PHP 7.1 (which you do by running `composer update` on PHP 7.1), it's going to do that.

If your code is supposed to run on PHP 5.6, there are two solutions:
* develop on 5.6 (which would make sense, since you would've spotted the error before you pushed your code to production
* use the `platform.config` setting in composer.json to tell composer you're using PHP 5.6, not whatever is installed on the system where you're running it.

I've said it before and I'll say it again: if you run into this issue, it's your own fault, not ours.

thorstenk:
>> we track the dependency locks in our projects and so would not have updated ocramius/proxy-manager because it would have been tracked in the composer.lock of the respective project

>😳 I don't even know what to say to that.

Maybe "I don't have a clue what `composer.lock` is for"? ;-)

Sorry for obviously not being clear enough about that: we're not installing Doctrine Migrations via Composer. We're using the `.phar` build, and that's what's broken in release 1.5.0. I assumed the change I commented on would be enough context and just answered to your statement without pointing to the `.phar` build process explicitly enough.

And to answer your assumptions about our build process:

If we would have installed Doctrine Migrations via Composer in our project, we would not have that issue: the dependencies of Doctrine Migrations, as stated in its `composer.json` are resolved when installing it and nailed down in `composer.lock`. So an update of `ocramius/proxy-manager` from 1.0.0 to 2.0.4. would have not slipped through in our project. And that's absolutely independent from the PHP version that runs Composer (which btw. is 5.6 from dev to production).

bendavies:
I think what @thorstenk is getting at is that the phar was built on php 7.

alcaeus:
@thorstenk I missed the part about you using the phar build - sorry about that. In that case, I'd agree with you: the phar build does not make it clear which dependencies it has, which in my opinion is a big problem with it. We should either start mentioning what system the phar was built on (e.g. lowest dependencies, highest) or stop publishing it since it can lead to cases like the one you unfortunately encountered.

mikeSimonson:
@thorstenk I will republish the phar build on php5.6

mikeSimonson:
done

PR #404, opened 2015-12-26
add possibility to read doctrine's config/cli-config.php

m00t:
As well as cli-config.php


mikeSimonson:
@m00t Thanks. Can you add a test case for this ?

m00t:
@mikeSimonson nope, because doctrine-migrations.php is not covered by tests at all =(

mikeSimonson:
@m00t thanks

PR #405, opened 2015-12-29
Correcting composer constraints: allowing PHP 7, dropping 5.4 support

Ocramius:


composer.json
```diff
@@ -11,7 +11,7 @@
         {"name": "Michael Simonson", "email": "contact@mikesimonson.com" }
     ],
     "require": {
-        "php": ">=5.4.0",
+        "php": "^5.5|^7.0",
```
Aliance: Silly question, but will this works with php 5.6? According to [composer docs](https://getcomposer.org/doc/articles/versions.md#caret) `^5.5` === `>=5.5.0 <5.6.0`.

Does `~5.5` will be better?

alcaeus: `^5.5` === `>= 5.5.0 && < 6`
`~5.5` === `>= 5.5.0 && < 6`

The difference only is visible when specifying a patch version:
`^5.5.1` === `>= 5.5.1 && < 6`
`~5.5.1` === `>= 5.5.1 && < 5.6`

PR #407, opened 2015-12-30
Replacing the Schema by a proxy

mikeSimonson:
This PR try to explore the possibility of using Proxy object instead of the costly initialization of the schema object.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/LazySchemaDiffProvider.php
+++ b/lib/Doctrine/DBAL/Migrations/LazySchemaDiffProvider.php
@@ -83,7 +83,7 @@ class LazySchemaDiffProvider implements SchemaDiffProviderInterface
      * @param Schema $fromSchema
      * @param Schema $toSchema
      *
-     * @return array
+     * @return string[]
      */
     public function getSqlDiffToMigrate(Schema $fromSchema,Schema $toSchema)
     {
--- a/lib/Doctrine/DBAL/Migrations/LazySchemaDiffProvider.php
+++ b/lib/Doctrine/DBAL/Migrations/LazySchemaDiffProvider.php
@@ -19,20 +19,20 @@
 
 namespace Doctrine\DBAL\Migrations;
 
+use Doctrine\DBAL\Migrations\Provider\SchemaDiffProviderInterface;
 use Doctrine\DBAL\Schema\Schema;
-use ProxyManager\Factory\LazyLoadingGhostFactory;
 use ProxyManager\Factory\LazyLoadingValueHolderFactory;
 use ProxyManager\Proxy\LazyLoadingInterface;
 
-class LazySchemaManipulator
+class LazySchemaDiffProvider implements SchemaDiffProviderInterface
 {
     /** @var  LazyLoadingValueHolderFactory */
     private $proxyFactory;
 
-    /** @var SchemaManipulator */
+    /** @var SchemaDiffProviderInterface */
     private $originalSchemaManipulator;
 
-    public function __construct(LazyLoadingValueHolderFactory $proxyFactory, SchemaManipulator $originalSchemaManipulator)
+    public function __construct(LazyLoadingValueHolderFactory $proxyFactory, SchemaDiffProviderInterface $originalSchemaManipulator)
     {
         $this->proxyFactory = $proxyFactory;
         $this->originalSchemaManipulator = $originalSchemaManipulator;
--- a/lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
+++ b/lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
@@ -21,17 +21,18 @@ namespace Doctrine\DBAL\Migrations;
 
 use Doctrine\DBAL\Schema\Schema;
 use ProxyManager\Factory\LazyLoadingGhostFactory;
+use ProxyManager\Factory\LazyLoadingValueHolderFactory;
 use ProxyManager\Proxy\LazyLoadingInterface;
 
 class LazySchemaManipulator
 {
-    /** @var  LazyLoadingGhostFactory */
+    /** @var  LazyLoadingValueHolderFactory */
     private $proxyFactory;
 
     /** @var SchemaManipulator */
     private $originalSchemaManipulator;
 
-    public function __construct(LazyLoadingGhostFactory $proxyFactory, SchemaManipulator $originalSchemaManipulator)
+    public function __construct(LazyLoadingValueHolderFactory $proxyFactory, SchemaManipulator $originalSchemaManipulator)
     {
         $this->proxyFactory = $proxyFactory;
         $this->originalSchemaManipulator = $originalSchemaManipulator;
@@ -46,7 +47,7 @@ class LazySchemaManipulator
 
         return $this->proxyFactory->createProxy(
             Schema::class,
-            function (& $wrappedObject, $method, array $parameters, & $initializer) use ($originalSchemaManipulator) {
+            function (& $wrappedObject, $proxy, $method, array $parameters, & $initializer) use ($originalSchemaManipulator) {
                 $initializer   = null;
                 $wrappedObject = $originalSchemaManipulator->createFromSchema();
 
@@ -66,7 +67,7 @@ class LazySchemaManipulator
         if ($fromSchema instanceof LazyLoadingInterface && ! $fromSchema->isProxyInitialized()) {
             return $this->proxyFactory->createProxy(
                 Schema::class,
-                function (& $wrappedObject, $method, array $parameters, & $initializer) use ($originalSchemaManipulator, $fromSchema) {
+                function (& $wrappedObject, $proxy,  $method, array $parameters, & $initializer) use ($originalSchemaManipulator, $fromSchema) {
                     $initializer   = null;
                     $wrappedObject = $originalSchemaManipulator->createToSchema($fromSchema);
 
--- a/lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
+++ b/lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
@@ -1,0 +1,69 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the LGPL. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Migrations;
+
+class LazySchemaManipulator
+{
+    private $proxyFactory;
+
+    private $originalSchemaManipulator;
+
+    public function __construct($proxyFactory, $originalSchemaManipulator)
+    {
+        $this->proxyFactory = $proxyFactory;
+        $this->originalSchemaManipulator = $originalSchemaManipulator;
+    }
+
+    public function createFromSchema()
+    {
+        return $this->proxyFactory->createProxy(
+            Schema::class,
+            function (& $wrappedObject, $proxy, $method, array $parameters, & $initializer) {
+                $initializer   = null;
+                $wrappedObject = $this->originalSchemaManipulator->createFromSchema();
+
+                return true;
+            }
+        );
+    }
+
+    public function createToSchema($fromSchema)
+    {
+        if ($fromSchema instanceof LazyLoadingInterface && ! $fromSchema->isProxyInitialized()) {
+            return $this->createFromSchema();
+        }
+
+        return $this->originalSchemaManipulator->createToSchema($fromSchema);
+    }
+
+    public function addSqlToSchema($fromSchema, $toSchema)
+    {
+        if (
+            $fromSchema instanceof LazyLoadingInterface
+            && $toSchema instanceof LazyLoadingInterface
+            && ! $fromSchema->isProxyInitialized()
+            && ! $toSchema->isProxyInitialized()
+        ) {
+            return;
+        }
+
+        return $this->originalSchemaManipulator->addSqlToSchema($fromSchema, $toSchema);
+    }
+}
--- a/lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
+++ b/lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
@@ -19,23 +19,32 @@
 
 namespace Doctrine\DBAL\Migrations;
 
+use Doctrine\DBAL\Schema\Schema;
+use ProxyManager\Factory\LazyLoadingGhostFactory;
+use ProxyManager\Proxy\LazyLoadingInterface;
+
 class LazySchemaManipulator
 {
+    /** @var  LazyLoadingGhostFactory */
     private $proxyFactory;
 
+    /** @var SchemaManipulator */
     private $originalSchemaManipulator;
 
-    public function __construct($proxyFactory, $originalSchemaManipulator)
+    public function __construct(LazyLoadingGhostFactory $proxyFactory, SchemaManipulator $originalSchemaManipulator)
     {
         $this->proxyFactory = $proxyFactory;
         $this->originalSchemaManipulator = $originalSchemaManipulator;
     }
 
+    /**
+     * @return Schema
+     */
     public function createFromSchema()
     {
         return $this->proxyFactory->createProxy(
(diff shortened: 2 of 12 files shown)


mikeSimonson:
@Ocramius ?

Ocramius:
@mikeSimonson the first run was just pseudo-code

mikeSimonson:
it needs a first run to warm up some sort of cache ?

Ocramius:
@mikeSimonson you didn't import any of the proxy lib stuff :-)

lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
```diff
@@ -0,0 +1,88 @@
...
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use ProxyManager\Factory\LazyLoadingGhostFactory;
+use ProxyManager\Proxy\LazyLoadingInterface;
+
+class LazySchemaManipulator
```
stof: not implementing a common interface looks wrong to me

Ocramius: Agreed. The name also needs fixing: we just mashed it together real quick in a video call yesterday :)

lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * @param Schema $fromSchema
+     * @return Schema
+     */
+    public function createToSchema(Schema $fromSchema)
+    {
+        if ($fromSchema instanceof LazyLoadingInterface && ! $fromSchema->isProxyInitialized()) {
+            return $this->createFromSchema();
```
stof: this looks wrong to me. This would then read the database schema twice (for from and to schemas), while it should be done only once (it is a costly process)

lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
```diff
@@ -0,0 +1,88 @@
...
+     * @param $toSchema
+     *
+     * @return array|string
+     */
+    public function getSqlDiffToMigrate($fromSchema, $toSchema)
+    {
+        if (
+            $fromSchema instanceof LazyLoadingInterface
+            && $toSchema instanceof LazyLoadingInterface
+            && ! $fromSchema->isProxyInitialized()
```
stof: the fromSchema will never be initialized when reaching this (as the user migration will never access it), so the feature would be broken

stof:
@mikeSimonson I guess this is an alternative to #323.
I see 1 drawback: if the user makes any call to the Schema by mistake, they would go to the slower codebase even though they might not want it. My proposal makes it an explicit choice, making people aware that they are going the slower way. What do you think ?

lib/Doctrine/DBAL/Migrations/SchemaManipulator.php
```diff
@@ -0,0 +1,65 @@
...
+use Doctrine\DBAL\Schema\Schema;
+
+class SchemaManipulator
+{
+    private $platform;
+
+    /** @var  AbstractSchemaManager */
+    private $schemaManager;
+
+    public function _construct(AbstractSchemaManager $schemaManager, $platform)
```
stof: missing ``_``, so this is not a constructor. This is why you are calling on ``null``

lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
Ocramius: @stof `createFromSchema` would be a lazy version of it

stof: yeah, but a different instance than the fromSchema, meaning you have 2 separate proxies needing to introspect the DB to initialize themselves. And this is precisely the issue.

stof: you should create a proxy which would initialize the fromSchema and then clone it when it needs to initialize the toSchema (this would also solve the other bug below, where the fromSchema never gets initialized)

lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
```diff
@@ -0,0 +1,88 @@
...
+        return $this->originalSchemaManipulator->createToSchema($fromSchema);
+    }
+
+    /**
+     * @param $fromSchema
+     * @param $toSchema
+     *
+     * @return array|string
+     */
+    public function getSqlDiffToMigrate($fromSchema, $toSchema)
```
stof: should be typehinted

lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        return $this->originalSchemaManipulator->createToSchema($fromSchema);
+    }
+
+    /**
+     * @param $fromSchema
+     * @param $toSchema
+     *
+     * @return array|string
```
stof: this can never return a string

lib/Doctrine/DBAL/Migrations/SchemaManipulator.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        return clone $fromSchema;
+    }
+
+    /**
+     * @param $fromSchema Schema
+     * @param $toSchema Schema
+     * @return array|string
+     */
+    public function getSqlDiffToMigrate($fromSchema, $toSchema)
```
stof: should be typehinted

lib/Doctrine/DBAL/Migrations/SchemaManipulator.php
```diff
@@ -0,0 +1,65 @@
...
+     * @return Schema
+     */
+    public function createToSchema(Schema $fromSchema)
+    {
+        return clone $fromSchema;
+    }
+
+    /**
+     * @param $fromSchema Schema
+     * @param $toSchema Schema
```
stof: invalid phpdoc (swapped type and name)

lib/Doctrine/DBAL/Migrations/SchemaManipulator.php
```diff
@@ -0,0 +1,65 @@
...
+     * @return array|string
```
stof: this is always an array

lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
Ocramius: @stof good point about the two proxies.

stof: my proposal would solve it though

lib/Doctrine/DBAL/Migrations/SchemaManipulator.php
stof: @mikeSimonson please fix this so that tests can run for real

stof:
@mikeSimonson the PR should describe what it really does IMO, not the issue you are facing currently

lib/Doctrine/DBAL/Migrations/LazySchemaManipulator.php
mikeSimonson: Any idea on a better naming ? for the class and the interface ?

mikeSimonson:
@stof The problem with your solution is that right now most of the people wouldn't benefit from it.
But it's definitely the way to go and the way it will be for v2 (1 interface that allow the schema and one that don't).

Can you look again at the last change I made.
I think that I took care of the issue with the double instantiation instead of the cloning.
All your others remarks should be fixed too except for the interface name.

I was thinking to something along the line of SchemaDiffProvider or SchemaProvider.
Do you have an opinion ?

Thanks for your review@stof The problem with your solution is that right now most of the people wouldn't benefit from it.
But it's definitely the way to go and the way it will be for v2 (1 interface that allow the schema and one that don't).

Can you look again at the last change I made.
I think that I took care of the issue with the double instantiation instead of the cloning.
All your others remarks should be fixed too except for the interface name.

I was thinking to something along the line of SchemaDiffProvider or SchemaProvider.
Do you have an opinion ?

Thanks for your review

mikeSimonson:
@stof Do you have some time to look at this again ?

lib/Doctrine/DBAL/Migrations/LazySchemaDiffProvider.php
```diff
@@ -0,0 +1,99 @@
...
+    public function createToSchema(Schema $fromSchema)
+    {
+        $originalSchemaManipulator = $this->originalSchemaManipulator;
+
+        if ($fromSchema instanceof LazyLoadingInterface && ! $fromSchema->isProxyInitialized()) {
+            return $this->proxyFactory->createProxy(
+                Schema::class,
+                function (& $wrappedObject, $proxy,  $method, array $parameters, & $initializer) use ($originalSchemaManipulator, $fromSchema) {
+                    $initializer   = null;
+                    $wrappedObject = $originalSchemaManipulator->createToSchema($fromSchema);
```
stof: shouldn't you force initializing the fromSchema here before passing it to the originalSchemaManipulator ?

lib/Doctrine/DBAL/Migrations/Provider/SchemaDiffProviderInterface.php
```diff
@@ -0,0 +1,56 @@
...
+    public function createToSchema(Schema $fromSchema);
+
+    /**
+     * Return an array of sql statement that migrate the database state from the
+     * fromSchema to the toSchema.
+     *
+     * @param Schema $fromSchema
+     * @param Schema $toSchema
+     *
+     * @return array
```
stof: I would document it as ``string[]``

lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -105,6 +106,9 @@ class Version
      */
     private $state = self::STATE_NONE;
 
+    /** @var  */
```
stof: missing type

lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -115,6 +119,8 @@ public function __construct(Configuration $configuration, $version, $class)
         $this->platform = $this->connection->getDatabasePlatform();
         $this->migration = new $class($this);
         $this->version = $version;
+        $schemaManipulator = new SchemaDiffProvider($this->sm, $this->platform);
```
stof: do we still need ``$this->sm`` or could it become a local variable ?

stof:
@mikeSimonson it is great to see the build being green here.

lib/Doctrine/DBAL/Migrations/LazySchemaDiffProvider.php
mikeSimonson: The createToSchema method clone the fromSchema and the clone trigger the initialization.

lib/Doctrine/DBAL/Migrations/Version.php
mikeSimonson: I was wondering if removing the private property would be an BC break. I think not and if not then we can safely remove it.

Ocramius: @mikeSimonson removing private properties is not a BC break: that's the beauty of making things `private` ;-)

lib/Doctrine/DBAL/Migrations/LazySchemaDiffProvider.php
stof: then it is good

lib/Doctrine/DBAL/Migrations/Provider/SchemaDiffProviderInterface.php
```diff
@@ -0,0 +1,56 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the LGPL. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\DBAL\Migrations\Provider;
```
stof: why is it in a subnamespace while implementations are not ?

mikeSimonson: no valid reason

mikeSimonson:
@stof @Ocramius Ok I think that this can be merged now ?

mikeSimonson:
@stof Do you think that you could do a profile of v1.2.1 against master to see the benefits ?

stof:
Here you go. This is the result of running 109 migrations on PostgreSQL 9.3 and PHP 5.6.16 on my laptop (with XDebug enabled as I forgot to disable it before profiling):

- Doctrine 1.2.1 https://blackfire.io/profiles/f96bd1bc-4539-4158-a4d6-e1823c5eea46/graph
- Doctrine 1.3.0-dev (9aa590e) https://blackfire.io/profiles/4a377dc2-811d-40f7-845b-06db1198fc4c/graph
 And here is the comparison: https://blackfire.io/profiles/compare/2be3159d-2821-4018-9a20-a80215695dd4/graph

We have a 97% performance improvement here (from 2min20s to 4.68s). This is similar to the 94% I reported in #323 (the difference is probably related to the fact that I have more migrations than at that time, making the effect even bigger)

Ocramius:
That is quite... impressive :O

@mikeSimonson still, are you sure that we want proxymanager installed as hard dependency?

stof:
@Ocramius I think we do (I'm looking forward the release of a zend-code release without the zend-stdlib dependency though, to have less transitive deps)

Ocramius:
As far as I know, it doesn't require stdlib, but the eventmanager: https://github.com/zendframework/zend-code/blob/4972c33d975fa0c2b86ff381e83fde177079b249/composer.json#L17

That can't be fixed until we drop annotations support ( zendframework/zend-code#32 ), which will only happen in a major revision.

Making it an optional dependency without dropping that feature will just lead to a lot of people having headaches when they want to actually use the annotations support :-\

stof:
There is a drawback in this change though: look at the metrics about the GC runs. We went from 1 to 81.
I think this is related to a circular object graph in unitialized proxies. It may be worth investigating it to see whether we can break the object graph when we don't need the proxies anymore (at the end of each migration). But the time spent in GC is not comparable to the time win, so this does not block a 1.3.0 release.

stof:
And regarding the dependencies, I think a 97% perf increase justifies the mandatory dependency. If we make it optional, people would have to opt-in the fast way, which would not be much better than #323 (well, except the opt-in would be in a single place), and we would have to expose this layer outside the library (currently, it is hidden from the outside, allowing us to treat it as fully internal)

Ocramius:
> It may be worth investigating it to see whether we can break the object graph when we don't need the proxies anymore (at the end of each migration).

The initializer closures will obviously cause cyclic graphs. There is a way to "fix that" (simply `setInitializer(null)` on proxies), but:
 * it breaks the abstraction
 * obviously isn't that relevant

Agree on all the conclusions.

stof:
@Ocramius the idea is to reset the initializer at the end of the migration, when we don't need the Schema object anymore (the refcount will then destroy it)

Ocramius:
@stof I wouldn't do that anyway, as it may lead to unexpected bugs (in case the objects are needed after migration run, or are used in any other context)

Ocramius:
@mikeSimonson this patch still lacks some tests that are required to prevent any regression that may by accident initialize a migration. I know it is a pain, but that means extremely detailed mocking :-\

lib/Doctrine/DBAL/Migrations/Provider/SchemaDiffProviderInterface.php
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Return an array of sql statement that migrate the database state from the
+     * fromSchema to the toSchema.
+     *
+     * @param Schema $fromSchema
+     * @param Schema $toSchema
+     *
+     * @return string[]
+     */
+    public function getSqlDiffToMigrate(Schema $fromSchema,Schema $toSchema);
```
Tobion: missing space

lib/Doctrine/DBAL/Migrations/Provider/SchemaDiffProvider.php
```diff
@@ -0,0 +1,67 @@
...
+    {
+        return clone $fromSchema;
+    }
+
+    /**
+     * @param Schema $fromSchema
+     * @param Schema $toSchema
+     * @return string[]
+     */
+    public function getSqlDiffToMigrate(Schema $fromSchema,Schema $toSchema)
```
Tobion: missing space

lib/Doctrine/DBAL/Migrations/Provider/SchemaDiffProviderInterface.php
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Generates `Schema` objects to be passed to the migrations class.
+ *
+ * @since   1.3
+ */
+interface SchemaDiffProviderInterface
+{
+    /**
+     * Create the schema that represent the current state of the database.
```
Tobion: represents

mikeSimonson:
@Ocramius What sort of test do you think are missing ?

mikeSimonson:
@stof detail but when I look at the profile of the future 1.3 and that I go in the Metrics tab I see one call to GC et 81 for the profile of the 1.2. But in the comparison page in the metrics tab the numbers seems inverted. Same for the number of PDO and SQL queries. Bug in blackfire ?

Btw it make more sense to me that there would be less call to the GC with the Proxy because we are not instantiating all those Tables objects in the Schema.

Ocramius:
@mikeSimonson the tests that I'm referring to would expect the schema object methods to never be called: basically a set of `$schema->expects(self::never())->method('theMethod')` for all methods.

mikeSimonson:
@Ocramius I don't understand the goal this test ? Do you mean testing that none of the method of the schema are called in the pre and post method or something else ?

Ocramius:
@mikeSimonson correct. That is just to prevent regressions though.

stof:
> Bug in blackfire ?

Looks like it is. then everything is good (and the issue would then be about the Schema system in DBAL relying on cyclic object graphs, which is off-topic)

stof:
I reported the bug to the Blackfire team

mikeSimonson:
@Ocramius There are already tests for that. It test that if those methods are used it has no impact on the sql produced https://github.com/doctrine/migrations/commit/fe1447a6033b14ff9dd26d2fee7e8c7d153f133d.
@stof Thanks

Ocramius:
@mikeSimonson that test is still calling API methods on the schema objects - a test where no method calls on the schema ever happen is required

mikeSimonson:
@Ocramius A test like that one ? #414

PR #413, opened 2016-01-08
New lines harmonization

mikeSimonson:


lib/Doctrine/DBAL/Migrations/SqlFileWriter.php
```diff
@@ -77,12 +77,12 @@ public function write(array $queriesByVersion, $direction)
 
     private function buildMigrationFile(array $queriesByVersion, $direction)
     {
-        $string = sprintf("# Doctrine Migration File Generated on %s\n", date('Y-m-d H:i:s'));
+        $string = sprintf("# Doctrine Migration File Generated on %s" . PHP_EOL, date('Y-m-d H:i:s'));
```
stof: I vote -1 for forcing PHP_EOL in generated files. In any project involving multiple OS, the standard is generally to use LF for files, not the line endings of the OS on which the file was created

mikeSimonson: @stof As everyone is on probably on git and git take care of the line ending standardisation. Do you still think that it would be a problem ?

stof: It does not always takes care of it. It depends on the config

PR #414, opened 2016-01-09
Regression test

mikeSimonson:


lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -111,9 +111,12 @@ public function __construct(Configuration $configuration, $version, $class)
         $this->migration = new $class($this);
         $this->version = $version;
 
-        $schemaProvider = new SchemaDiffProvider($this->connection->getSchemaManager(),
-            $this->connection->getDatabasePlatform());
-        $this->schemaProvider = new LazySchemaDiffProvider(new LazyLoadingValueHolderFactory(), $schemaProvider);
+        $this->schemaProvider = $schemaProvider;
+        if(is_null($this->schemaProvider)) {
```
Ocramius: can it ever be null? the docblock states it can't...

tests/Doctrine/DBAL/Migrations/Tests/Functional/FunctionalTest.php
```diff
@@ -13,6 +16,7 @@
 use Doctrine\DBAL\Migrations\Tests\Stub\Functional\MigrationModifySchemaInPreAndPost;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Migrations\Configuration\Configuration;
+use \Mockery as m;
```
Ocramius: I wouldn't introduce mockery, since it's not used in the project so far

tests/Doctrine/DBAL/Migrations/Tests/Functional/FunctionalTest.php
```diff
@@ -347,4 +351,18 @@ public function provideTestMigrationNames()
...
+    {
+        $schema = m::mock(Schema::class);
+        $schemaDiffProvider = new SchemaDiffStubProvider($schema);
+
+        $version = new Version($this->config, 1, MigrateNotTouchingTheSchema::class, $schemaDiffProvider);
+        $version->execute('up');
+
+        $methods = get_class_methods(Schema::class);
+        foreach($methods as $method) {
+            $schema->shouldNotHaveReceived($method);
```
Ocramius: Exactly this, yes :-)

tests/Doctrine/DBAL/Migrations/Tests/Provider/SchemaDiffStubProvider.php
```diff
@@ -0,0 +1,64 @@
...
+ */
+
+namespace Doctrine\DBAL\Migrations\Tests\Provider;
+
+use Doctrine\DBAL\Migrations\Provider\SchemaDiffProviderInterface;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Schema\AbstractSchemaManager;
+use Doctrine\DBAL\Schema\Schema;
+
+class SchemaDiffStubProvider implements SchemaDiffProviderInterface
```
Ocramius: Why do you need to stub out the provider like this? A mocker can do this...

mikeSimonson: How ?

tests/Doctrine/DBAL/Migrations/Tests/Functional/FunctionalTest.php
mikeSimonson: It's at least already used in the SqlFileWriter https://github.com/doctrine/migrations/blob/master/tests/Doctrine/DBAL/Migrations/Tests/SqlFileWriterTest.php#L24

lib/Doctrine/DBAL/Migrations/Version.php
mikeSimonson: Does this 
```php
/** @var SchemaDiffProviderInterface */
    private $schemaProvider;
```
means that the property can never be null ?

It will never be null it's just the parameter of the constructor that is optional.

Ocramius: Then it should state `/** @var SchemaDiffProviderInterface|null */`

tests/Doctrine/DBAL/Migrations/Tests/Functional/FunctionalTest.php
Ocramius: Is it widely used? While I understand that each project has its own preferences, I'd avoid having another mocking framework in here. I can replace the mocks myself, given some time

tests/Doctrine/DBAL/Migrations/Tests/Provider/SchemaDiffStubProvider.php
Ocramius:
```php
$mock = $this->getMock(SchemaDiffProviderInterface::class);

$mock->expects(self::any())->method('createFromSchema')->willReturn($schema);
$mock->expects(self::any())->method('createFromSchema')->willReturnCallback(function () use ($schema) { return clone $schema; });
$mock->expects(self::any())->method('getSqlDiffToMigrate')->willReturn([]);
```

tests/Doctrine/DBAL/Migrations/Tests/Stub/Functional/MigrateNotTouchingTheSchema.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace Doctrine\DBAL\Migrations\Tests\Stub\Functional;
+
+use Doctrine\DBAL\Migrations\AbstractMigration;
+use Doctrine\DBAL\Schema\Schema;
+
+class MigrateNotTouchingTheSchema extends AbstractMigration
```
Ocramius: Should be implemented as spy: we want to verify that all methods are actually being called with our mock schema

mikeSimonson: The spy doesn't work because the Version class instantiate the class based on it's name. And I didn't found a way to instantiate the spy based on it's name.

lib/Doctrine/DBAL/Migrations/Version.php
mikeSimonson: Does the new way of setting that property prevent the need for null in the docblock ?

tests/Doctrine/DBAL/Migrations/Tests/Stub/Functional/MigrateNotTouchingTheSchema.php
Ocramius: Good point: didn't know this, although it evidences the lack of a collaborator there (to be fixed another time, if you think it is relevant).

lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -111,9 +111,14 @@ public function __construct(Configuration $configuration, $version, $class)
         $this->migration = new $class($this);
         $this->version = $version;
 
-        $schemaProvider = new SchemaDiffProvider($this->connection->getSchemaManager(),
-            $this->connection->getDatabasePlatform());
-        $this->schemaProvider = new LazySchemaDiffProvider(new LazyLoadingValueHolderFactory(), $schemaProvider);
+        if (!is_null($schemaProvider)) {
```
Ocramius: `null === ...`

lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -111,9 +111,14 @@ public function __construct(Configuration $configuration, $version, $class)
...
+            $this->schemaProvider = $schemaProvider;
+        }
+        if(is_null($schemaProvider)) {
```
Ocramius: `null === ...`

lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -111,9 +111,14 @@ public function __construct(Configuration $configuration, $version, $class)
...
+            $schemaProvider = new SchemaDiffProvider($this->connection->getSchemaManager(),
```
Ocramius: Alignment is scrambled here

PR #418, opened 2016-01-11
refactor the version alias in the migrations infos

mikeSimonson:


lib/Doctrine/DBAL/Migrations/Tools/Console/Helper/MigrationStatusInfosHelper.php
```diff
@@ -72,6 +56,30 @@ public function getMigrationsInfos()
         return $infos;
     }
 
+    private function getVersionAlias($alias)
```
stof: This is not really getting an alias. It is formatting it

mikeSimonson: @stof getFormattedVersionAlias ?

PR #420, opened 2016-01-11
refactoring the code that sets the query parameters

mikeSimonson:


lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -176,21 +176,26 @@ public function addSql($sql, array $params = [], array $types = [])
...
             if (!empty($params)) {
-                $index = count($this->sql) - 1;
-                $this->params[$index] = $params;
-                $this->types[$index]  = $types;
+                $this->addQueryParams($params, $types);
             }
         }
     }
 
+    private function addQueryParams($params, $types)
```
stof: missing phpdoc

PR #421, opened 2016-01-14
Fix issue with some method when the migrations were not loaded

mikeSimonson:
Previously you needed to make sure manually that registerMigrations was
called before to make any call to a method that was using the migrations
loaded.

This might cause performance issue in some edge case but it's unlikely.
Unfortunately in the current desgin I don't see any other solution.


mikeSimonson:
@stof Can you review this change ?

TomasVotruba:
Awesome, I came across this issue many times! Thank you

PR #423, opened 2016-01-17
Trying to configure the proxy so that it can be embeded in the phar file

mikeSimonson:


mikeSimonson:
@Ocramius Isn't this going to cause issue after the proxy is generated and people try to composer update ?

Ocramius:
> @Ocramius Isn't this going to cause issue after the proxy is generated and people try to composer update ?

That is always an issue, but I'm trying to add proxy signing to ProxyManager v2 (which will indeed solve this, or at least I do hope so!)

lib/Doctrine/DBAL/Migrations/Provider/LazySchemaDiffProvider.php
```diff
@@ -32,8 +33,23 @@ class LazySchemaDiffProvider implements SchemaDiffProviderInterface
     /** @var SchemaDiffProviderInterface */
     private $originalSchemaManipulator;
 
-    public function __construct(LazyLoadingValueHolderFactory $proxyFactory, SchemaDiffProviderInterface $originalSchemaManipulator)
+    /** @var  Configuration */
+    private $proxyConfig;
+
+    public function __construct(LazyLoadingValueHolderFactory $proxyFactory=null, SchemaDiffProviderInterface $originalSchemaManipulator)
```
Ocramius: Please don't allow the proxy factory to be null. The `LazySchemaDiffProvider` is a new class, and there is no need to provide defaults in this way. If you need to provide sane defaults, add a `public static function fromDefaultProxyFacyoryConfiguration()`, for example

lib/Doctrine/DBAL/Migrations/Provider/LazySchemaDiffProvider.php
```diff
@@ -32,8 +33,23 @@ class LazySchemaDiffProvider implements SchemaDiffProviderInterface
...
     {
+        if ($proxyFactory === null ) {
```
Ocramius: Block to be moved to a factory or to a named constructor

lib/Doctrine/DBAL/Migrations/Provider/LazySchemaDiffProvider.php
```diff
@@ -32,8 +33,23 @@ class LazySchemaDiffProvider implements SchemaDiffProviderInterface
...
+            $this->proxyConfig = new Configuration();
+            $proxyFolder = __DIR__ . DIRECTORY_SEPARATOR . 'SchemaProxy';
```
Ocramius: No need to rely on `DIRECTORY_SEPARATOR` btw (PHP abstracts this for us since ancient versions around 2.x)

Ocramius: I suggest using a name that is more clear than `SchemaProxy` (`DoctrineMigrationShemaProxy`, for example)

stof: writing proxies inside the codebase is wrong IMO (this path may not even be writable)

lib/Doctrine/DBAL/Migrations/Provider/LazySchemaDiffProvider.php
```diff
@@ -32,8 +33,23 @@ class LazySchemaDiffProvider implements SchemaDiffProviderInterface
...
+            if (!is_dir($proxyFolder)) {
```
Ocramius: Instead of relying on the filesystem, given that this is a rarely-used feature, I suggest using the [`EvaluatingGeneratorStrategy`](https://github.com/Ocramius/ProxyManager/blob/57e9272ec0e8deccf09421596e0e2252df440e11/src/ProxyManager/GeneratorStrategy/EvaluatingGeneratorStrategy.php#L29)  instead, which will not write file to disk (and save you a huge amount of headaches). You will also not need the custom autoloader in that case

lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -117,7 +117,7 @@ public function __construct(Configuration $configuration, $version, $class, Sche
         if($schemaProvider === null) {
             $schemaProvider = new SchemaDiffProvider($this->connection->getSchemaManager(),
                 $this->connection->getDatabasePlatform());
-            $this->schemaProvider = new LazySchemaDiffProvider(new LazyLoadingValueHolderFactory(), $schemaProvider);
+            $this->schemaProvider = new LazySchemaDiffProvider(null, $schemaProvider);
```
Ocramius: Use the named constructor here, IMO

lib/Doctrine/DBAL/Migrations/Provider/LazySchemaDiffProvider.php
Ocramius: Indeed, see https://github.com/doctrine/migrations/pull/423/files#r50260316 for a better suggestion

stof:
> That is always an issue, but I'm trying to add proxy signing to ProxyManager v2 (which will indeed solve this, or at least I do hope so!)

Once using the evaluating strategy, it is not really an issue, as you would probably run the migration command in a separate process than composer (and so would not have run it before and after the update in the same process)

Ocramius:
@stof indeed, using a just-in-time evaluation removes the problem overall, as proxies are not persisted across multiple usages of the migrations process.

stof:
IMO, it is fine to use just-in-time evaluation for migrations. This is not a library you will use in the hot path of your app (you should only be using it at deployment time. I don't see any reason to execute migrations elsewhere)

Ocramius:
@stof this is still a general-purpose library, heh...

PR #425, opened 2016-01-23
ProxyManager ~2.0 is also compatible with migrations

Ocramius:
Note: can't seem to be able to install it with the dev dependencies unless I remove `"doctrine/coding-standard": "dev-master"`, which relies on an old PHP_CodeSniffer version. Other than that, runs fine.


TomasVotruba:
:+1:

stof:
:+1: I also thought about doing such change yesterday evening

Ocramius:
The dependency issue was fixed by @stof via zendframework/zend-code#38

PR #431, opened 2016-02-16
[DX] Formatted diffs

Padam87:
This would be really convinient, especially for new projects.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,11 @@
         "satooshi/php-coveralls": "0.6.*",
         "doctrine/coding-standard": "dev-master",
         "mockery/mockery": "^0.9.4",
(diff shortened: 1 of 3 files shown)


Padam87:
Come to think of it, the composer suggestion might not be the best solution because of the phar release. Should I move it to require-dev?

composer.json
```diff
@@ -25,6 +25,9 @@
         "mockery/mockery": "^0.9.4",
         "johnkary/phpunit-speedtrap": "~1.0@dev"
     },
+    "suggest": {
+        "jdorn/sql-formatter": "Allows to generate formatted SQL with the diff command."
```
stof: you should also add a dev requirement to allow the tests to run

mikeSimonson:
@Padam87 Thanks

PR #432, opened 2016-02-16
Refactoring a method to not depend on the input but on the individual…

mikeSimonson:
… options

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
@@ -102,8 +102,18 @@ EOT
             }
         }
 
-        $up = $this->buildCodeFromSql($configuration, $input, $fromSchema->getMigrateToSql($toSchema, $platform));
-        $down = $this->buildCodeFromSql($configuration, $input, $fromSchema->getMigrateFromSql($toSchema, $platform));
+        $up = $this->buildCodeFromSql(
+            $configuration,
+            $fromSchema->getMigrateToSql($toSchema, $platform),
+            $input->getOption('formatted'),
+            $input->getOption('line-length')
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
```diff
@@ -117,7 +127,7 @@ public function execute(InputInterface $input, OutputInterface $output)
         $output->writeln(sprintf('Generated new migration class to "<info>%s</info>" from schema differences.', $path));
     }
 
-    private function buildCodeFromSql(Configuration $configuration, InputInterface $input, array $sql)
+    private function buildCodeFromSql(Configuration $configuration, array $sql, $formatted=false, $lineLength=0)
```
stof: ``0`` look like a weird default

mikeSimonson: false ?

stof: no. It needs to be an integer. But any value smaller than 27 will be broken (see the logic line 147)

mikeSimonson: 80 then.

stof: yeah, or 120 to be consistent with the default value of the CLI option itself :)

mikeSimonson: ok thanks

mikeSimonson: done

PR #433, opened 2016-02-16
Fix: ExecuteCommand

rcatlin:
* Fix `\Doctrine\DBAL\Migrations\Tools\Console\Command\ExecuteCommand` unable to find Migration Version by registering Migration Versions on call to `\Doctrine\DBAL\Migrations\Configuration\Configuration#getVersion()`
* Adds missing check for empty migrations array in `\Doctrine\DBAL\Migrations\Configuration\Configuration#getVersion()`
* Commit https://github.com/doctrine/migrations/commit/916bee27bae598fbaf90bc21117fb6b2f76bba1c previously added checks for registered migrations (Pull Request https://github.com/doctrine/migrations/pull/421)
* Added tests for `\Doctrine\DBAL\Migrations\Configuration\Configuration#getVersion()` 

> Fix issue with some method call order when the migrations were not loaded
>
> Previously you needed to make sure manually that registerMigrations was
called before to make any call to a method that was using the migrations
loaded.
>
> This might cause performance issue in some edge case but it's unlikely.
Unfortunately in the current desgin I don't see any other solution.

### DIFF
--- a/tests/Doctrine/DBAL/Migrations/Tests/ConfigurationTest.php
+++ b/tests/Doctrine/DBAL/Migrations/Tests/ConfigurationTest.php
@@ -4,7 +4,6 @@ namespace Doctrine\DBAL\Migrations\Tests;
 
 use Doctrine\DBAL\Migrations\Configuration\Configuration;
 use Doctrine\DBAL\Migrations\MigrationException;
-use Doctrine\DBAL\Migrations\Tests\Stub\Configuration\AutoloadVersions\Version1Test;
 
 class ConfigurationTest extends MigrationTestCase
 {
@@ -193,7 +192,7 @@ class ConfigurationTest extends MigrationTestCase
     }
 
     /**
-     * @dataProvider getVersionProvider
+     * @dataProvider autoloadVersionProvider
      *
      * @param $version
      */
@@ -243,7 +242,7 @@ class ConfigurationTest extends MigrationTestCase
     /**
      * @return array
      */
-    public function getVersionProvider()
+    public function autoloadVersionProvider()
     {
         return [
             ['1Test'],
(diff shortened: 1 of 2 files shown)


rcatlin:
@mikeSimonson review?

mikeSimonson:
@rcatlin Can you add a test for it ?

rcatlin:
@mikeSimonson Absolutely!

rcatlin:
@mikeSimonson Added tests for `Configuration#getVersion()`. Please take another look when you have a chance.

mikeSimonson:
@rcatlin Thanks

tests/Doctrine/DBAL/Migrations/Tests/ConfigurationTest.php
```diff
@@ -191,6 +192,31 @@ public function testFormatVersion($version, $return)
...
+        $config->setMigrationsDirectory(__DIR__ . '/Stub/Configuration/AutoloadVersions');
+
+        $result = $config->getVersion($version);
+
+        $this->assertNotNull($result);
+    }
+
+    public function testGetVersionNotFound()
+    {
+        $this->setExpectedException(MigrationException::class);
```
stof: Putting this after the call to ``getSqliteConfiguration`` would be even better, as it would enforce the exception to happen in ``getVersion`` rather than ``getSqliteConfiguration`` (which is the drawback of the ``@expectedException`` annotation btw)

rcatlin: That's an excellent point! I'll put up a PR with the change.

rcatlin: Note: Pull Request created https://github.com/doctrine/migrations/pull/435

PR #434, opened 2016-02-17
Fixing an issue in the order at which some configuration key are loaded

mikeSimonson:
There is validation that check that the namespace and folder name are
set. Namely the validate method in the configuration class.

Because of that check we need to make sure that the namespace is set
before any migration are loaded, as the check happen there.


LogansUA:
@mikeSimonson thanks for solve this issue.

May i ask why do you add 2 key => value pair with the same key `name` and value for `$configurationProperties` array (`AbstractFileConfiguration` line `56`)?

mikeSimonson:
@LogansUA thanks for the notice. Fixed

PR #435, opened 2016-02-18
Fix: Move Expected Exception

rcatlin:
* Moves `expectException()` call in Unit Test after the call to `getSqlConfiguration` to ensure exception is thrown from call to `getVersion` as intended.


rcatlin:
@stof review?

stof:
:+1:

rcatlin:
@stof Thank you!

PR #439, opened 2016-03-04
add missing dependency

sheridan-dev:
ocramius library is needed to list migration files from disk


mikeSimonson:
@sheridan-dev Thanks

PR #443, opened 2016-03-11
Adds the missing doctrine-migrations.php to bin

wshafer:


mikeSimonson:
The doctrine-migrations.php doesn't need to be added to the bin only the doctrine-migration file is needed.
But you other commit on the autoloaders might be needed. Can you explain it and put it in a separate PR ?

Thanks

wshafer:
Actually I found the issue.  It had nothing to do with this package.

Thanks.

bin/doctrine-migrations.php
```diff
@@ -18,7 +18,7 @@
  */
 
 $autoloadFiles = array(
-    __DIR__ . '/../vendor/autoload.php',
+    __DIR__ . '/../autoload.php',
```
stof: This change is wrong. When the project is the root folder, the autoloader will indeed be at ``__DIR__ . '/../vendor/autoload.php'``

PR #446, opened 2016-03-15
Add ConsoleRunner similar to that in DBAL and ORM

c960657:
To make it easier to add the migration commands to existing console applications, this patch adds ConsoleRunner similar to that in DBAL and ORM.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/ConsoleRunner.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/ConsoleRunner.php
@@ -1,0 +1,86 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Migrations/Tools/Console/ConsoleRunner.php
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * @param Application $cli
+     *
+     * @return void
+     */
+    static public function addCommands(Application $cli)
+    {
+        $cli->addCommands(array(
+            new Command\ExecuteCommand(),
+            new Command\GenerateCommand(),
```
mikeSimonson: Thanks for the PR.
You forgot the Command\LatestCommand().
Can you update the PR ?

lib/Doctrine/DBAL/Migrations/Tools/Console/ConsoleRunner.php
```diff
@@ -0,0 +1,85 @@
...
+{
+    /**
+     * Runs console with the given helperset.
+     *
+     * @param \Symfony\Component\Console\Helper\HelperSet  $helperSet
+     * @param \Symfony\Component\Console\Command\Command[] $commands
+     *
+     * @return void
+     */
+    static public function run(HelperSet $helperSet, $commands = array())
```
stof: should be ``public static`` according to PSR-2 (even though Doctrine does not use exactly PSR-2)

lib/Doctrine/DBAL/Migrations/Tools/Console/ConsoleRunner.php
c960657: Fixed.

Thanks for the review.

lib/Doctrine/DBAL/Migrations/Tools/Console/ConsoleRunner.php
c960657: Fixed

mikeSimonson:
@c960657 Thanks a lot

PR #448, opened 2016-03-25
Fix: Remove unused imports

localheinz:
This PR

* [x] removes unused imports

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Connection/Loader/ConnectionConfigurationLoader.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Connection/Loader/ConnectionConfigurationLoader.php
@@ -21,7 +21,6 @@ namespace Doctrine\DBAL\Migrations\Configuration\Connection\Loader;
 
 
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Migrations\Configuration\Configuration;
 use Doctrine\DBAL\Migrations\Configuration\Connection\ConnectionLoaderInterface;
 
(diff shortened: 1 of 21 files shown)


mikeSimonson:
@localheinz Thanks

mikeSimonson:
@localheinz Did you use a tool to do this ? (which one). Is it possible to put back all the use statement that are used.

I mean that in the sense We have a class A in the namespace Z ( Z\A ) and it use another class B which is also in the namespace Z (Z\B). I would like the class A to import the class B even though it will be find because it's in the same namespace as A.

Also is there a we to change all the \DateTime to use \DateTime ?

Thanks for your time

localheinz:
@mikeSimonson 

I usually use [`fabpot/php-cs-fixer`](https://github.com/FriendsOfPHP/PHP-CS-Fixer), for example

```
$ composer global require fabpot/php-cs-fixer:dev-master
```

then, for example, 

```
$ php-cs-fixer fix --verbose --rules=no_unused_imports src
```

I also use PHPStorm and run inspections, it comes up with unused imports, too. Additionally, it provides intentions, for example, to import a class (or simplify a class reference), rather than using a FQCN.

Does that help?

mikeSimonson:
Yes thanks. Those rules don't look like they do what I would like though. :p

localheinz:
@mikeSimonson 

Can submit a PR if you like! Not right now, but towards the weekend, what do you say?

stof:
@mikeSimonson a use statement is about a class in the current namespace. If it is already in the current namespace, there is no point importing it (it is already there)

mikeSimonson:
@stof Even if it's in the current namespace I like the fact that it document the dependencies of a class or you think it's overkill ?

PR #449, opened 2016-03-25
Fix: Use class keyword

localheinz:
This PR

* [x] makes use of the `class` keyword instead of referring to class names using string literals

Follows #393.

:information_desk_person: Let me know if you prefer to get rid of FQCNs within this PR as well, otherwise I will follow up!


mikeSimonson:
@localheinz Yes, I would rather merge it after you get of the FQCN.

localheinz:
@mikeSimonson 

Amended!

localheinz:
@mikeSimonson 

About some of these imports: I personally prefer not to import everything, but instead import, let's say, parent namespace segments, what do you think?

mikeSimonson:
@localheinz I prefer to import everything to be explicit about the dependencies of the code

mikeSimonson:
@localheinz Thanks

PR #450, opened 2016-03-25
Fix: Use short array syntax

localheinz:
This PR

* [x] uses short array syntax

:information_desk_person: Fixed running

```
$ composer global require fabpot/php-cs-fixer
$ php-cs-fixer fix  --using-cache=0 --rules=short_array_syntax bin
$ php-cs-fixer fix  --using-cache=0 --rules=short_array_syntax lib
$ php-cs-fixer fix  --using-cache=0 --rules=short_array_syntax tests
```

### DIFF
--- a/bin/doctrine-migrations.php
+++ b/bin/doctrine-migrations.php
@@ -17,10 +17,10 @@
  * <http://www.doctrine-project.org>.
  */
 
-$autoloadFiles = array(
+$autoloadFiles = [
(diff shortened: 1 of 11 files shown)


mikeSimonson:
@localheinz Can you redo this ? It's probably easier than trying to rebase.
Thanks for the nice streak btw.

localheinz:
@mikeSimonson 

:+1:

PR #451, opened 2016-03-25
Fix: Case-insensitive method calls and types

localheinz:
This PR

* [x] fixes case-insensitive method calls and references to types


mikeSimonson:
@localheinz Thanks

PR #462, opened 2016-05-09
Fix #457

### LINKED ISSUE #457: Fix sorting: SORT_STRING on ksort()
We should add the optional parameter ``SORT_STRING`` to the ``ksort()`` function here: 
https://github.com/doctrine/migrations/blob/master/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php#L412

Reason: https://github.com/doctrine/DoctrineMigrationsBundle/issues/48

Thanks to @Charles68

JHGitty:
Fix #457.


mikeSimonson:
Thanks for the PR @JHGitty .

Can you write a test for this use case ?

mikeSimonson:
@JHGitty Thanks this fix has been imported and tested in the #463 PR

PR #463, opened 2016-05-13
Fixed relative version calculation for non-numeric values

syzygymsu:


mikeSimonson:
Any test ?

mikeSimonson:
@syzygymsu Thanks, I will look that up next week. Don't hesitate to ping me if I forget about it.

lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
```diff
@@ -601,15 +601,15 @@ public function getRelativeVersion($version, $delta)
...
+        $versions = array_map('strval', array_keys($this->migrations));
+        array_unshift($versions, '0');
+        $offset = array_search((string)$version, $versions);
         if ($offset === false || !isset($versions[$offset + $delta])) {
             // Unknown version or delta out of bounds.
             return null;
         }
 
-        return (string) $versions[$offset + $delta];
+        return $versions[$offset + $delta];
```
mikeSimonson: I would keep the casting to string there.

syzygymsu: What for? All values are already converted by `strval` and non-string case is covered by tests.
My idea was to fix type discrepancy as soon as possible and do not rely on extra casts.

syzygymsu:
BTW, should we also fix #457 here? It could be covered by this test with just a little change.

lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
mikeSimonson: Sorry I missed it.

mikeSimonson:
If you could add the test for it it would be perfect.

syzygymsu:
Modified test proves that #457 is fixed. I hope we don't need to test that `ksort` realy sorts.

tests/Doctrine/DBAL/Migrations/Tests/ConfigurationTest.php
```diff
@@ -139,7 +139,7 @@ public function testRelativeVersion()
         $config->registerMigrations([
             '0initial' => Version1Test::class,
             1 => Version2Test::class,
-            '2final' => Version3Test::class,
+            'final' => Version3Test::class,
```
mikeSimonson: did the test broke with the number prepending the migration name ?

syzygymsu: Without the number. Yep, until added `SORT_STRING`.

PR #470, opened 2016-06-18
add migrations:up-to-date command

abacaphiliac:
add up-to-date command, which simply tells the user whether or not there are migrations to run. returns a non-zero exit code if there are migrations available, which can be used more easily in diagnostic checks and build/deploy processes, without the need to parse output.


abacaphiliac:
i ran the build on my fork for commit 8c749fd:
https://travis-ci.org/abacaphiliac/migrations

mikeSimonson:
@abacaphiliac Thanks

PR #472, opened 2016-06-20
Use the standard comment marker in SQL migration files

### LINKED ISSUE #466: --write-sql creates broken sql files on postgresql
It writes comments about the migration into the sql file (see [here](https://github.com/doctrine/migrations/blob/0d0ff5da10c5d30846da32060bd9e357abf70a05/lib/Doctrine/DBAL/Migrations/Migration.php#L95)). Unfortunately it uses the # character which is only a comment in Mysql but not in Postgresql (here it is an [operator](https://www.postgresql.org/docs/current/static/sql-syntax-lexical.html#SQL-SYNTAX-SPECIAL-CHARS)).

Because of this we have to transform the file each time before we can actually use it.

I think changing it to `/* */` or `-- ` would be the best way because this is pretty standard among the different vendors. (e.g.: [postgresql](https://www.postgresql.org/docs/current/static/sql-syntax-lexical.html#SQL-SYNTAX-COMMENTS), [mysql](http://dev.mysql.com/doc/refman/5.7/en/comments.html), [oracle](https://docs.oracle.com/cd/B14117_01/server.101/b10759/sql_elements006.htm))

aaa2000:
Close #466

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/SqlFileWriter.php
+++ b/lib/Doctrine/DBAL/Migrations/SqlFileWriter.php
@@ -77,10 +77,10 @@ class SqlFileWriter
 
     private function buildMigrationFile(array $queriesByVersion, $direction)
     {
-        $string = sprintf("# Doctrine Migration File Generated on %s\n", date('Y-m-d H:i:s'));
+        $string = sprintf("-- Doctrine Migration File Generated on %s\n", date('Y-m-d H:i:s'));
 
         foreach ($queriesByVersion as $version => $queries) {
-            $string .= "\n# Version " . $version . "\n";
+            $string .= "\n-- Version " . $version . "\n";
             foreach ($queries as $query) {
                 $string .= $query . ";\n";
             }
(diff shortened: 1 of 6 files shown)


lib/Doctrine/DBAL/Migrations/SqlFileWriter.php
```diff
@@ -107,7 +107,6 @@ private function buildMigrationFilePath()
     {
         $path = $this->destPath;
         if (is_dir($path)) {
-            $path = realpath($path);
```
aaa2000: I need to remove the realpath call because https://github.com/mikey179/vfsStream/wiki/Known-Issues

I hope that adding this dependency is not a problem.

mikeSimonson:
@aaa2000 Can you add tests to verify that the removing of the realpath call did not break anything ?

aaa2000:
@mikeSimonson I have keept the realpath call and modified tests by mocking realpath in order to be sure that it did not break anything.

mikeSimonson:
@aaa2000 That works too.
Thanks

PR #477, opened 2016-07-05
Add Parameters to the Output of Dry Run Migrations

### LINKED ISSUE #186: Usability improvements for migrations with addSql and parameters
The `addSql`method accepts parameters for creating inserting in a query. This is very useful for data migrations. However, the usability could be improved:

1) By adding an example to the docs that uses placeholders and a parameters array (or links to http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/data-retrieval-and-manipulation.html where the placeholder syntax is explained)

2) By including them in the output of `--dry-run`

chrisguitarguy:
Closes #186 

Right now this looks like...

```
    -> INSERT INTO whatever VALUES (?, ?) with parameters (one, two)
```

with numeric indices in the param array. If there are string indices, those names are included in the output (under the assumption that they match placeholders in the SQL).

```
    -> INSERT INTO whatever VALUES (:one, :two) with parameters (:one => one, :two => two)
```

The DBAL type classes are used to convert the values to something suitable for output (specifically [Type::convertToDatabaseValue](https://github.com/doctrine/dbal/blob/97f423a42c100c0ea9a8ff78f17a9ed020b8b6bf/lib/Doctrine/DBAL/Types/Type.php#L110).

SQL statements without any params look the same as they do currently.

This is probably a good enough first pass at this, but it's not going to protect users from doing something wonky like...

```php
$this->addSql('INSERT INTO whatever VALUES (?)', [new \stdClass]);
```

and causing it the dryrun fail (granted the migration itself would fail anyway).

In other words: there's not a lot of validation. But that's not necessarily a bad thing. Types aren't validated to make sure they exist before fetching them with `Type::getType(...)`, but that's something thay may help a user catch bugs.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Version.php
+++ b/lib/Doctrine/DBAL/Migrations/Version.php
@@ -19,6 +19,7 @@
 
 namespace Doctrine\DBAL\Migrations;
 
+use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Migrations\Configuration\Configuration;
 use Doctrine\DBAL\Migrations\Provider\LazySchemaDiffProvider;
 use Doctrine\DBAL\Migrations\Provider\SchemaDiffProvider;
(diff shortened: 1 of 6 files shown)


mikeSimonson:
@chrisguitarguy I like this. Can you add a test with a datetime being passed ?

chrisguitarguy:
Added, @mikeSimonson. Also added a test for an `array` type. The test is setup to be able to easily add more types if necessary.

mikeSimonson:
@chrisguitarguy Thanks

PR #478, opened 2016-07-06
Switch the Generated Version Classes to use the Current time in UTC

### LINKED ISSUE #330: using gmdate instead of date for the default versionNumber ?
https://github.com/doctrine/DoctrineMigrationsBundle/issues/97#event-373566509

chrisguitarguy:
Rather than the current time in whatever PHP's default time zone happens to be set to.

Closes #330 

Couple big things here:
1. new tests for the generate and diff commands. Both just ensure that the expected files are created
2. I refactored the version number generation into the configuration (explained below)

I moved the version number generation into `Configuration::generateVersionNumber` for a few reasons. The first is that this change is *probably* a BC break. If someone upgraded a minor version, there's a chance that they may generate a new migration with a version number before their latest version. Not good.

Moving the version number generation into `Configuration` means we could preserve BC by adding a configuration option for a timezone (which defaults to UTC). Also means additional configuration could be presented around version numbers if desired (probably not a great idea, but it's there).

Second reason is that it makes the system as a whole more testable and centralizes the version number generation in once place (was in both the diff and generate commands before).

If this is too much for such a small change feel free to close @mikeSimonson.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
@@ -94,7 +94,8 @@ EOT
     {
         $configuration = $this->getMigrationConfiguration($input, $output);
 
-        $path = $this->generateMigration($configuration, $input);
+        $version = $configuration->generateVersionNumber();
+        $path = $this->generateMigration($configuration, $input, $version);
 
         $output->writeln(sprintf('Generated new migration class to "<info>%s</info>"', $path));
     }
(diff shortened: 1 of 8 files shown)


lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
```diff
@@ -115,7 +114,7 @@ protected function generateMigration(Configuration $configuration, InputInterfac
         ];
         $replacements = [
             $configuration->getMigrationsNamespace(),
-            $version,
+            $version = $configuration->generateVersionNumber(),
```
mikeSimonson: @chrisguitarguy That's the only part of the PR that bothers me. I would prefer that the generateMigration function still gets the version "name" passed to it rather than hardcode it in the method.

mikeSimonson:
@chrisguitarguy Besides my little remarks I think this is good to go.
Thanks for the refactor of the tests.

chrisguitarguy:
Should be good to go, @mikeSimonson. You worried about a possible BC break here? Or not a big deal?

mikeSimonson:
@chrisguitarguy Thanks

No it's mostly a design thing.

PR #479, opened 2016-07-06
Explicitly Call `Connection::connect` in Configuration

### LINKED ISSUE #336: Migrations fail when slave replication is running significantly behind master
Hi,

We have a problem where our build script will fail if run after another longer running migration.

If our build script executes and starts running a long migration it'll randomly (it appears) pick one of our three database instances, where 2 are slaves and one is master. The following can happen:

1. Migration begins on randomly chosen database instance. Let's say the master is picked.
2. Migration is large, and replication is unusually slow so the 2 slaves lag behind.
3. Build completes, but replication still running.
4. A new build (e.g. a hotfix) is kicked off.
5. Migrations runs again, but picks a slave this time - i.e. one of the database instances which is running behind.
6. It checks the migrations, and decides the migration hasn't been run yet because the replication hasn't finished.
7. Build fails because of an integrity constraint - the change it wants to apply has already been (at least partially) applied.

I'd expect, or assume there would be an option, to force the migrations to operate against the master (or a specific) database.

I'm using MySQL in Amazon's RDS if that helps.

Thanks!
Ed

chrisguitarguy:
And account for uses of `MasterSlaveConnection`. If there is a `MasterSlaveConnection` in use, the migrations system should always talk to master just in case replication is behind the next migration execution.

Should (hopefully) close #336 

When using `MasterSlaveConnection`, any calls to the connections `executeQuery` method (which includes all the `fetch*` methods) will connect to followers (slaves). Even things like checking if the migration table exists fall into this category. If the replication is behind, the followers may not have the real, already executed migrations.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
@@ -20,6 +20,7 @@
 namespace Doctrine\DBAL\Migrations\Configuration;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Connections\MasterSlaveConnection;
 use Doctrine\DBAL\Migrations\Finder\MigrationDeepFinderInterface;
 use Doctrine\DBAL\Migrations\MigrationException;
 use Doctrine\DBAL\Migrations\OutputWriter;
@@ -495,6 +496,7 @@ class Configuration
      */
     public function hasVersionMigrated(Version $version)
     {
+        $this->connect();
         $this->createMigrationTable();
 
         $version = $this->connection->fetchColumn(
@@ -512,6 +514,7 @@ class Configuration
      */
     public function getMigratedVersions()
     {
+        $this->connect();
         $this->createMigrationTable();
 
         $ret = $this->connection->fetchAll("SELECT " . $this->migrationsColumnName . " FROM " . $this->migrationsTableName);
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
```diff
@@ -816,6 +822,24 @@ public function generateVersionNumber(\DateTimeInterface $now=null)
...
+     * follower when fetching the executed migrations. If a follower is lagging
+     * significantly behind that means the migrations system may see unexecuted
+     * migrations that were actually executed earlier.
+     *
+     * @return bool The same value returned from the `connect` method
+     */
+    protected function connect()
+    {
+        if ($this->connection instanceof MasterSlaveConnection) {
+            return $this->connection->connect('master');
```
mikeSimonson: @chrisguitarguy This does assume that there is always a connection named master in a master slave connection setup right ?
If so, is it the case or should we make the name of the master connection configurable ?

chrisguitarguy: Good question. `MasterSlaveConection` [only accepts `master` and `slave` as arguments to connect](https://github.com/doctrine/dbal/blob/69b57859c8674f2c8ba3798a2541ac18128a0bdb/lib/Doctrine/DBAL/Connections/MasterSlaveConnection.php#L147-L149), so we should be okay.

mikeSimonson:
@chrisguitarguy And one more fixed.
Thanks

PR #480, opened 2016-07-07
Don't Prompt the User for Confirmation on the `migrations:migrate` Command When There is Nothing to Do

### LINKED ISSUE #306: Big scary warning when there are actually no migrations to execute.
Often when I run:

```
php app/console doctrine:migrations:migrate
```

I get this:

```
WARNING! You are about to execute a database migration that could result in schema changes and data lost. Are you sure you wish to continue? (y/n)y

No migrations to execute.
```

I think it would aid usability if the command did a pre-check for 0 migrations to run and just reported that first, so that the WARNING would be a special case certainly requiring extra scrutiny.

```
php app/console doctrine:migrations:migrate

No migrations to execute.
```

chrisguitarguy:
Possibly closes #306

Couple big things here:
- Better tests for the `MigrateCommand` so I could feel more comfortable changing it
- Moved the "No migrations ..." message into `Migration::migrate`.
- Reworked `Migration::migrate` to take an optional fourth argument for "confirm" the migrations can execute. This is similar to how `OuputWriter` abstracts away the Symfony console output in `Migration`. The default behavior is to always "confirm" the migrations if no `$confirm` callback is given (same as it is currently).

Not sure if adding another argument is okay, but it seemed to make the most sense. Even if none of the changes are okay, I do think the tests for the `MigrateCommand` are useful. Let me know!

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Migration.php
+++ b/lib/Doctrine/DBAL/Migrations/Migration.php
@@ -156,7 +156,7 @@ class Migration
          * to signify that there is nothing left to do.
          */
         if ($from === $to && empty($migrationsToExecute) && !empty($migrations)) {
-            return [];
+            return $this->noMigrations();
         }
 
         $output = $dryRun ? 'Executing dry run of migration' : 'Migrating';
@@ -168,6 +168,8 @@ class Migration
          */
         if (empty($migrationsToExecute) && !$this->noMigrationException) {
             throw MigrationException::noMigrationsToExecute();
+        } elseif (empty($migrationsToExecute)) {
+            return $this->noMigrations();
         }
 
         $sql = [];
@@ -184,5 +186,11 @@ class Migration
         $this->outputWriter->write(sprintf("  <info>++</info> %s sql queries", count($sql, true) - count($sql)));
 
         return $sql;
+    }
+
+    private function noMigrations()
+    {
+        $this->outputWriter->write('<comment>No migrations to execute.</comment>');
+        return [];
     }
 }
--- a/lib/Doctrine/DBAL/Migrations/Migration.php
+++ b/lib/Doctrine/DBAL/Migrations/Migration.php
@@ -119,7 +119,9 @@ class Migration
      * @param boolean $dryRun         Whether or not to make this a dry run and not execute anything.
      * @param boolean $timeAllQueries Measuring or not the execution time of each SQL query.
      * @param callable|null $confirm A callback to confirm whether the migrations should be executed.
+     *
      * @return array|false An array of migration sql statements or false if the confirm callback denied execution
+     *
      * @throws MigrationException
      */
     public function migrate($to = null, $dryRun = false, $timeAllQueries = false, callable $confirm = null)
--- a/lib/Doctrine/DBAL/Migrations/Migration.php
+++ b/lib/Doctrine/DBAL/Migrations/Migration.php
@@ -120,7 +120,7 @@ class Migration
      * @param boolean $timeAllQueries Measuring or not the execution time of each SQL query.
      * @param callable|null $confirm A callback to confirm whether the migrations should be executed.
      *
-     * @return array|false An array of migration sql statements or false if the confirm callback denied execution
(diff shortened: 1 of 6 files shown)


mikeSimonson:
@chrisguitarguy Thanks

It will take a little more time to review that PR for me but don't hesitate to ping me if I don't do it by next week.

lib/Doctrine/DBAL/Migrations/Migration.php
```diff
@@ -118,12 +118,11 @@ public function setNoMigrationException($noMigrationException = false)
      * @param string  $to             The version to migrate to.
      * @param boolean $dryRun         Whether or not to make this a dry run and not execute anything.
      * @param boolean $timeAllQueries Measuring or not the execution time of each SQL query.
-     *
-     * @return array $sql     The array of migration sql statements
-     *
```
stof: please keep these lines separating params from the return value. It makes things more readable IMO

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php
```diff
@@ -124,31 +124,37 @@ public function execute(InputInterface $input, OutputInterface $output)
...
         }
     }
 
     /**
+     * Create a new migration instance to execute the migrations.
+     *
+     * @param $configuration The configuration with which the migrations will be executed
+     * @return Migration a new migration instance
+     */
+    protected function createMigration(Configuration $configuration)
```
stof: Why adding a new method for this ?

chrisguitarguy: Purely to make the command easier to unit test using [`getMockForAbstractClass` to override this factory method](https://github.com/chrisguitarguy/migrations/blob/b9f207b9423d27f7e2c9e192f24b43d0817edb89/tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php#L193-L205). Ideally there'd probably be a factory object/interface or something, but that seemed like overkill.

Some of the functional tests cover the happy paths that deal with an actual `Migration` object, but the unit test coverage of the command was pretty slim.

gdscei:
@chrisguitarguy @mikeSimonson What's the status on this pull request? This would be an extremely helpful improvement.

mikeSimonson:
I will review and merge this tomorrow without fault.

chrisguitarguy:
Let me know if you need anything @mikeSimonson, I kind of lost track of this one.

chrisguitarguy:
Any updates here @mikeSimonson? It's okay to reject too, I know it's a big-ish change.

mikeSimonson:
@chrisguitarguy As I said my only reserve is the change of the signature of the migrate method to make it return false. What do you think about that ? Wouldn't it be better to make it return an empty array ?

chrisguitarguy:
> What do you think about that ? Wouldn't it be better to make it return an empty array ?

I guess I was looking for a way to signify the difference between no migrations being found and the cancel callback failing. This means that the command can exit with a `1` status code if the user cancels.

`return array|false` is a pretty crappy way to do that, however. Would probably be better to return a value object that iterable to mimic the current pattern and has a method to see whether the migraitons were canceled.

Thoughts?

mikeSimonson:
Wouldn't you consider returning 0 when the user cancel ? After all there was no error.

chrisguitarguy:
> After all there was no error.

That's fair.

I'll make the change tonight!

chrisguitarguy:
@mikeSimonson got back to this a bit late, sorry.

I remember why I did the `false` thing now: it was to keep the [same interface](https://github.com/doctrine/migrations/blob/dbbcc24ecaf04da11fd9d4dada9b6787eff29fba/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php#L132-L139) that's currently in master.

I reworked `Migration::migrate` to return an empty arrray and found another way to do that in the console command. Let me know if that works!

mikeSimonson:
@chrisguitarguy Thanks

PR #483, opened 2016-07-26
Dry run added to pre and post methods

MikeTralala:
When adding a column, we can't use `$this->addSql()` in postUp because this is not executed anymore. So we use `$this->connection->executeSql()`. This works fine if you execute the migration directly the database. But if you are dry-running to check if all is well, i dont want my postUp to execute any statements.

### DIFF
--- a/.idea/workspace.xml
+++ b/.idea/workspace.xml
@@ -1,0 +1,330 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="ChangeListManager">
+    <list default="true" id="a2b921dd-40d5-4f2f-a8b1-3cf14d2b7b0a" name="Default" comment="">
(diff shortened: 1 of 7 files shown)


stof:
this is impossible, because it is a BC break (child classes must respect the signature

MikeTralala:
Is there any other kind of setup that would do the trick?

chrisguitarguy:
You could add a property to `AbstractMigration` along with a `isDryRun` method and a setter for the property. Then set the dry run value in `Version` and let folks check `$this->isDryRun()` in their migration methods.

mikeSimonson:
The obvious solution here is to split your migration in 2 different migrations.

There is no other way that I can think off that wouldn't cause a massive BC break.

mikeSimonson:
@MikeTralala Also don't forget to put your .idea folder in your global gitignore to avoid committing it by accident.

dbu:
funny, we just stumbled over the same problem. i think this is something that should be re-thought for version 2. maybe postUp does not need to be executed at all in dry run mode? preUp might need it though, and it will be particularly relevant to know whether we are in dry run mode.

in our postUp, we use this workaround to check if the actual migration was executed or not:

```php
if (!in_array($newTableName, $schema->getTableNames())) {
    echo 'skip post update, this is a dry run';
    // this was a dry-run
    return;
}
```

MikeTralala:
@dbu Thanks haven't thought of that fix... and @mikeSimonson Sorry i forgot, thanks for the reminder.

PR #489, opened 2016-09-07
Fixed doctrine:migrations:execute command help message when using

eko:
This is a really quick fix but when using `doctrine:migrations:execute` command, the version number is required but is not displayed in the command help message when using `--no-interaction` option.

I fix that because a new person I'm teaching Symfony just told me that the command suggested in the help of the command don't work... ;)

Thank you!


mikeSimonson:
@eko Thanks

PR #492, opened 2016-09-20
Fixed #491

### LINKED ISSUE #491: The commands try to create the migrations table event if it exists
While I execute the command:

```
$ ./doctrine-migrations status
```

The exception is thrown:

```
 [Doctrine\DBAL\Exception\TableExistsException]
  An exception occurred while executing 'CREATE TABLE _migration (version VARCHAR(255) NOT NULL, PRIMARY KEY(version))':
  SQLSTATE[42P07]: Duplicate table: 7 ERROR:  relation "_migration" already exists

  [Doctrine\DBAL\Driver\PDOException]
  SQLSTATE[42P07]: Duplicate table: 7 ERROR:  relation "_migration" already exists

  [PDOException]
  SQLSTATE[42P07]: Duplicate table: 7 ERROR:  relation "_migration" already exists
```

The problem is because the migrations db table existence is not detected when its name is ignored by DBAL's filterSchemaAssetsExpression.

The existence of the table is checked by SchemaManager's method tablesExist() resp. listTableNames() which honors the regular expression for assets filtering.

The solution is to reset the regular expression before the existence check.

mishal:


mikeSimonson:
@mishal As explained in the bug report of #491 it doesn't make any sense for doctrine migrations to make assumption on the setup of the users

PR #497, opened 2016-11-14
Show newly generated migration

ErikTrapman:
Show the newly generated diff-migration upon creation. It helps speed up programming by knowing you can directly migrate the new migration if all proposed changes are in order. No need to check your IDE.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
@@ -125,6 +125,7 @@ EOT
         $path = $this->generateMigration($configuration, $input, $version, $up, $down);
 
         $output->writeln(sprintf('Generated new migration class to "<info>%s</info>" from schema differences.', $path));
+        $output->writeln(file_get_contents($path));
     }
 
     private function buildCodeFromSql(Configuration $configuration, array $sql, $formatted=false, $lineLength=120)


mikeSimonson:
@ErikSaunier Interesting, how readable is the output without any formatting like that ?

ErikTrapman:
screenshot_20161114_213450

Looks fine to me. Default terminal-view, the migration itself has enough spacing and you already know what part you are looking for (the up-function)

mikeSimonson:
@ErikTrapman Thanks

ErikTrapman:
@mikeSimonson thanks!

stof:
I would do it only when running the command in verbose mode

ErikTrapman:
@stof I was expecting this comment. I think this change adds most value when showing this as a default, instead of 'hiding' it behind a verbose-parameter. It's now a new, unexpected addition that I hope/think people will find useful.

trickeyone:
Just upgraded to 1.50 and saw this change. Personally, I'd prefer it be off by default. When I'm generating diffs, it tends to be sizable because the SchemaTool doesn't do a good job of telling when ManyToX/OneToX relations haven't actually changed. I tend to have ALTER/DROP INDEX/CREATE INDEX/etc statements that don't do anything because there were no changes to the entities or tables, but that's beside the point.

When looking at the options for the diff command, I don't see a way to turn this off at all except, maybe, `-q, --quiet` but I would still like to see the normal output.

My 2¢

alcaeus:
Thanks for your feedback everyone. I've changed the feature in #572 which was released as 1.6.1: the diff output is now only shown in verbose mode. To be honest, the output doesn't add a lot of information and it actually made the user experience of the Symfony Maker bundle quite ugly. After a short discussion with @weaverryan and @Simperfit we decided to change the feature.

PR #499, opened 2016-11-25
Added hasColumn() method

nicosomb:
Before creating a new column, it can be useful to check if this column already exists.

Doctrine migrations can be painful when you are a software editor: you don't when a user installed your software, so some migrations can be played whereas the schema of the user is good. 
We have this problem when we add new columns in our schema. 
That's why we wanted to add a check just before the `ALTER TABLE`. 

This code is very inspired by phinx from @robmorgan: [MySQL](https://github.com/robmorgan/phinx/blob/0.6.x-dev/src/Phinx/Db/Adapter/MysqlAdapter.php#L362), [SQLite](https://github.com/robmorgan/phinx/blob/0.6.x-dev/src/Phinx/Db/Adapter/SQLiteAdapter.php#L293), [Postgres](https://github.com/robmorgan/phinx/blob/0.6.x-dev/src/Phinx/Db/Adapter/PostgresAdapter.php#L346).

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/AbstractMigration.php
+++ b/lib/Doctrine/DBAL/Migrations/AbstractMigration.php
@@ -25,42 +25,44 @@ use Doctrine\DBAL\Schema\Schema;
  * Abstract class for individual migrations to extend from.
  *
  * @license     http://www.opensource.org/licenses/lgpl-license.php LGPL
+ *
  * @link        www.doctrine-project.org
  * @since       2.0
+ *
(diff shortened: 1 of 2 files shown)


mikeSimonson:
Hi @nicosomb, thanks for your PR but this feature already exists.
You can use the schema object and do $schema->getTable($tableName)->hasColumn($columnName).

Cheers

nicosomb:
oO 
Thank you for that. I looked for this feature in Doctrine Migration, not in Doctrine.

PR #501, opened 2016-12-09
Fix writeln call when no changes are to be done

Seldaek:
Without this fix I get:

```
  [Symfony\Component\Debug\Exception\ContextErrorException]
  Warning: A non-numeric value encountered
```

Because the OutputInterface expects options and not a string 'ERROR', I have no idea what this is supposed to do.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/DiffCommand.php
@@ -116,7 +116,7 @@ EOT
         );
 
         if (! $up && ! $down) {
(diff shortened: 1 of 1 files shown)


mikeSimonson:
@Seldaek Thanks

mikeSimonson:
That issue seems to be there since a while ...

Seldaek:
Yeah I suppose nobody but me runs this command without a good reason to (i.e. having changes..) :p

msvrtan:
@mikeSimonson any chance of new release that includes this patch?

mikeSimonson:
@msvrtan Done

msvrtan:
@mikeSimonson thank you!

Issue #468, opened 2016-06-14
Percona toolkit pt-online-schema-change integration


rossmotley:
Hi,

A really handy feature for us would be the integration of [Percona Toolkit](https://www.percona.com/doc/percona-toolkit/)'s [pt-online-schema-change](https://www.percona.com/doc/percona-toolkit/2.2/pt-online-schema-change.html) in some way. It's a great tool for executing migrations on large tables without blocking access to them, currently we're using the tool manually for migrations which need it.

It would be great if we could do something like this in our up/down methods:

``` php
$command = PerconaCommand::create("ADD new_field INT(10) DEFAULT NULL", "huge_table");

$this->addCommand($command);
```

Would this be useful for others?

Thanks,

Ross

There's a similar tool for ruby here: https://github.com/svarks/percona-migrations

mikeSimonson:
That would be perfect.

chrisguitarguy:
Maybe a solution here is to provide some way to hook into migrations and perform actions. Keep the core of the migrations library pretty simple, but let others extend it with whatever they like.

A simple set of events would probably be sufficient and there's already an event emitter provided by DBAL (`Connection::getEventManager`). Using that would also make the event listeners pretty easy to integrate with Symfony -- just tag things with `doctrine.event_{listener,subscriber}` in the container. There's probably something similar in the various other frameworks.

chrisguitarguy:
@rossmotley I made a small example of how something like this might work -- not with `pt-online-schema-change`, but with a more generic "run a cli command" type thing.

https://github.com/chrisguitarguy/cli-migrations-example

The basic gist is you [extend a custom migration base class](https://github.com/chrisguitarguy/cli-migrations-example/blob/f4a3fa89c728060c31604c76d0f377b11c1d5deb/migrations/Version00000000000001.php#L9) that would let you `addCommand` in the `up` and `down` methods. And an event listener would [hook in](https://github.com/chrisguitarguy/cli-migrations-example/blob/f4a3fa89c728060c31604c76d0f377b11c1d5deb/cli-config.php#L11) and [run those commands](https://github.com/chrisguitarguy/cli-migrations-example/blob/f4a3fa89c728060c31604c76d0f377b11c1d5deb/src/CliMigrationListener.php#L23-L40) after the version has been executed.

[...]

PR #505, opened 2016-12-23
Handle Connections with Auto Commit Off

### LINKED ISSUE #496: Executing migrations with auto commit off
If auto commit is disabled for the underlying dbal connection, migrations are never committed because the connection is already in nesting level 1. Each transactional migration starts a new transaction running in nesting level 2 (committed at the end, returning to level 1), but there is no commit for level 1 after all migrations are executed.

chrisguitarguy:
Possibly fixes #496 

This adds a test to verify that the issue is indeed the case and has a possible fix: adding a final `Connection::commit` call in `Migration::migrate` if auto commit is off. `Migration` being the actual application object, this seems like the right place for that final commit. Not really sure on that.

Also seems like the total time for the migrations should include the time of that final commit?

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Migration.php
+++ b/lib/Doctrine/DBAL/Migrations/Migration.php
@@ -19,6 +19,7 @@
 
 namespace Doctrine\DBAL\Migrations;
 
+use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Migrations\Configuration\Configuration;
 
 /**
@@ -44,6 +45,11 @@ class Migration
     private $configuration;
 
     /**
+     * @var Connection
+     */
+    private $connection;
+
+    /**
      * @var boolean
      */
     private $noMigrationException;
@@ -57,6 +63,7 @@ class Migration
     {
         $this->configuration = $configuration;
         $this->outputWriter = $configuration->getOutputWriter();
+        $this->connection = $configuration->getConnection();
         $this->noMigrationException = false;
     }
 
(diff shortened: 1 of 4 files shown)


mikeSimonson:
Hum, I don't really like that kind of magic. If the user pass a connection where the auto-commit has been set to off it's his responsibility to commit of not the connection after the migration.
If I merge your change it's not possible anymore to not commit the migration change even when you pass a connection with autocommit off which seems really odd.

chrisguitarguy:
That's fair, but it also means the command line tools are less usable for folks without auto commit on. Or simply less predictable -- MySQL's DDLs are transactional, for instance, if they were mixed in with other statement it might appear that only certain things work.

How about this as a solution:

Don't do anything in the core but add some overall "migrations started" and "migrations finished" events. And possibly provide an event listener that can do a final commit.

mikeSimonson:
> That's fair, but it also means the command line tools are less usable for folks without auto commit on. Or simply less predictable

Unless I am wrong autocommit is the default.

mikeSimonson:
I am not sure that pulling a whole event library is a reasonable solution here.
Is there a less invasive way to do it ?

chrisguitarguy:
> 
I am not sure that pulling a whole event library is a reasonable solution here.

There's already an event library in place [via the connection](https://github.com/doctrine/dbal/blob/b551fcb26a81e230085c0dca88174573b6c36609/lib/Doctrine/DBAL/Connection.php#L316-L319).

mikeSimonson:
@chrisguitarguy Ok then forget what I said.
I can accept a solution with an event and listener.

chrisguitarguy:
👍, closing this one then. I'll submit a PR with events only and a separate one with the auto commit listener. Events might help with [this issue as well](https://github.com/doctrine/migrations/issues/468#issuecomment-231365838).

PR #506, opened 2016-12-27
Add a Set of Events to Migrations

chrisguitarguy:
Per the discussion in #505 and #468

There are five events:

- `onMigrationsMigrating` - fired when the `Migrations::migrate` method is about to start executing versions.
- `onMigrationsMigrated` - fired when the `Migrations::migrate` method finishes executing all its versions.
- `onMigrationsVersionExecuting` - Fired immediately before a version does anything
- `onMigrationsVersionExecuted` - Fired when a version finishes its execution
- `onMigrationsVersionSkipped` - Fired when a version is skipped

Not sure on the naming. I figured more unique names would be better (hence the `onMigrations` prefix).

Maybe needs an event when a version fails?

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
@@ -19,6 +19,7 @@
 
 namespace Doctrine\DBAL\Migrations\Configuration;
 
+use Doctrine\Common\EventArgs;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Connections\MasterSlaveConnection;
 use Doctrine\DBAL\Migrations\Finder\MigrationDeepFinderInterface;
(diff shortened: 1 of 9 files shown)


mikeSimonson:
@chrisguitarguy Thanks for the PR.

I left you some questions and I think that we need to work on the naming a bit more.

What about Event and MigrationEvent or EventName and MigrationEvent ?

Thanks and Happy new year eve.

chrisguitarguy:
> What about Event and MigrationEvent or EventName and MigrationEvent ?

Do you mean for the [`Events`](https://github.com/chrisguitarguy/migrations/blob/70edf46e7f546df16567f9163f3fbe6320eb096b/lib/Doctrine/DBAL/Migrations/Events.php) and [`MigrationsEventArgs`](https://github.com/chrisguitarguy/migrations/blob/70edf46e7f546df16567f9163f3fbe6320eb096b/lib/Doctrine/DBAL/Migrations/Event/MigrationsEventArgs.php) respectively?

I'm good with whatever you'd like them to be named just let me know. As is I just followed what it looked like DBAL and ORM did.

chrisguitarguy:
Any guidance on the naming stuff @mikeSimonson? Do you want me to make the changes ore stick with how things are (in line with the ORM and DBAL)?

chrisguitarguy:
Any thoughts here? Again, I'm good with whatever. Tried to follow DBAL & ORM conventions initially.

mikeSimonson:
@chrisguitarguy Sorry for the delay.

PR #509, opened 2017-02-07
Add delta in migrate command

Safranil:
Hi all,

I needed to come back 3 migrations earlier and I don't found a rapid way to do it (without execute `migrations:migrate prev` 3 times). So I added a new dynamic alias like `git checkout HEAD~3`, for my case the command will be `migrations:migrate current-3`. You can go forward with `+` and backward with `-`.

Hope this can be help someone else :)

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
@@ -664,7 +664,7 @@ class Configuration
                     }
 
                     if ($symbol == "+" || $symbol == "-") {
-                        return $this->getRelativeVersion($this->getCurrentVersion(), intValue($symbol.$number));
+                        return $this->getRelativeVersion($this->getCurrentVersion(), (int)($symbol.$number));
                     }
                 }
                 return null;
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
@@ -655,6 +655,18 @@ class Configuration
             case 'latest':
                 return $this->getLatestVersion();
             default:
+                if (substr($alias, 0, 7) == 'current') {
+                    $symbol = substr($alias, 7, 1);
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
```diff
@@ -655,6 +655,18 @@ public function resolveVersionAlias($alias)
...
+                    $number = (int) substr($alias, 8);
+
+                    if ($number <= 0) {
+                        return null;
+                    }
+
+                    if ($symbol == "+" || $symbol == "-") {
+                        return $this->getRelativeVersion($this->getCurrentVersion(), (int)($symbol.$number));
+                    }
+                }
```
mikeSimonson: This logic should be abstracted in a function so that it's easily testable and that the switch stays readable.

You also need to add test before merge.

mikeSimonson:
@Safranil Thanks for your PR.

Can you add test ?

lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
Safranil: Ok, I'll move it in a new function.

For the test, I can try :)

Safranil:
OK, I moved the code in a new method and added the correct test.

mikeSimonson:
@Safranil Thanks

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php
```diff
@@ -58,6 +58,11 @@ protected function configure()
 You can specify the version you wish to migrate to using an alias:
 
     <info>%command.full_name% prev</info>
+    <info>These alias are defined : first, lastest, prev, current and next</info>
```
stof: typo: `latest`

lcobucci: @stof thanks, I'll fix it

PR #513, opened 2017-03-07
Show the sql statement parameters if they exist

mikeSimonson:


lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -407,7 +407,8 @@ private function executeRegisteredSql($dryRun = false, $timeAllQueries = false)
                         $this->outputWriter->write('     <comment>-></comment> ' . $query);
                         $this->connection->executeQuery($query);
                     } else {
-                        $this->outputWriter->write(sprintf('    <comment>-</comment> %s (with parameters)', $query));
+                        $parameters = implode(', ', $this->params[$key]);
+                        $this->outputWriter->write(sprintf('    <comment>-</comment> %s (%s)', $query, $parameters));
```
chrisguitarguy: This whole bit (including the `if/else` check for the params) could be replaced with [`$this->outputSqlQuery($key, $query)`](https://github.com/mikeSimonson/migrations/blob/1f9e1ed1af23565ac0028c9c379bc8342db96e3e/lib/Doctrine/DBAL/Migrations/Version.php#L437-L449).

I added that in #477 -- uses the doctrine types system to convert the params into something printable.

mikeSimonson: It would be better indeed.

mikeSimonson:
@chrisguitarguy Thanks for the review

lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -404,10 +404,10 @@ private function executeRegisteredSql($dryRun = false, $timeAllQueries = false)
                     $queryStart = microtime(true);
 
                     if ( ! isset($this->params[$key])) {
-                        $this->outputWriter->write('     <comment>-></comment> ' . $query);
+                        $this->outputSqlQuery($key, $query);
                         $this->connection->executeQuery($query);
                     } else {
-                        $this->outputWriter->write(sprintf('    <comment>-</comment> %s (with parameters)', $query));
+                        $this->outputSqlQuery($key, $query);
```
chrisguitarguy: The calls to `$this->outputSqlQuery` could be moved above the `if/else` block here.

lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -403,11 +403,10 @@ private function executeRegisteredSql($dryRun = false, $timeAllQueries = false)
                 foreach ($this->sql as $key => $query) {
                     $queryStart = microtime(true);
 
+                    $this->outputSqlQuery($key, $query);
```
sstalle: This will probably break migrations that use `\PDO::*` or `\Doctrine\DBAL\Connection::*` parameter types (see #528)

jwage: @mikeSimonson is this a problem?

mikeSimonson: This should have been fixed by #560. I will try to add a test for that specific case.

jwage:
This is implemented and covered by tests now.

PR #515, opened 2017-03-10
Introduce AutoCommitListener

### LINKED ISSUE #496: Executing migrations with auto commit off

chrisguitarguy:
Provided for folks who are using auto commit. By adding this listener to
their connection's event manager a finally `commit` call will be made
when migrations are complete.

Should close #496

See the discussion #505 for the rational here.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Event/Listeners/AutoCommitListener.php
+++ b/lib/Doctrine/DBAL/Migrations/Event/Listeners/AutoCommitListener.php
@@ -20,30 +20,17 @@
 namespace Doctrine\DBAL\Migrations\Event\Listeners;
 
 use Doctrine\Common\EventSubscriber;
-use Doctrine\DBAL\Version;
 use Doctrine\DBAL\Migrations\Events;
 use Doctrine\DBAL\Migrations\Event\MigrationsEventArgs;
 
-
 /**
  * Listens for `onMigrationsMigrated` and, if the conneciton is has autocommit
- * makes sure to do the final commit to make sure changes stick around.
+ * makes sure to do the final commit to ensure changes stick around.
  *
  * @since 1.6
  */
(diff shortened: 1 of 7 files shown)


lib/Doctrine/DBAL/Migrations/Event/Listeners/AutoCommitListener.php
```diff
@@ -0,0 +1,62 @@
...
+final class AutoCommitListener implements EventSubscriber
+{
+    public function __construct()
+    {
+        if (Version::compare('2.5') > 0) {
+            throw new \LogicException(sprintf(
+                'Autocommit was introduced in DBAL 2.5, version %s detected. You cannot use %s',
+                Version::VERSION,
+                __CLASS__
+            ));
```
chrisguitarguy: Wasn't really sure what to do about the version requirement here. Autocommit was introduced in DBAL 2.5, but migrations requires DBAL ~2.2. Thought just throwing a logic exception here was an okay way to let the user know they were doing something wrong, but not sure if that's okay.

mikeSimonson: I would just bump the minimum requirement in composer.

mikeSimonson:
@chrisguitarguy That way we can get rid of the if in the tests and dbal 2.5 is not exactly new so it shouldn't be an issue.

Thanks

chrisguitarguy:
@mikeSimonson sorry it took me a while to get back to this, but hopefully this is good to go now.

mikeSimonson:
@chrisguitarguy Thanks a lot

PR #523, opened 2017-03-31
Remove Unnecessary Docblocks from Generated Migrations

chrisguitarguy:
They only specified a type -- a type that was already apparent due to a typehint.


mikeSimonson:
@chrisguitarguy Thanks

PR #529, opened 2017-06-04
Support PDO::PARAM_* & Connection::PARAM_* in SQL Output

### LINKED ISSUE #528: --dry-run and --write-sql fail when a parameter of type \PDO::* or Connection::* is passed to addSql
Here is an example migration:
```php
<?php

namespace Application\Migrations;

use Doctrine\DBAL\Migrations\AbstractMigration;
use Doctrine\DBAL\Connection;

class Version20170602195852 extends AbstractMigration
{
    /**
     * @param Schema $schema
     */
    public function up(Schema $schema)
    {
        $this->addSql(
            'SELECT id FROM users WHERE id IN (:ids)',
            ['ids' => [1, 2]],
            ['ids' => Connection::PARAM_INT_ARRAY]
        );
    }

    /**
     * @param Schema $schema
     */
    public function down(Schema $schema)
    {
    }
}
```

It can be executed successfully with `migrations:execute 20170602195852`. But adding a `--dry-run` or `--write-sql` option makes it throw an exception:

> [Doctrine\DBAL\DBALException]
>  Unknown column type "101" requested. Any Doctrine type that you use has to be registered with \Doctrine\DBAL\Types\Type::addType(). You can get a list of all the known types with \Doctrine\DBAL\Types\Type::getTypesMap(). If this error occurs during database introspection then you might have forgot to register all database types for a Doctrine Type. Use AbstractPlatform#registerDoctrineTypeMapping() or have your custom types implement Type#getMappedDatabaseTypes(). If the type name is empty you might have a problem with the cache or forgot some mapping information.

Here is a simple [test case](https://github.com/sstalle/migrations/pull/1/files).

chrisguitarguy:
This uses the DBAL type system as a fallback when it can, and otherwise
runs the value through `json_encode` and hopes for the best. The
reasoning here is that on a dry or actual run shouldn't fail because
migrations doesn't know how to output a type.

Fixes #528 

I'm not 100% on this solution. Not sure if falling back to a DBAL type is a good idea -- maybe we should just `(string) $value` in those cases?

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Version.php
+++ b/lib/Doctrine/DBAL/Migrations/Version.php
@@ -19,6 +19,8 @@
 
 namespace Doctrine\DBAL\Migrations;
 
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\DBAL\Migrations\Configuration\Configuration;
 use Doctrine\DBAL\Migrations\Event\MigrationsVersionEventArgs;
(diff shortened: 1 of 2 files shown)


chrisguitarguy:
#560 too care of #528, closing this one.

PR #533, opened 2017-06-18
Generate migrations with declared strict types

radekhubner:


lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
```diff
@@ -38,7 +38,7 @@ class GenerateCommand extends AbstractCommand
 {
 
     private static $_template =
-            '<?php
+            '<?php declare(strict_types = 1);
```
alcaeus: Master still allows for PHP 5, which could cause errors for some people. I'd suggest dropping 5.x support from master first, then adding this in.

radekhubner: Thank you

fidraj: It's been dropped https://github.com/doctrine/migrations/blob/20dc1f98387ae33bba0f6a621815750b518d98f8/composer.json#L14

mikeSimonson:
@radekhubner Thanks

PR #534, opened 2017-07-12
Change RecursiveRegexFinder to follow symlinks.

burci:
We have a monolithic repository with shared entities and migrations between different instances of the default implementation.

The core entities related migrations are in a separated shared folder which we are symlinking into the migration folder (the namespace is the same).

Project structure:
```
vendor/something/something-migrations/DoctrineMigrations/
        Version20150101000000.php
        Version20150101000001.php
       ...

app /
    DoctrineMigrations /
        shared -> ../../vendor/something/something-migrations/DoctrineMigrations/
        Version20160101000000.php
        Version20160101000001.php
        ...
```

This change in RecursiveRegexFinder solves to handle the symlink.

### DIFF
--- a/tests/Doctrine/DBAL/Migrations/Tests/Finder/RecursiveRegexFinderTest.php
+++ b/tests/Doctrine/DBAL/Migrations/Tests/Finder/RecursiveRegexFinderTest.php
@@ -34,7 +34,7 @@ class RecursiveRegexFinderTest extends MigrationTestCase
     {
         $migrations = $this->finder->findMigrations(__DIR__ . '/_files', 'TestMigrations');
 
-        self::assertCount(6, $migrations);
+        self::assertCount(7, $migrations);
 
         $tests = [
             '20150502000000' => 'TestMigrations\\Version20150502000000',
@@ -43,6 +43,7 @@ class RecursiveRegexFinderTest extends MigrationTestCase
             '20150502000004' => 'TestMigrations\\Version20150502000004',
             '20150502000005' => 'TestMigrations\\Version20150502000005',
             '1_reset_versions' => 'TestMigrations\\Version1_reset_versions',
+            '1_symlinked_file' => 'TestMigrations\\Version1_symlinked_file',
         ];
         foreach ($tests as $version => $namespace) {
             self::assertArrayHasKey($version, $migrations);
--- a/tests/Doctrine/DBAL/Migrations/Tests/Finder/_files/_symlinked_files
+++ b/tests/Doctrine/DBAL/Migrations/Tests/Finder/_files/_symlinked_files
@@ -1,0 +1,1 @@
+../_symlinked_files/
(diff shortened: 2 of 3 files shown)


mikeSimonson:
@burci Thanks for the PR

You know that you can use your finder right ?

If you want me to consider this PR You would need to add a test case that show that it effectively works.

burci:
@mikeSimonson I am using doctrine migration bundle, I did not find any option to override the finder. 
If there is any way, I can solve my issue that way as well, but I can write tests for this pull request too.

chrisguitarguy:
@burci You can can call [`setMigrationsFinder` on the configuration](https://github.com/doctrine/migrations/blob/v1.5.0/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php#L366-L378) to swap in a custom implementation.

I'm not really sure how would work with the migrations bundle, though.

tests/Doctrine/DBAL/Migrations/Tests/Finder/RecursiveRegexFinderTest.php
```diff
@@ -63,6 +63,7 @@ public function testFindMigrationsReturnsTheExpectedFilesFromDirectory()
             '20150502000004' => 'TestMigrations\\Version20150502000004',
             '20150502000005' => 'TestMigrations\\Version20150502000005',
             '1_reset_versions' => 'TestMigrations\\Version1_reset_versions',
+            '1_symlinked_file' => 'TestMigrations\\Version1_symlinked_file',
```
chrisguitarguy: I'd suggest adding a whole new test (and set of fixture files) for the new feature here.

burci: That function is going to be exactly the same, and the test name is `testFindMigrationsReturnsTheExpectedFilesFromDirectory` which quite well describes this new test as well.

What about calling this test two times, with different dirs (using dataProvider) and from the second directory I would symlink the original directory?

chrisguitarguy: Yes, it is pretty much the same but you're testing two different things:

1. the behavior without any symlinks
2. the behavior with symlinks

That difference is enough to deserve another test with a name that reflects what you're testing. That's my opinion, anyway. @mikeSimonson is the maintainer here, so it's ultimately his call.

To make it simpler you could make a new directory and symlink the `_files` directory into it and make sure the symlink is followed.

mikeSimonson:
As long as there is a test I am happy.

burci:
@mikeSimonson it is there! :)

tests/Doctrine/DBAL/Migrations/Tests/Finder/_files/_symlinked_files
```diff
@@ -0,0 +1 @@
+../_symlinked_files/
```
Majkl578: Will this work on Windows? Maybe _skip-if_ should be added for that poor thing? :)

alcaeus: Probably not, but then our test suite doesn't run on Windows right now and I'd expect the directory iterators to know how to handle Symlinks on Windows ;)

alcaeus:
@burci could you rebase this branch on top of master? I'd like to get this merged soon so we can release it in 1.6.0. Thanks!

alcaeus:
@burci never mind - I've taken the liberty to rebase and force push since you enabled maintainer access to the branch. I'll merge this once the build is stable, thank you!

PR #535, opened 2017-07-27
Ability to change the class template

PapsOu:
Related to this issue : https://github.com/doctrine/DoctrineMigrationsBundle/issues/199

Related to PR : https://github.com/doctrine/DoctrineMigrationsBundle/pull/200

Here's the PR that introduce custom template for command doctrine:migrations:generate.
It also add a command option in order to set custom template just for the current command.

E.G: bin/console do:mi:ge --template src/MyBundle/Resources/tpl/migration.tpl

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
@@ -89,7 +89,7 @@ EOT
     {
         $configuration = $this->getMigrationConfiguration($input, $output);
 
-        $this->manageCustomTemplate($configuration, $input, $output);
+        $this->loadCustomTemplate($configuration, $output);
 
         $version = $configuration->generateVersionNumber();
         $path    = $this->generateMigration($configuration, $input, $version);
@@ -137,31 +137,29 @@ EOT
         return $path;
     }
 
-    protected function manageCustomTemplate(
-        Configuration $configuration,
-        InputInterface $input,
-        OutputInterface $output
-    ) : void {
+    private function loadCustomTemplate(Configuration $configuration, OutputInterface $output) : void
+    {
         $customTemplate = $configuration->getCustomTemplate();
 
-        if ($customTemplate !== null) {
-            $this->loadTemplateFile($customTemplate, $output);
+        if ($customTemplate === null) {
+            return;
         }
-    }
 
-    private function loadTemplateFile(string $customTemplate, OutputInterface $output) : void
-    {
-        $customTemplate = getcwd() . '/' . $customTemplate;
+        $filePath = getcwd() . '/' . $customTemplate;
 
-        if ( ! is_file($customTemplate) || ! is_readable($customTemplate)) {
+        if ( ! is_file($filePath) || ! is_readable($filePath)) {
             throw new \InvalidArgumentException(
                 'The specified template "' . $customTemplate . '" cannot be found or is not readable.'
             );
         }
 
-        $templateContent = file_get_contents($customTemplate);
+        $content = file_get_contents($filePath);
+
+        if ($content === false) {
+            throw new \InvalidArgumentException('The specified template "' . $customTemplate . '" could not be read.');
+        }
 
         $output->writeln(sprintf('Using custom migration template "<info>%s</info>"', $customTemplate));
-        $this->instanceTemplate = $templateContent;
+        $this->instanceTemplate = $content;
     }
 }
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
@@ -137,8 +137,11 @@ EOT
         return $path;
     }
 
-    protected function manageCustomTemplate(Configuration $configuration, InputInterface $input, OutputInterface $output)
-    {
+    protected function manageCustomTemplate(
+        Configuration $configuration,
+        InputInterface $input,
+        OutputInterface $output
+    ) : void {
         $customTemplate = $configuration->getCustomTemplate();
 
         if ($customTemplate !== null) {
@@ -146,9 +149,10 @@ EOT
         }
     }
 
-    private function loadTemplateFile(string $customTemplate, OutputInterface $output): void
+    private function loadTemplateFile(string $customTemplate, OutputInterface $output) : void
     {
         $customTemplate = getcwd() . '/' . $customTemplate;
+
         if ( ! is_file($customTemplate)) {
             throw new \InvalidArgumentException('The specified template « ' . $customTemplate . ' » cannot be found.');
         }
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
@@ -36,6 +36,8 @@ use Symfony\Component\Console\Input\InputOption;
  */
 class GenerateCommand extends AbstractCommand
 {
+    private $_instanceTemplate;
+
     private static $_template =
             '<?php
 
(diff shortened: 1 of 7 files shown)


lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
```diff
@@ -127,4 +136,30 @@ protected function generateMigration(Configuration $configuration, InputInterfac
...
+    protected function manageCustomTemplate(Configuration $configuration, InputInterface $input, OutputInterface $output)
+    {
+        if ($tpl = $input->getOption('template')) {
+            $this->loadTemplateFile($tpl, $output);
+        } elseif ($configuredTemplate = $configuration->getCustomTemplate()) {
+            $this->loadTemplateFile($configuredTemplate, $output);
+        }
+    }
+
+    private function loadTemplateFile($tpl, OutputInterface $output)
```
lcobucci: Please don't abbreviate variables

lcobucci: Also add type declarations:

```php
private function loadTemplateFile(string $template, OutputInterface $output): void
```

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
```diff
@@ -127,4 +136,30 @@ protected function generateMigration(Configuration $configuration, InputInterfac
...
+    {
+        $tpl = getcwd() . '/' . $tpl;
+        if(!is_file($tpl)) {
+            throw new \InvalidArgumentException('The specified template « ' . $tpl . ' » cannot be found.');
+        }
+
+        $tplContent = file_get_contents($tpl);
```
lcobucci: Don't abbreviate here as well

lcobucci (changes requested):
It's a nice addition indeed, can you please address these things?

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
```diff
@@ -127,4 +136,30 @@ protected function generateMigration(Configuration $configuration, InputInterfac
 
         return $path;
     }
+
+    protected function manageCustomTemplate(Configuration $configuration, InputInterface $input, OutputInterface $output)
+    {
+        if ($tpl = $input->getOption('template')) {
```
PapsOu: Using the null coalesce operator is only supported by php 7. Avoid using it for php 5.6 should be a good thing.

lcobucci: Most of doctrine projects are dropping support for PHP < 7.1, we simply forgot to do it on doctrine/migrations. See #537

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
PapsOu: Is php 5.6 support is abandonned ?

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
PapsOu: OK well, I will use php 7 syntaxe when needed

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
lcobucci: Bug fixes will be backported but new feature will be PHP 7.1+

PapsOu:
@lcobucci : I've taken into account your remarks.

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
```diff
@@ -127,4 +136,30 @@ protected function generateMigration(Configuration $configuration, InputInterfac
...
+        }
+
+        $templateContent = file_get_contents($customTemplate);
+
+        if ($templateContent === false) {
+            throw new \InvalidArgumentException('Cannot read file contents of template « ' . $customTemplate . ' ».');
+        }
+
+        $output->writeln(sprintf('Using custom migration template "<info>%s</info>"', $customTemplate));
+        self::$_template = $templateContent;
```
stof: I'm not sure overwriting the static property containing the default template is a good idea. A mutable template should be restricted to the current instance

PapsOu: @stof : I agree this should be only on object instance, but it need to check every methods of current command to avoid a break.

I'll see this point when comming back from holidays (in a week)

PapsOu:
The unit test is failing. please do not merge this PR for now.

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
PapsOu: @stof : 5325fa5 introduce instance template property.

PapsOu:
Unit tests are OK with php 7.1 (5.6 and 7.0 are failing)

The @stof's remark has been added on commit 5325fa5

@lcobucci : I've done all of your reviews remarks.

lcobucci:
@PapsOu awesome, I'm fixing the version stuff on #537 and things will be just fine 😉

lcobucci (changes requested):
@PapsOu LGTM, just two small things that we forgot 😄

lcobucci:
@PapsOu @mikeSimonson @stof I was wondering, what do you think about keeping this change on the configuration level only? For me it doesn't make much sense to override the template for a specific execution.

Removing that possibility also means less changes in the codebase, which is always good IMO 😜

mikeSimonson:
@PapsOu Can you rebase this or do you prefer that we do it ?

PapsOu:
I can remove the command option if you want. I will do that in few hours

PapsOu:
@mikeSimonson @lcobucci : Everything's done

lcobucci (approved):
Awesome, great work. 

Since I have some time now, I'm going to edit your branch to add the final nitpickings and fix the CS violations that Travis is complaining about 😉

lcobucci:
@PapsOu 🚢 thanks for your contribution

PapsOu:
@lcobucci : You're welcome. Good luck for fitting the CS ;)

alister:
This change, with the strong type requirement for setting (and a nullable getter), breaks the DoctrineMigrationsBundle, because the configuration there has a default value for the 'doctrine_migrations.custom_template' parameter of null.

https://github.com/doctrine/DoctrineMigrationsBundle/blob/582f503e1f92236da4772c58e53ac06aab534f96/DependencyInjection/Configuration.php#L45

lcobucci:
@alister this is already being handled by https://github.com/doctrine/migrations/pull/545

enumag:
@PapsOu Is there a reason why the template is used only for Generate command but not for Diff command?

PapsOu:
I ve forgive to handle the diff command.

But I don't have much time to do that addition.

alcaeus:
@PapsOu tracked in #566. I'd try to release 1.6 this or next week and I'd like to see the feature completed. Do you want to give it a shot? Otherwise I'll try to find some time for it.

PapsOu:
I hope adding this feature in the week or maybe this week end.

alcaeus:
Awesome, thank you!

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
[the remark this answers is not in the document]
wayne530: Does anyone know what happened to this `--template` option for the generate command? It seems to have been removed?

wayne530 (commented):
Is there a discussion of how we came to the conclusion that this option is no longer needed? Some projects will have multiple templates and changing the configuration to select one vs. another is tedious vs. a command-line option.

Issue #514, opened 2017-03-07
utf8mb4 character set causes migrations table creation to fail


jaikdean:
I have a project using MySQL 5.6 set up to use `utf8mb4` character set and `utf8mb4_unicode_ci` collation. Having this set means the `migration_versions` table fails to be created with the following error:

`1071 Specified key was too long; max key length is 767 bytes`

The `version` field length of 255 is hardcoded in `Doctrine\DBAL\Migrations\Configuration\Configuration::createMigrationTable()`, as is creation of the table itself, so there is no way to override the character set or adjust the field length without overriding this method.

mikeSimonson:
That's one of the reason why utf8mb4_general_ci is recommended over utf8mb4_unicode_ci.
Nonetheless if you feel like it, I would welcome a PR to configure that.
You would need to enforce that new limitation in the MigrationFinder too thought.

lcobucci:
Related to #538

jwage:
Closing this as you can now set the column length.

PR #538, opened 2017-08-17
#514 -- Add support for configurable migration column length.

SlvrEagle23:
Because the Migrations Configuration object's `createMigrationTable()` function references private properties, it is difficult if not impossible to override on a per-project basis if necessary to specify a user-configurable string length. Adding another private property seems like the most low-friction way of resolving the issue.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
@@ -112,6 +112,15 @@ class Configuration
     private $migrationsColumnName = 'version';
 
     /**
+     * The length of the string field in the migrations table. This can often be left at its default
+     * value, but can be modified if necessary, for example, to 191 for early MySQL servers using the
+     * UTF8MB4 character set.
+     *
+     * @var int
+     */
+    private $migrationsColumnLength = 255;
+
+    /**
      * The path to a directory where new migration classes will be written
      *
      * @var string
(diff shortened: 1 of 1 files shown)


mikeSimonson:
@SlvrEagle23 Can you add a test ?

Thanks for the PR

lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
```diff
@@ -319,6 +328,26 @@ public function getMigrationsColumnName()
     }
 
     /**
+     * Set the length of the migration string column.
+     *
+     * @param int $columnLength The migration string column length.
+     */
+    public function setMigrationsColumnLength($columnLength)
```
lcobucci: Missing type declaration

lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
```diff
@@ -319,6 +328,26 @@ public function getMigrationsColumnName()
...
+    {
+        $this->migrationsColumnLength = $columnLength;
+    }
+
+    /**
+     * Returns the length of the migration string column.
+     * 
+     * @return int $migrationsColumnLength The migration string column length.
+     */
+    public function getMigrationsColumnLength()
```
lcobucci: Missing type declaration

lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
SlvrEagle23: The codebase I'm PR'ing against doesn't have type declarations, so I was trying to stay consistent with that.

lcobucci: The codebase has already been changed to require PHP 7.1 😉

jwage:
@SlvrEagle23 do you want to try updating this for 2.0? Do you mind if I update it and merge?

SlvrEagle23:
@jwage If you want to take it on, by all means I don't mind. I don't have the same dev environment set up as I did when I submitted it, and I updated much of my own code to use MariaDB 10.2, which allows the longer fields.

PR #545, opened 2017-09-02
Allow nullable customTemplate in Configuration setter

PapsOu:
Related to this discussion https://github.com/doctrine/DoctrineMigrationsBundle/commit/11459465d5049812ea83003a761b9787bf01ec0f#commitcomment-24042795

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
@@ -387,7 +387,7 @@ class Configuration
      *
      * @param string $customTemplate The custom template path
      */
-    public function setCustomTemplate(string $customTemplate) : void
+    public function setCustomTemplate(?string $customTemplate) : void
     {
         $this->customTemplate = $customTemplate;
     }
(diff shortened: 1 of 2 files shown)


lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
```diff
@@ -387,7 +387,7 @@ public function getCustomTemplate() : ?string
      *
      * @param string $customTemplate The custom template path
      */
-    public function setCustomTemplate(string $customTemplate) : void
+    public function setCustomTemplate(string $customTemplate = null) : void
```
Majkl578: Should be `?string`.

PapsOu:
@Majkl578 thanks I forgot that nullable type was applicable also on method parameters.

lcobucci:
@PapsOu it's good now, can you please cover that scenario with a test to ensure we don't break it in the future?

PapsOu:
Ok I will do it this week

celowsky:
The issue that Scrutinizer is reporting does not seem valid. The code style complaint regard the PSR-2 is made on code that seems to be following that standard.

Majkl578:
Don't worry about Scrutinizer, sometimes it can be annoying. 😉 Simple test is still needed though...

PapsOu:
I don't know if the test case is well writen (I'm not really conformtable with Unit testing).

lcobucci:
@PapsOu I had some time now and I've managed to simplify a bit your test. Thanks a lot for the contribution!

PR #549, opened 2017-09-13
use yoda comparison

OskarStark:

PR #551, opened 2017-09-18
use yoda style comparison in generated migrations

dbu:
small nitpick, but it does not hurt to have generated code follow best practices.

(and popular code sniffers complain when code is not doing yoda style comparisons...)


greg0ire:
@dbu did you see #549 ?

dbu:
no. pitty, i am in the camp of ocramious on that question. but whatever, i guess we can configure our code style checker to ignore migrations, not a huge deal.

PR #555, opened 2017-10-02
Use schema config for migrations table

aik099:
When creating `migration_versions` table use same table options (e.g. `charset`, `collate` and `engine`), that are used during migration diff generation.

Solves the problem, where `doctrine.dbal.default_table_options` setting wasn't respected by https://github.com/doctrine/DoctrineMigrationsBundle.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
@@ -31,6 +31,7 @@ use Doctrine\DBAL\Migrations\Version;
 use Doctrine\DBAL\Migrations\Finder\MigrationFinderInterface;
 use Doctrine\DBAL\Migrations\Finder\RecursiveRegexFinder;
 use Doctrine\DBAL\Schema\Column;
+use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Types\Type;
 
@@ -782,18 +783,18 @@ class Configuration
         }
 
         $this->connect();
-        if ($this->connection->getSchemaManager()->tablesExist([$this->migrationsTableName])) {
+        $schemaManager = $this->connection->getSchemaManager();
+        if ($schemaManager->tablesExist([$this->migrationsTableName])) {
             $this->migrationTableCreated = true;
 
             return false;
         }
 
-        $columns = [
-            $this->migrationsColumnName => new Column($this->migrationsColumnName, Type::getType('string'), ['length' => 255]),
-        ];
-        $table   = new Table($this->migrationsTableName, $columns);
+        $schema = new Schema([], [], $schemaManager->createSchemaConfig());
+        $table  = $schema->createTable($this->migrationsTableName);
+        $table->addColumn($this->migrationsColumnName, 'string', ['length' => 255]);
         $table->setPrimaryKey([$this->migrationsColumnName]);
-        $this->connection->getSchemaManager()->createTable($table);
+        $schemaManager->createTable($table);
 
         $this->migrationTableCreated = true;
 


lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
```diff
@@ -782,18 +783,18 @@ public function createMigrationTable()
...
             return false;
         }
 
-        $columns = [
-            $this->migrationsColumnName => new Column($this->migrationsColumnName, Type::getType('string'), ['length' => 255]),
-        ];
-        $table   = new Table($this->migrationsTableName, $columns);
+        $schema = new Schema([], [], $schemaManager->createSchemaConfig());
+        $table = $schema->createTable($this->migrationsTableName);
+        $table->addColumn($this->migrationsColumnName, 'string', ['length' => 255]);
```
mikeSimonson: Can you keep the getType here ?

mikeSimonson:
@aik099 Thanks for the PR

lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
aik099: When I change `string` into `Type::getType('string')` I've getting an exception back, because `\Doctrine\DBAL\Schema\Table::addColumn` method expects type as string. Internally it's doing the same thing: https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Schema/Table.php#L321

mikeSimonson: ok

lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
```diff
@@ -782,18 +783,18 @@ public function createMigrationTable()
...
 
             return false;
         }
 
-        $columns = [
-            $this->migrationsColumnName => new Column($this->migrationsColumnName, Type::getType('string'), ['length' => 255]),
-        ];
-        $table   = new Table($this->migrationsTableName, $columns);
+        $schema = new Schema([], [], $schemaManager->createSchemaConfig());
+        $table = $schema->createTable($this->migrationsTableName);
```
mikeSimonson: The build is failling because there is a space missing before the equal to align it with the one from the line above.

aik099: Fixed.

lcobucci:
@aik099 Patch LGTM btw 👍

aik099:
@lcobucci , any example?

mikeSimonson:
Look here https://github.com/doctrine/migrations/blob/master/tests/Doctrine/DBAL/Migrations/Tests/Functional/CliTest.php

aik099:
Almost all commands use (and create if missing) migrations table and I've changed code, used to create it.

lcobucci:
> Almost all commands use (and create if missing) migrations table and I've changed code, used to create it.

Sure, but no automated test cover that specific scenario. So it means that you're not breaking existing code but you're not adding tests that ensure that in the future no one will break what you've done. Got it?

aik099:
> Sure, but no automated test cover that specific scenario.

Yes, none of the tests likely check actual structure of the table being created after the `migrate:status` command was executed. To test it I'm afraid, that real MySQL database is needed because charset & collation stuff might not manifest itself in say SQLLite database.

mikeSimonson:
I thinking to a test where you would pass a schema where you changed the charset and you verify that after execution it's that charset that was used for the new code. I will try that tomorrow.

aik099:
@mikeSimonson , did it work, I mean idea about new test?

mikeSimonson:
@aik099 I taught that you where going to take a look at it. If you can't, just tell me and I will try.

aik099:
Yeah, I can't, because I haven't found any tests that are using MySQL or similar database engine where we can actually dump database structure and verify that it was using correct collation, charset and type (InnoDB or MyISAM).

jwage:
Since master has changed so much since this PR was originally created, I continued the work for this in #679

PR #561, opened 2017-10-29
Fix docblocks

mikeSimonson:


lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
```diff
@@ -638,7 +638,9 @@ public function getNextVersion()
     /**
      * Returns the version with the specified offset to the specified version.
      *
-     * @return string|null A version string, or null if the specified version
+     * @param $version string
```
alcaeus: The type precedes the variable name. This goes for all param documentation in this PR.

lib/Doctrine/DBAL/Migrations/FileQueryWriter.php
```diff
@@ -51,6 +51,10 @@ public function __construct(string $columnName, string $tableName, ?OutputWriter
 
     /**
      * TODO: move SqlFileWriter's behaviour to this class - and kill it with fire (on the next major release)
+     * @param string $path
```
alcaeus: You might as well omit these as the method signature already contains type information.

lib/Doctrine/DBAL/Migrations/QueryWriter.php
```diff
@@ -28,7 +28,10 @@
 interface QueryWriter
 {
     /**
-     * @throws \Doctrine\DBAL\Exception\InvalidArgumentException
+     * @param string $path
```
alcaeus: Can be omitted

lib/Doctrine/DBAL/Migrations/Tools/Console/Helper/MigrationStatusInfosHelper.php
```diff
@@ -77,7 +77,9 @@ private function getFormattedVersionAlias($alias)
         return $this->configuration->getDateTime($version) . ' (<comment>' . $version . '</comment>)';
     }
 
-    /** @var Version[] */
+    /**
+     * @return array|Version[]
```
alcaeus: The array return type is useless when you add Version[]

lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
mikeSimonson: Weird are you sure ? Because most of those were auto generated by PHPStorm.

alcaeus: See other examples in the same file.

PR #562, opened 2017-10-29
Allow to provide a path for the sqlFile dump

mikeSimonson:
This is a weird one because there is already a test for this
but the test passed. Might be an issue with symfony or an impossibility
to test that behavior.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php
@@ -43,7 +43,7 @@ class MigrateCommand extends AbstractCommand
             ->setName('migrations:migrate')
             ->setDescription('Execute a migration to a specified version or the latest available version.')
             ->addArgument('version', InputArgument::OPTIONAL, 'The version number (YYYYMMDDHHMMSS) or alias (first, prev, next, latest) to migrate to.', 'latest')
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php
```diff
@@ -43,7 +43,7 @@ protected function configure()
             ->setName('migrations:migrate')
             ->setDescription('Execute a migration to a specified version or the latest available version.')
             ->addArgument('version', InputArgument::OPTIONAL, 'The version number (YYYYMMDDHHMMSS) or alias (first, prev, next, latest) to migrate to.', 'latest')
-            ->addOption('write-sql', null, InputOption::VALUE_NONE, 'The path to output the migration SQL file instead of executing it.')
+            ->addOption('write-sql', null, InputOption::VALUE_OPTIONAL, 'The path to output the migration SQL file instead of executing it.')
```
alcaeus: Where is the file written when no value is provided?

mikeSimonson: https://github.com/doctrine/migrations/blob/218fce0cbe0f7070ab3d66e4a31a529fa6a1d8d2/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php#L131

To the current working directory, I will add it in the description.

alcaeus: Ah, didn't see that. Thanks!

PR #565, opened 2017-11-04
Adding Symfony 4 support

weaverryan:
The DoctrineMigrationsBundle is currently uninstallable in Symfony 4 - as we need to allow Symfony 4 support in this library. I don't expect any issues, but let's see if the tests pass :).


mikeSimonson:
Thanks @weaverryan

weaverryan:
Thanks @mikeSimonson! Any change you might be able to release a new version soon? :)

PR #568, opened 2017-11-07
Add ConfigurationHelperInterface

alcaeus:
I had this problem a few days ago: needed to override the `ConfigurationHelper` to be able to inject a custom finder implementation in DoctrineMigrationsBundle. Since the class is marked as `internal` (and should probably be `final` in 2.0), it makes sense adding an interface to allow people to supply their own configuration helper.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/AbstractCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/AbstractCommand.php
@@ -9,6 +9,7 @@ use Doctrine\DBAL\Migrations\Configuration\Connection\Loader\ConnectionHelperLoa
 use Doctrine\DBAL\Migrations\Configuration\Connection\Loader\ConnectionConfigurationChainLoader;
 use Doctrine\DBAL\Migrations\OutputWriter;
 use Doctrine\DBAL\Migrations\Tools\Console\Helper\ConfigurationHelper;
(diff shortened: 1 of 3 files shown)


lib/Doctrine/DBAL/Migrations/Tools/Console/Command/AbstractCommand.php
```diff
@@ -88,7 +89,7 @@ protected function getMigrationConfiguration(InputInterface $input, OutputInterf
     {
         if ( ! $this->migrationConfiguration) {
             if ($this->getHelperSet()->has('configuration')
-                && $this->getHelperSet()->get('configuration') instanceof ConfigurationHelper) {
+                && $this->getHelperSet()->get('configuration') instanceof ConfigurationHelperInterface) {
                 $configHelper = $this->getHelperSet()->get('configuration');
```
weaverryan: If the `configuration` helper set exists, but it does *not* implement `ConfigurationHelperInterface`, could we throw an Exception instead? It seems like a misconfiguration - and it would silently fallback to the normal one.

alcaeus: As much as I'd love to throw an exception, `configuration` is too generic to assume that there should only be a migrations configuration helper in there. An exception might be overkill, but we should think about how to make this more specific.

weaverryan: Hmm, that makes sense, ok :)

PR #569, opened 2017-11-07
Allow using absolute paths for custom templates

### LINKED ISSUE #567: Custom template cannot be configured with absolute path
Related to https://github.com/doctrine/migrations/pull/535.

There is one more issue with the custom templates that I'd like to have fixed in 1.6. When I use doctrine-migrations-bundle and configure it like this:

```yml
doctrine_migrations:
    dir_name: '%kernel.project_dir%/src/Core/Infrastructure/Doctrine/Version'
    namespace: App\Core\Infrastructure\Doctrine\Version
    table_name: _db_version
    name: Application Migrations
    organize_migrations: BY_YEAR_AND_MONTH
    custom_template: '%kernel.project_dir%/src/Library/Resource/migration.tpl'
```

Then it fails with this exception:

```
  [InvalidArgumentException]                                                                                                   
  The specified template "/home/enumag/Projects/foobar/src/Library/Resource/migration.tpl" cannot be found or is not readable.
```

The reason is that the command always [prepends](https://github.com/doctrine/migrations/blob/master/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php#L131) the current working directory. It should only do so if the path is relative but not when it's absolute.

cc @PapsOu

@alcaeus Can I ask you to add the 1.6 milestone here?

alcaeus:
Fixes #567.

@enumag care giving this a quick spin? This should fix your issue.

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
@@ -128,15 +128,13 @@ EOT
             return;
         }
 
-        $filePath = getcwd() . '/' . $customTemplate;
-
-        if ( ! is_file($filePath) || ! is_readable($filePath)) {
+        if ( ! is_file($customTemplate) || ! is_readable($customTemplate)) {
             throw new \InvalidArgumentException(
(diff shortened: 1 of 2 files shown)


enumag:
Yep, exactly what I was thinking. 👍

PR #577, opened 2017-12-07
Add void return type for migration template.

Aliance:


alcaeus:
I'm conflicted: on one hand, I absolutely agree (and have made this change myself using a custom template); on the other hand, the contract in `AbstractMigration` does not specify the return type (because it would be a BC break) while PHP (currently) allows adding the `void` return type hint.

What do you think @lcobucci?

Majkl578:
I think this PR is fine, adding return types to methods where parents have none is valid and part of the [RFC](https://wiki.php.net/rfc/return_types):

> If the parent does not declare a return type then the child is allowed to declare one.

PHP won't change this behavior, it would be a massive BC break.

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
```diff
@@ -32,13 +32,13 @@ class GenerateCommand extends AbstractCommand
  */
 class Version<version> extends AbstractMigration
 {
-    public function up(Schema $schema)
+    public function up(Schema $schema): void
```
Majkl578: To conform Doctrine CS, there should be a space around `:`.

alcaeus:
> PHP won't change this behavior, it would be a massive BC break

Sounds good!

@Aliance please rebase on top of master branch, that will fix build failures.

Aliance:
@alcaeus done

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
Aliance: This is a generated code, used in other projects; not part of doctrine sources. May be leave it as is, like proposed [PSR-12](https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md) does?

alcaeus: I agree with @Aliance here. Most code I've seen and written, as well as the current draft of PSR-12 use this style:
> When you have a return type declaration present there MUST be one space after the colon followed by the type declaration. The colon and declaration MUST be on the same line as the argument list closing parentheses with no spaces between the two characters. The declaration keyword (e.g. string) MUST be lowercase.

I also don't remember any discussion regarding the style of this in the doctrine styleguide.

Majkl578: It's still part of our code base so it should follow the same CS, but I don't really care, declare is currently wrong as well.

> PSR-12

I know nothing about PSR-12, haven't ever seen it used anywhere.

> I also don't remember any discussion regarding the style of this in the doctrine styleguide.

We had some random discussions in comments here on GitHub. Space around colon is part of our CS already (according to readme in CS repo).

lcobucci: I agree with @Majkl578 if it's part of our code it must follow our guidelines.

lcobucci:
@alcaeus PR is indeed fine, just need the CS fix

Aliance:
@lcobucci okay, done.

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
stof: @lcobucci technically, the generated code is part of the user codebase, not of the Doctrine one.

Majkl578: @stof True, but you can't realistically accommodate needs of all consumers. So the best we can do here is consistency.
Also the template is configurable so anyone can provide their own custom template:
https://github.com/doctrine/migrations/blob/d3e8b903233d5fb7f00193c95eec501f27c0b908/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php#L139-L144

PR #583, opened 2017-12-13
Make version class final

### LINKED ISSUE #582: Generated migration classes should be final
Because there is no reason to extend a migration class AFAIK.

https://ocramius.github.io/blog/when-to-declare-classes-final/

Soullivaneuh:
Closes #582


lcobucci:
@Soullivaneuh 🚢  thanks!

PR #584, opened 2017-12-15
Add PhpDoc tags

zk2:
Just for the convenience of working in PhpStorm

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
+++ b/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
@@ -24,6 +24,7 @@ class GenerateCommand extends AbstractCommand
 
 namespace <namespace>;
 
+use Doctrine\DBAL\Migrations\AbortMigrationException;
 use Doctrine\DBAL\Migrations\AbstractMigration;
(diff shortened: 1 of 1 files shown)


lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
```diff
@@ -32,12 +33,22 @@ class GenerateCommand extends AbstractCommand
  */
 final class Version<version> extends AbstractMigration
 {
+    /**
+     * @param Schema $schema
```
Majkl578: This adds no information to what is already expressed by native type. Please remove it.

lib/Doctrine/DBAL/Migrations/Tools/Console/Command/GenerateCommand.php
```diff
@@ -32,12 +33,22 @@ class GenerateCommand extends AbstractCommand
...
+     *
+     * @throws AbortMigrationException
```
alcaeus: This should be added if you abort the migration. By default, this doesn't throw an exception (see the empty method body).

alcaeus:
Closing as per review above.

Issue #521, opened 2017-03-29
Dry run creates migration_versions table


tehplague:
Currently the `--dry-run` parameter does not prevent Doctrine Migrations from creating the migration_versions table. This is especially awful when combining `--dry-run` with `--write-sql` to test the initial setup.

This is because the `\Doctrine\DBAL\Migrations\Tools\Console\Command\MigrateCommand` first loads the list of already executed migrations before it checks if a dry run should be performed.

I'd opt for altering the order of the two tasks.

stof:
well, a dry-run needs to check the list of executed migrations too, as having migrations which are already executed affects the output of the dry-run.

tehplague:
Agreed, but then the check should assume the list to be empty if migration_version is absent instead of silently creating that table. That means, the process of checking and creating migration_versions should be separated (currently both operations happen unconditionally inside the same function).

mikeSimonson:
@tehplague PR is welcome to improve that.

chrisguitarguy:
This was done in #588, maybe it can be closed? Or is it waiting on an official release?

alcaeus:
Looks like it hasn’t been released yet. @mikeSimonson what’s the status in a release, shall we do one?

PR #588, opened 2017-12-25
Do not create migration_versions when running with --dry-run

egircys:
The `migration_versions` table is being created when running `doctrine:migrations:migrate` with `--dry-run` option, as described by @tehplague in #521

Here's my proposal to prevent the table creation when running with `--dry-run`.


mikeSimonson:
@egircys Thanks

PR #590, opened 2017-12-29
Apply Doctrine CS 4.0, version composer.lock, merge stages

### LINKED ISSUE #596: Scrutinizer reports wrong coding standard patches
Scrutinizer is not configured for the Doctrine coding standards, and so removes spaces around `!` in its suggested patches.
The scrutinizer config should be updated with the right CS, or the CS checks should be disabled. Checking with the wrong config is worse than not checking.

Majkl578:
Changes to follow updated Doctrine CS. Most of them are:
- removed annotations
- updated phpDoc: especially array-related (please verify especially those around Migration/Version classes), removed useless types
- `"` -> `'`
- weak to strict comparison (hopefully not breaking anything)
- some other whitespace-related stuff etc.

Should be OK to go to 1.x branch - API stays untouched and strict types are not enabled. That would target 2.0.

~CS is not tagged yet, thus dev-master (WIP).~

Fixes #596.

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -1,10 +1,22 @@
 build:
-    environment:
-        php:
-            version: 7.1
-
-before_commands:
-    - "composer install --no-dev --prefer-source"
+    nodes:
+        analysis:
+            environment:
+                php:
+                    version: 7.1
+            cache:
+                disabled: false
+                directories:
+                    - ~/.composer/cache
+            project_setup:
+                override: true
+            tests:
+                override:
+                    - php-scrutinizer-run
+                    - phpcs-run
+    dependencies:
+        override:
+            - composer install --ignore-platform-reqs --no-interaction
 
 tools:
     external_code_coverage:
@@ -15,5 +27,3 @@ build_failure_conditions:
     - 'issues.label("coding-style").new.exists'                 # No new coding style issues allowed
     - 'issues.severity(>= MAJOR).new.exists'                    # New issues of major or higher severity
     - 'project.metric_change("scrutinizer.test_coverage", < 0)' # Code Coverage decreased from previous inspection
-    - 'patches.label("Doc Comments").new.exists'                # No new doc comments patches allowed
-    - 'patches.label("Unused Use Statements").new.exists'       # No new unused imports patches allowed
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Connection/Loader/ArrayConnectionConfigurationLoader.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Connection/Loader/ArrayConnectionConfigurationLoader.php
@@ -8,6 +8,7 @@ use Doctrine\DBAL\Migrations\Configuration\Connection\ConnectionLoaderInterface;
 
 class ArrayConnectionConfigurationLoader implements ConnectionLoaderInterface
 {
+    /** @var string */
     private $filename;
 
     public function __construct($filename)
@@ -18,6 +19,7 @@ class ArrayConnectionConfigurationLoader implements ConnectionLoaderInterface
     /**
      * read the input and return a Configuration, returns `false` if the config
      * is not supported
+     *
      * @return Connection|null
      */
     public function chosen()
@@ -26,12 +28,12 @@ class ArrayConnectionConfigurationLoader implements ConnectionLoaderInterface
             return null;
         }
 
-        if ( ! file_exists($this->filename)) {
+        if (! file_exists($this->filename)) {
             return null;
         }
 
         $params = include $this->filename;
-        if ( ! is_array($params)) {
+        if (! is_array($params)) {
             throw new \InvalidArgumentException('The connection file has to return an array with database configuration parameters.');
         }
 
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Connection/Loader/ConnectionHelperLoader.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Connection/Loader/ConnectionHelperLoader.php
@@ -9,9 +9,7 @@ use Symfony\Component\Console\Helper\HelperSet;
 
 class ConnectionHelperLoader implements ConnectionLoaderInterface
 {
-    /**
-     * @var string
-     */
+    /** @var string */
     private $helperName;
 
     /** @var  HelperSet */
@@ -19,11 +17,9 @@ class ConnectionHelperLoader implements ConnectionLoaderInterface
 
 
     /**
-     * ConnectionHelperLoader constructor.
-     * @param HelperSet $helperSet
      * @param string $helperName
      */
-    public function __construct(HelperSet $helperSet = null, $helperName)
+    public function __construct(?HelperSet $helperSet = null, $helperName)
     {
         $this->helperName = $helperName;
         if ($helperSet === null) {
@@ -35,6 +31,7 @@ class ConnectionHelperLoader implements ConnectionLoaderInterface
     /**
      * read the input and return a Configuration, returns `false` if the config
      * is not supported
+     *
      * @return Connection|null
      */
     public function chosen()
--- a/tests/Doctrine/DBAL/Migrations/Tests/Functional/FunctionalTest.php
+++ b/tests/Doctrine/DBAL/Migrations/Tests/Functional/FunctionalTest.php
@@ -431,8 +431,8 @@ class FunctionalTest extends MigrationTestCase
      */
     public function testMigrateWithConnectionWithAutoCommitOffStillPersistsChanges()
     {
-        $listener            = new AutoCommitListener();
-        list($conn, $config) = self::fileConnectionAndConfig();
+        $listener        = new AutoCommitListener();
+        [$conn, $config] = self::fileConnectionAndConfig();
         $config->registerMigration(1, MigrateWithDataModification::class);
         $migration = new Migration($config);
         $conn->getEventManager()->addEventSubscriber($listener);
--- a/tests/Doctrine/DBAL/Migrations/Tests/Functional/FunctionalTest.php
+++ b/tests/Doctrine/DBAL/Migrations/Tests/Functional/FunctionalTest.php
@@ -5,9 +5,10 @@ namespace Doctrine\DBAL\Migrations\Tests\Functional;
 use Doctrine\DBAL\Configuration as DbalConfig;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\DriverManager;
+use Doctrine\DBAL\Migrations\Configuration\Configuration;
+use Doctrine\DBAL\Migrations\Event\Listeners\AutoCommitListener;
 use Doctrine\DBAL\Migrations\Events;
 use Doctrine\DBAL\Migrations\Migration;
-use Doctrine\DBAL\Migrations\Event\Listeners\AutoCommitListener;
 use Doctrine\DBAL\Migrations\Provider\SchemaDiffProviderInterface;
 use Doctrine\DBAL\Migrations\Tests\MigrationTestCase;
 use Doctrine\DBAL\Migrations\Tests\Stub\EventVerificationListener;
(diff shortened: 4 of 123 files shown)


lib/Doctrine/DBAL/Migrations/Configuration/Connection/Loader/ArrayConnectionConfigurationLoader.php
```diff
@@ -8,6 +8,7 @@
 
 class ArrayConnectionConfigurationLoader implements ConnectionLoaderInterface
 {
+    /** @var string */
     private $filename;
 
     public function __construct($filename)
```
alcaeus: Too bad we can't require the typehint sniffs because they'd introduce BC breaks. I'll create an issue to make it optional so we can require parameter documentation without introducing BC breaks.

tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/DiffCommandTest.php
```diff
@@ -2,21 +2,33 @@
...
 use Doctrine\DBAL\Migrations\Tools\Console\Command\DiffCommand;
+use Doctrine\DBAL\Schema\Schema;
+use org\bovigo\vfs\vfsStream;
 
 class DiffCommandTest extends CommandTestCase
 {
-    const VERSION                       = '20160705000000';
-    const CUSTOM_RELATIVE_TEMPLATE_NAME = 'tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/_files/migration.tpl';
-    const CUSTOM_ABSOLUTE_TEMPLATE_NAME = __DIR__ . '/_files/migration.tpl';
+    public const VERSION                       = '20160705000000';
```
alcaeus: If the constants are not used outside of the class, I'd say to make them `private` or `protected`: there is no BC promise for test classes.

tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
```diff
@@ -2,28 +2,30 @@
...
+    }
+
     public function testPreviousVersionErrorsWhenThereIsNoPreviousVersion()
     {
         $this->willResolveVersionAlias('prev', null);
 
-        list($tester, $statusCode) = $this->executeCommand([
-            'version' => 'prev',
-        ]);
+        list($tester, $statusCode) = $this->executeCommand(['version' => 'prev']);
```
alcaeus: Not sure if there's a sniff for it, but what do you think of using short syntax for `list` assignments?
`[$tester, $statusCode] = $this->executeCommand(['version' => 'prev']);`

tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/UpToDateCommandTest.php
```diff
@@ -6,22 +6,24 @@
...
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
  * @covers \Doctrine\DBAL\Migrations\Tools\Console\Command\UpToDateCommand
  */
-class UpToDateCommandTest extends \PHPUnit\Framework\TestCase
+class UpToDateCommandTest extends TestCase
 {
-    /** @var \PHPUnit_Framework_MockObject_MockObject|OutputInterface */
+    /** @var OutputInterface|MockObject */
```
alcaeus: This is probably used in a bunch of other places as well, but it should probably be `OutputInterface&MockObject`.

tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
Majkl578: There is no sniff for this, but it'd be cool to have one. I changed these manually for now.

tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/UpToDateCommandTest.php
Majkl578: True, but I'm avoiding using `&` for now since it's not supported i.e. by PhpStormn at all. :( I hope it'll improve in future and PhpStorm will understand intersections.

tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/DiffCommandTest.php
Majkl578: Made all protected except `MigrateAddSqlPostAndPreUpAndDownTest::TABLE_NAME` which is used on multiple places.

Majkl578:
@alcaeus Adressed your comments. For now I did not apply Doctrine CS 2.1 (with the function/constant imports) since some tests rely on this behavior to mock realpath(), it should be handled separately.

@stof Tried to fix #596, but it seems Scrutinizer does not work [as documented](https://scrutinizer-ci.com/docs/tools/php/code-sniffer/) and does not pick the phpcs.xml.dist correctly - the patches in recent build are invalid. Do you have any suggestion?

(Or we can disable CS checks on Scrutinizer completely and check it only on Travis, leaving to someone else to decide.)

Majkl578:
Hmm, setting phpcs.xml.dist explicitly didn't help, patches are still invalid...

lib/Doctrine/DBAL/Migrations/Configuration/Connection/Loader/ArrayConnectionConfigurationLoader.php
alcaeus: Just cross-referencing https://github.com/slevomat/coding-standard/issues/242 here - the sniff needs to be split according to the maintainers there. Maybe if it hurts me too much I'll give that a shot.

tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
alcaeus: I'll see if I can come up with something - shouldn't be too hard 👍

tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Command/UpToDateCommandTest.php
alcaeus: Makes sense - let's hope for the best. PhpStan supports it, and there's an open issue in PhpStorm for it (although relatively new, so we might have to wait a while): https://youtrack.jetbrains.com/issue/WI-39419

alcaeus:
Agree on splitting out `use function` and `use const`, let's address this later. As for scrutinizer, I have no issue disabling scrutinizer checks if it doesn't want to use our coding standard. There's no point checking phpcs twice anyways (although we could remove the travis step if we get scrutinizer to play ball).

lcobucci:
@Majkl578 have you tried to upgrade to the "new" Scrutinizer analyser?

Majkl578:
@lcobucci Do you have any reference for that -i.e. what needs to be changed or an example config?

.scrutinizer.yml
```diff
@@ -2,6 +2,9 @@ build:
     environment:
         php:
             version: 7.1
+    tests:
+        override:
+            - phpcs-run --standard=phpcs.xml.dist
 
 before_commands:
     - "composer install --no-dev --prefer-source"
```
stof: should be removed IMO, as Scrutinizer handles installing with some caching (but should probably be done separately)

.scrutinizer.yml
```diff
@@ -2,6 +2,9 @@ build:
     environment:
         php:
             version: 7.1
+    tests:
+        override:
+            - phpcs-run --standard=phpcs.xml.dist
```
stof: I fear this might not be running the analysis anymore though

lib/Doctrine/DBAL/Migrations/Configuration/Connection/Loader/ArrayConnectionConfigurationLoader.php
stof: for the constructor, you could typehint the string. It does not create a signature mismatch. And passing something else than a string would explode on the `file_exists` call already.

lcobucci:
@Majkl578 try something like this:

```yml
build:
    nodes: # this enables the multi-node stuff (which might be helpful to run different PHP versions for example)
        analysis:
            environment:
                php:
                    version: 7.1

            project_setup:
                override: true

            tests:
                override:
                    - php-scrutinizer-run # this is the new analyser (which AFAIK generates the issues and patches)
                    - phpcs-run # this will run PHP_CodeSniffer using their wrapper (and should pick up our CS config)

tools:
    external_code_coverage:
        timeout: 600

build_failure_conditions:
    - 'elements.rating(<= C).new.exists'                        # No new classes/methods with a rating of C or worse allowed
    - 'issues.label("coding-style").new.exists'                 # No new coding style issues allowed
    - 'issues.severity(>= MAJOR).new.exists'                    # New issues of major or higher severity
    - 'project.metric_change("scrutinizer.test_coverage", < 0)' # Code Coverage decreased from previous inspection
    - 'patches.label("Doc Comments").new.exists'                # No new doc comments patches allowed
    - 'patches.label("Unused Use Statements").new.exists'       # No new unused imports patches allowed
```

Majkl578:
Adjusted with the snippet from @lcobucci + added cache (inspired by ORM config).
It seems to run CS fine now (note: by default, composer is executed with `--no-scripts --no-plugins` which breaks our CS).
The failure is unrelated to this PR.

Majkl578:
@alcaeus Updated this PR to include CS 4.0. Excluded fallback function references as per discussion above.

tests/Doctrine/DBAL/Migrations/Tests/Functional/FunctionalTest.php
```diff
@@ -430,8 +431,8 @@ public function testSkippedMigrationsDispatchesTheExpectedEvents()
      */
     public function testMigrateWithConnectionWithAutoCommitOffStillPersistsChanges()
     {
-        $listener            = new AutoCommitListener();
-        list($conn, $config) = self::fileConnectionAndConfig();
```
Majkl578: @carusogabriel We may want to add sniff for this.

carusogabriel: What is these? I didn’t get it

lib/Doctrine/DBAL/Migrations/Configuration/Connection/Loader/ConnectionHelperLoader.php
```diff
@@ -9,21 +9,17 @@
...
     private $helperSet;
 
 
     /**
-     * ConnectionHelperLoader constructor.
-     * @param HelperSet $helperSet
      * @param string $helperName
      */
-    public function __construct(HelperSet $helperSet = null, $helperName)
+    public function __construct(?HelperSet $helperSet = null, $helperName)
```
alcaeus: The signature needs updating: can't have an argument without default value after one with default value. I suggest removing the `null` default for `$helperSet`.

tests/Doctrine/DBAL/Migrations/Tests/Functional/FunctionalTest.php
alcaeus: Short list syntax: `list($foo, $bar) = $someArray;` can be written as `[$foo, $bar] = $someArray;`.

lib/Doctrine/DBAL/Migrations/Configuration/Connection/Loader/ConnectionHelperLoader.php
Majkl578: Actually, you can.

alcaeus: Yes and no: having a method signature like `foo($foo = 'foo', $bar)` will always require you to give all arguments, making the default value for `$foo` obsolete: https://3v4l.org/l1qrv

Since nobody could've been relying on the default value, we should drop it, especially since it was just there to allow for a nullable object type.

PR #592, opened 2018-01-09
Use $strict param in functions that have it

carusogabriel:
Some functions, e.g. `in_array`, have a third param to enable strict comparison :shield:


mikeSimonson:
@carusogabriel Thanks

PR #595, opened 2018-01-25
Fix output of the executed SQL query when using parameters

stof:
The executed SQL is printed prepended by a `->` in the output when the query does not have parameters, but it only has `-` when adding `with parameters` in the end of the message. This looks like a bug to me (I'm finding it weird each time I read the migration output). So here is the single-char fix.


alcaeus:
Agree, that does look weird. Thanks @stof!

PR #597, opened 2018-01-29
Added info about database host to …InfosHelper output

VBelozyorov:
I think it is quite important to see: on which host the migrations are going to apply


mikeSimonson:
Thanks @VBelozyorov, can you add a test about this ?

VBelozyorov:
I didn't be able to quickly find any tests which covers `Doctrine\DBAL\Migrations\Tools\Console\Helper\MigrationStatusInfosHelper` to simple add check of this line. Does those test exists?

mikeSimonson:
You can add a MigrationsStatusInfosHelperTest class.
Then test that the host, that is in the configuration, that you pass in the constructor is in the ouput of the method.

PR #599, opened 2018-01-31
Improve DB2 support

tobiasstadler:
If the migrations column name is a reserved keyword in the used database (e.g. DB2), consistently quote the migrations column name.


tobiasstadler:
Anyone willing to have a look at this?

PR #606, opened 2018-03-27
Fixed --write-sql input option default value

podorozhny:


stof:
This looks wrong to me:

- `getOption()` will now be `true` even when the option is not passed explicitly (as I said, using optional values in options is tricky, as it would actually require 2 different defaults)
- `true` is not a path

podorozhny:
What about something like that?

stof:
have you tried it for real for the different cases ?

podorozhny:
> have you tried it for real for the different cases ?

Yep, `migrations:migrate -n`, `migrations:migrate --write-sql`, `migrations:migrate --write-sql=test.sql` works as expected (first one executes migrations and the rest - writes it to current working dir).

jwage:
This will be fixed in 2.0 via #666

podorozhny:
iwkad22

jwage:
@podorozhny is there a problem?

jwage:
@podorozhny I am assume the passive aggressive picture is because you want to see a fix in 1.x? or you wanted this PR to be merged? Let me know. I am open to merging a fix in to 1.7.x release.

PR #609, opened 2018-04-03
Set output writer on injected configuration

lcobucci:
When injecting the configuration and connection via dependency injection the output writer was not being set, making the internal objects use an output writer that doesn't write to the STDOUT.

### DIFF
--- a/tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Helper/ConfigurationHelperTest.php
+++ b/tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Helper/ConfigurationHelperTest.php
@@ -73,13 +73,10 @@ class ConfigurationHelperTest extends MigrationTestCase
             $configurationHelper = new ConfigurationHelper($this->getSqliteConnection());
             $configfileLoaded    = $configurationHelper->getMigrationConfig($this->input, $this->getOutputWriter());
 
-            unlink($configFile);
-
             return trim($this->getOutputStreamContent($this->output));
-        } catch (\Exception $e) {
-            unlink($configFile);//i want to be really sure to cleanup this file
+        } finally {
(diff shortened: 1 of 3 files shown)


tests/Doctrine/DBAL/Migrations/Tests/Tools/Console/Helper/ConfigurationHelperTest.php
```diff
@@ -73,13 +73,16 @@ protected function getConfigurationHelperLoadsASpecificFormat($baseFile, $config
             $configurationHelper = new ConfigurationHelper($this->getSqliteConnection());
             $configfileLoaded    = $configurationHelper->getMigrationConfig($this->input, $this->getOutputWriter());
 
+            self::assertAttributeSame($this->getOutputWriter(), 'outputWriter', $configfileLoaded);
+
             unlink($configFile);
 
             return trim($this->getOutputStreamContent($this->output));
         } catch (\Exception $e) {
```
Ocramius: This should be a `finally`

lcobucci: I didn't change to `finally` because this needs to be backported and I wasn't sure about the version, now we saw that we're requiring `^7.1` for `v1.6.x` so it's all good.

mikeSimonson:
@lcobucci Thanks

lcobucci:
@mikeSimonson shouldn't we backport this to `v1.6.x`?

mikeSimonson:
Well there is no BC break so I don't really see the point of backporting. What do you think ?

lcobucci:
It doesn't break the BC indeed but it's a bug and it would be nice if we could have it released and since I don't know when do we plan to release v1.7 I thought it would be valuable to backport it

PR #610, opened 2018-04-03
Fix class constants as type in sql queries output

mikeSimonson:


tests/Doctrine/DBAL/Migrations/Tests/VersionTest.php
```diff
@@ -407,8 +443,8 @@ public function testDryRunCausesSqlToBeOutputViaTheOutputWriter()
         $ow       = new OutputWriter(function ($msg) use (&$messages) {
             $messages[] = trim($msg);
         });
-        $config  = new Configuration($this->getSqliteConnection(), $ow);
-        $version = new Version(
+        $config   = new Configuration($this->getSqliteConnection(), $ow);
```
lcobucci: Isn't this already fixed in `master`?

lcobucci: (handled by #608)

lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -461,7 +460,7 @@ private function formatParamsForOutput(array $params, array $types)
             if (Type::hasType($type)) {
```
lcobucci: I think that we can simplify some things, what do you think about this:

```php
private function formatParamsForOutput(array $params, array $types): string
{
    if (empty($params)) {
        return '';
    }

    $out = [];

    foreach ($params as $key => $value) {
        $formatted = $this->formatParameter($value, $types[$key] ?? 'string');

        $out[] = is_string($key) ? sprintf(':%s => %s', $key, $formatted) : $formatted;
    }

    return sprintf('with parameters (%s)', implode(', ', $out));
}

private function formatParameter($value, string $type) : string
{
    if (Type::hasType($type)) {
        return Type::getType($type)->convertToDatabaseValue(
            $value,
            $this->connection->getDatabasePlatform()
        );
    }

    // This is quite important since `$value` and `$type` are things provided
    // by the user, so we can't really ensure that `$value` will always be an array.
    if (is_array($value)) { 
        return implode(', ', array_map([$this, 'parameterToString'], $value));
    }

    return $this->parameterToString($value);
}

private function parameterToString($value) : string
{
    if (is_int($value) || is_string($value)) {
        return (string) $value;
    }

    if (is_bool($value)) {
        return $value === true ? 'true' : 'false';
    }

    return '?';
}
```

stof:
Can you show what the new output looks like ?

mikeSimonson:
@stof I did not run that on an actual migrations but it basically turns some
```
INSERT INTO test VALUES (?) with parameters (?)
```
into 
```
INSERT INTO test VALUES (?) with parameters (1, 2, 3)
```

tests/Doctrine/DBAL/Migrations/Tests/VersionTest.php
stof: it would be great to rebase the branch to clean the diff here

stof:
Looks fine to me

lcobucci:
@mikeSimonson I think that imploding the array might be confusing when you more parameters, like:

```sql
INSERT INTO test VALUES (?, ?) with parameters (1, 2, 3)
```

Which values are related to the first parameter and which are to the second? Maybe this helps:

```diff
- INSERT INTO test VALUES (?, ?) with parameters (1, 2, 3)
+ INSERT INTO test VALUES (?, ?) with parameters ([1, 2], 3)
```

lib/Doctrine/DBAL/Migrations/Version.php
mikeSimonson: I will take it

mikeSimonson:
I will try to add a test for this one too.

lib/Doctrine/DBAL/Migrations/Version.php
lcobucci: We could also make `parameterToString()` recursive, if needed... like:

```php
private function formatParameter($value, string $type) : string
{
    if (Type::hasType($type)) {
        return Type::getType($type)->convertToDatabaseValue(
            $value,
            $this->connection->getDatabasePlatform()
        );
    }

    return $this->parameterToString($value);
}

private function parameterToString($value) : string
{
    if (is_array($value)) { 
        return implode(', ', array_map([$this, 'parameterToString'], $value));
    }

    if (is_int($value) || is_string($value)) {
        return (string) $value;
    }

    if (is_bool($value)) {
        return $value === true ? 'true' : 'false';
    }

    return '?';
}
```

lib/Doctrine/DBAL/Migrations/Version.php
```diff
@@ -454,16 +453,17 @@ private function formatParamsForOutput(array $params, array $types)
...
-            }
-            $out[] = is_string($key) ? sprintf(':%s => %s', $key, $outval) : $outval;
+            $type   = $types[$key] ?? 'string';
+            $outval = $this->formatParameter($value, $type);
+            $out[]  = is_string($key) ? sprintf(':%s => %s', $key, $outval) : $outval;
+        }
+
+        if (count($out) > 1) {
+            $out = array_map(function ($element) {
+                return '[' . $element . ']';
```
lcobucci: Couldn't we solve this when generating the value (in line 458) instead of re-iterating `$out`? Like:

```diff
- $outval = $this->formatParameter($value, $type);
+ $outval = '[' . $this->formatParameter($value, $type) . ']';
```

Or only when dealing with arrays, in line 497:

```diff
- return implode(', ', array_map([$this, 'parameterToString'], $value));
+ return '[' . implode(', ', array_map([$this, 'parameterToString'], $value)) . ']';
```

The output would be a bit more understandable.

PR #638, opened 2018-05-06
Add PHPStan strict rules

Majkl578:


composer.json
```diff
@@ -24,7 +24,8 @@
         "jdorn/sql-formatter": "~1.1",
         "mikey179/vfsStream": "^1.6",
         "phpstan/phpstan": "^0.9.2",
-        "symfony/process": "^4.0"
+        "symfony/process": "^4.0",
+        "phpstan/phpstan-strict-rules": "^0.9"
```
localheinz: How about keeping packages sorted?

See https://getcomposer.org/doc/06-config.md#sort-packages or https://github.com/localheinz/composer-normalize.

lib/Doctrine/Migrations/Configuration/Configuration.php
```diff
@@ -556,7 +556,7 @@ public function getMigrationsToExecute(string $direction, string $to) : array
         $this->loadMigrationsFromDirectory();
 
         if ($direction === Version::DIRECTION_DOWN) {
-            if (count($this->migrations)) {
+            if (count($this->migrations) !== 0) {
```
lcobucci: I know that this might be a premature optimisation and that `doctrine/migrations` is not really sensitive regarding performance, but comparing with an empty array is way faster than using `count` - even with PHP 7.2 optimisations - so I'd suggest to use `if ($this->migrations !== []) {` that instead.

lcobucci: Simple bench:

```php
<?php

namespace Blah;

use function assert;
use function count;

/**
 * @BeforeMethods({"init"})
 */
final class ArrayBench
{
    private $data;

    public function init(): void
    {
        $this->data = array_fill(0, 10000000, 'banana');
    }

    /**
     * @Revs(1000)
     * @Iterations(5)
     */
    public function benchCount()
    {
        assert(count($this->data) !== 0);
    }

    /**
     * @Revs(1000)
     * @Iterations(5)
     */
    public function benchComparison()
    {
        assert($this->data !== []);
    }
}
```

Result:

```sh
PhpBench 0.14-dev (60dfc50). Running benchmarks.
Using configuration file: /Users/luis/projects/chimera/bus-tactician/blah/phpbench.json

\Blah\ArrayBench

    benchCount                    I4 P0 	[μ Mo]/r: 0.092 0.092 (μs) 	[μSD μRSD]/r: 0.001μs 0.69%
    benchComparison               I4 P0 	[μ Mo]/r: 0.089 0.089 (μs) 	[μSD μRSD]/r: 0.000μs 0.55%

2 subjects, 10 iterations, 2,000 revs, 0 rejects
(best [mean mode] worst) = 0.088 [0.090 0.090] 0.089 (μs)
⅀T: 0.903μs μSD/r 0.001μs μRSD/r: 0.620%
```

lcobucci: Obviously that the size of the array and the data it contains would also affect the benchmark.

lcobucci: @Majkl578 I'll merge it now so we can discuss this and eventually change (or not)

lcobucci:
@Majkl578 🚢

lib/Doctrine/Migrations/Configuration/Configuration.php
Majkl578: We had discussion about this few days ago on Slack.
The difference is so minimal that I decided to ignore it completely. Approach with count() is more future proof, if we eventually decide to introduce i.e. a) first class collections or b) php-ds.
The difference is not by factor of 100, not even by 100, in fact it's like 5%-10%. Using `!` is much safer but that takes us to non-strict land and we don't want to go that path,

See https://gist.github.com/Majkl578/8fd1dec6b9e005e53c247a4953ccebbf

PR #639, opened 2018-05-06
composer.json: sorted packages, HTTPS, drop old versions

Majkl578:


composer.json
```diff
@@ -3,7 +3,7 @@
     "type": "library",
     "description": "Database Schema migrations using Doctrine DBAL",
     "keywords": ["migrations", "database"],
-    "homepage": "http://www.doctrine-project.org",
+    "homepage": "https://www.doctrine-project.org",
```
alcaeus: Should we link to the project page for migrations instead of the root?

composer.json
```diff
@@ -13,19 +13,19 @@
     "require": {
         "php": "^7.1",
         "doctrine/dbal": "^2.6",
-        "symfony/console": "~3.3|^4.0",
-        "ocramius/proxy-manager": "^1.0|^2.0"
+        "ocramius/proxy-manager": "^1.0|^2.0",
+        "symfony/console": "~3.3|^4.0"
```
lcobucci: Does it still make sense to support these old versions when they are just ignored because of the PHP version requirement (might be unrelated to this PR, though)?

composer.json
```diff
@@ -13,19 +13,19 @@
...
     },
     "require-dev": {
-        "doctrine/orm": "~2.5",
-        "symfony/yaml": "~3.3|^4.0",
-        "phpunit/phpunit": "~7.0",
         "doctrine/coding-standard": "^4.0",
+        "doctrine/orm": "~2.5",
```
lcobucci: IMO we should bump this to `2.6` too

composer.json
```diff
@@ -13,19 +13,19 @@
...
         "jdorn/sql-formatter": "~1.1",
         "mikey179/vfsStream": "^1.6",
         "phpstan/phpstan": "^0.9.2",
+        "phpstan/phpstan-strict-rules": "^0.9",
+        "phpunit/phpunit": "~7.0",
         "symfony/process": "^4.0",
-        "phpstan/phpstan-strict-rules": "^0.9"
+        "symfony/yaml": "~3.3|^4.0"
```
lcobucci: Same comment here

composer.json
Majkl578: yes, thought so too, didn't want to do here :smile:

composer.json
Majkl578: yes

composer.json
Majkl578: yes

composer.json
Majkl578: Not sure about that, haven't seen such pattern anywhere. :/

Majkl578:
Bumped ProxyManager to ^2.0.2 due to a bug: https://travis-ci.org/doctrine/migrations/jobs/375639450#L609

emodric:
Would it be possible to support Symfony 3.4, seeing how it is an LTS version?

jwage:
@emodric yes, can you create an issue for that?

emodric:
@jwage Thanks! Done in #651

composer.json
localheinz: How about adding a `support` section, then?

For reference, see https://getcomposer.org/doc/04-schema.md#support.

stof: Well, once we have a proper page for the product, it would make sense to link to it as the homepage.

jwage: It should link here https://www.doctrine-project.org/projects/migrations.html

PR #641, opened 2018-05-06
Extract parameter formatting from Version class and move to Parameter…

### LINKED ISSUE #640: Extract parameter formatting out of Version class

jwage:
Fixes #640

### DIFF
--- a/lib/Doctrine/Migrations/ParameterFormatter.php
+++ b/lib/Doctrine/Migrations/ParameterFormatter.php
@@ -1,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Migrations;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Types\Type;
+use function array_map;
+use function implode;
+use function is_array;
+use function is_bool;
+use function is_int;
+use function is_string;
+use function sprintf;
+
+final class ParameterFormatter
+{
+    /** @var Connection */
+    private $connection;
+
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+    }
+
+    /**
+     * @param mixed[] $params
+     * @param mixed[] $types
+     */
+    public function formatParameters(array $params, array $types) : string
+    {
+        if ($params === []) {
+            return '';
+        }
+
+        $formattedParameters = [];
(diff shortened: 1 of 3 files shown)


lib/Doctrine/Migrations/ParameterFormatter.php
```diff
@@ -0,0 +1,84 @@
...
+        $this->connection = $connection;
+    }
+
+    /**
+     * @param mixed[] $params
+     * @param mixed[] $types
+     */
+    public function formatParameters(array $params, array $types) : string
+    {
+        if (empty($params)) {
```
Majkl578: Please avoid `empty()`, either use `count(...) !== 0` or `!== []`

lib/Doctrine/Migrations/ParameterFormatter.php
```diff
@@ -0,0 +1,84 @@
...
+            return '';
+        }
+
+        $out = [];
+        foreach ($params as $key => $value) {
+            $type   = $types[$key] ?? 'string';
+            $outval = '[' . $this->formatParameter($value, $type) . ']';
```
Majkl578: `$formattedValue`

lib/Doctrine/Migrations/ParameterFormatter.php
```diff
@@ -0,0 +1,84 @@
...
+        return $this->parameterToString($value);
+    }
+
+    /**
+     * @param int[]|bool[]|string[]|array|int|string|bool $value
+     */
+    private function parameterToString($value) : string
+    {
+        if (is_array($value)) {
+            return implode(', ', array_map([$this, 'parameterToString'], $value));
```
Majkl578: Better use closure here instead of string name.

lib/Doctrine/Migrations/ParameterFormatter.php
```diff
@@ -0,0 +1,84 @@
...
+                $value,
+                $this->connection->getDatabasePlatform()
+            );
+        }
+
+        return $this->parameterToString($value);
+    }
+
+    /**
+     * @param int[]|bool[]|string[]|array|int|string|bool $value
```
Majkl578: the function is doing recursion 🤔

PR #646, opened 2018-05-07
Refactor internal classes

### LINKED ISSUE #629: Move event dispatching out of Configuration class
https://github.com/doctrine/migrations/pull/617#discussion_r186265183

### LINKED ISSUE #645: Refactor execution out of the Version class
Refactor the execution out of the Version class. Create a VersionExecutor class.

jwage:
Fixes #645 
Fixes #629

I moved stuff around and refactored but maintained BC in the public API. There is more to do in general but to avoid a monster PR I think this is all I want to do for now.

### DIFF
--- a/lib/Doctrine/Migrations/EventDispatcher.php
+++ b/lib/Doctrine/Migrations/EventDispatcher.php
@@ -1,26 +1,30 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Doctrine\Migrations;
 
 use Doctrine\Common\EventArgs;
-use Doctrine\DBAL\Connection;
+use Doctrine\Common\EventManager;
 use Doctrine\Migrations\Configuration\Configuration;
 use Doctrine\Migrations\Event\MigrationsEventArgs;
 use Doctrine\Migrations\Event\MigrationsVersionEventArgs;
-use Doctrine\Migrations\Version;
 
+/**
+ * @internal
+ */
 final class EventDispatcher
 {
     /** @var Configuration */
     private $configuration;
 
-    /** @var Connection */
-    private $connection;
+    /** @var EventManager */
+    private $eventManager;
 
-    public function __construct(Configuration $configuration, Connection $connection)
+    public function __construct(Configuration $configuration, EventManager $eventManager)
     {
         $this->configuration = $configuration;
-        $this->connection = $connection;
+        $this->eventManager  = $eventManager;
     }
 
     public function dispatchMigrationEvent(string $eventName, string $direction, bool $dryRun) : void
@@ -47,7 +51,7 @@ final class EventDispatcher
 
     public function dispatchEvent(string $eventName, ?EventArgs $args = null) : void
     {
-        $this->connection->getEventManager()->dispatchEvent($eventName, $args);
+        $this->eventManager->dispatchEvent($eventName, $args);
     }
 
     private function createMigrationEventArgs(string $direction, bool $dryRun) : MigrationsEventArgs
--- a/lib/Doctrine/Migrations/EventDispatcher.php
+++ b/lib/Doctrine/Migrations/EventDispatcher.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace Doctrine\Migrations;
+
+use Doctrine\Common\EventArgs;
+use Doctrine\DBAL\Connection;
+use Doctrine\Migrations\Configuration\Configuration;
+use Doctrine\Migrations\Event\MigrationsEventArgs;
+use Doctrine\Migrations\Event\MigrationsVersionEventArgs;
+use Doctrine\Migrations\Version;
+
+final class EventDispatcher
+{
+    /** @var Configuration */
+    private $configuration;
+
+    /** @var Connection */
+    private $connection;
+
+    public function __construct(Configuration $configuration, Connection $connection)
+    {
+        $this->configuration = $configuration;
+        $this->connection = $connection;
+    }
+
+    public function dispatchMigrationEvent(string $eventName, string $direction, bool $dryRun) : void
+    {
+        $event = $this->createMigrationEventArgs($direction, $dryRun);
+
+        $this->dispatchEvent($eventName, $event);
+    }
+
+    public function dispatchVersionEvent(
+        Version $version,
+        string $eventName,
+        string $direction,
+        bool $dryRun
+    ) : void {
+        $event = $this->createMigrationsVersionEventArgs(
+            $version,
+            $direction,
+            $dryRun
+        );
+
+        $this->dispatchEvent($eventName, $event);
+    }
+
+    public function dispatchEvent(string $eventName, ?EventArgs $args = null) : void
+    {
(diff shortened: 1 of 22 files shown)


lib/Doctrine/Migrations/EventDispatcher.php
```diff
@@ -0,0 +1,71 @@
...
+
+namespace Doctrine\Migrations;
+
+use Doctrine\Common\EventArgs;
+use Doctrine\DBAL\Connection;
+use Doctrine\Migrations\Configuration\Configuration;
+use Doctrine\Migrations\Event\MigrationsEventArgs;
+use Doctrine\Migrations\Event\MigrationsVersionEventArgs;
+
+final class EventDispatcher
```
Majkl578: I am personally not a big fan of introducing _yet another event dispatcher_ in Doctrine, given that we are looking for replacement of the one in Common for ORM 3.0.

I suppose this would be good as temporary solution until the replacement is found, but then it should be abstracted through interface + marked as interal.

lib/Doctrine/Migrations/EventDispatcher.php
```diff
@@ -0,0 +1,71 @@
...
+            $direction,
+            $dryRun
+        );
+
+        $this->dispatchEvent($eventName, $event);
+    }
+
+    public function dispatchEvent(string $eventName, ?EventArgs $args = null) : void
+    {
+        $this->connection->getEventManager()->dispatchEvent($eventName, $args);
```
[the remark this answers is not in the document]
Majkl578: I would not reuse EvM from Connection at all, it's entirely different layer.

lib/Doctrine/Migrations/Version.php
```diff
@@ -96,7 +81,17 @@ public function __construct(
 
         $this->schemaProvider = $schemaProvider;
 
-        $this->parameterFormatter = new ParameterFormatter($this->connection);
+        $this->versionExecutor = new VersionExecutor(
```
Majkl578: This isn't DI friendly.

lib/Doctrine/Migrations/Version.php
```diff
@@ -109,19 +104,46 @@ public function getConfiguration() : Configuration
...
     public function isMigrated() : bool
     {
         return $this->configuration->hasVersionMigrated($this);
     }
 
-    public function markMigrated() : void
+    public function setState(int $state) : void
     {
-        $this->markVersion(self::DIRECTION_UP);
+        $this->state = $state;
```
Majkl578: This needs validation of `$state`, I think simple `assert()` would do.

lib/Doctrine/Migrations/Version.php
```diff
@@ -174,9 +179,18 @@ public function writeSqlFile(
             ->write($path, $direction, $sqlQueries);
     }
 
-    public function getMigration() : AbstractMigration
-    {
-        return $this->migration;
+    public function exec(
```
Majkl578: No abbreviations :sob:, anything wrong with `execute()`?

lib/Doctrine/Migrations/Version.php
```diff
@@ -109,19 +104,46 @@ public function getConfiguration() : Configuration
...
+                return 'No State';
+        }
+    }
+
+    public function getTime() : ?float
+    {
+        return $this->time;
+    }
+
+    public function setTime(float $time) : void
```
Majkl578: Instead of doing this, could it be moved to the Result? Or possibly passed around through some ExecutionContext?

lib/Doctrine/Migrations/VersionExecutor.php
```diff
@@ -0,0 +1,323 @@
...
+        $this->configuration  = $configuration;
+        $this->connection     = $connection;
+        $this->schemaProvider = $schemaProvider;
+        $this->outputWriter   = $outputWriter;
+
+        $this->parameterFormatter = new ParameterFormatter($this->connection);
+    }
+
+    /**
+     * @param string[]|string $sql
```
Majkl578: I'd try to eliminate this by always passing an array of strings (or a collection). This leads to some nasty if-else tricks.

lib/Doctrine/Migrations/EventDispatcher.php
jwage: @Majkl578 this is not necessarily a new "EventDispatcher", it is a composition class that depends on the existing `Doctrine\Common\EventManager`. I am fine with marking it internal. I don't intend for anyone else to be using this outside of migrations internals.

jwage: This is really just an abstraction on top of EventManager and not necessarily a new EventDispatcher. I will mark it as internal.

lib/Doctrine/Migrations/Version.php
jwage: Yes, I think we can refactor this further to get rid of all the state in the Version class.

lib/Doctrine/Migrations/Version.php
jwage: I maintained backwards compatability. The `execute()` method already exists and I didn't want to change the signature of that method.

lib/Doctrine/Migrations/VersionExecutor.php
jwage: I can make this change here, I won't be able to make the change in the addSql() that the end user uses as it would be a big BC break, but I can make it so the internal addSql here always receives a `string[]`

stof:
@Majkl578 the advantage of keeping the EventManager of the Connection is that users have a single place to register their listeners, instead of having to gain access to the specific EventManager used by the doctrine/migrations layer (which would make it harder to use doctrine/migrations only as a tool and not as a library).

Majkl578:
That is currently true for DBAL 2.x, but since Doctrine's EvM has uncertain fate, I would be more careful about making it a hard contract, given that DBAL 3 and ORM 3 may ship something different (I assume Migrations 2.0 will eventually support DBAL 3).

stof:
Well, would doctrine/migrations wait for the DBAL 3 release to be released, or do you plan to release it sooner ? If you want to be able to support DBAL 3, it might be quite hard to release migrations 2.x sooner, as it might require other API changes to support it. And if we wait, we can keep this implementation until the future of EvM gets settled.

stof:
Then I vote for keeping the usage of the EvM of the connection for now, to make the upgrade to doctrine/migrations easy (listeners would work the same than in 1.x for users), and making a new major version of doctrine/migrations for DBAL 3.0 if needed, based on what the DBAL 3 support involves

jwage:
@stof Agree. This is my current plan.

lib/Doctrine/Migrations/VersionExecutionResult.php
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    public function isSkipped() : bool
+    {
+        return $this->skipped;
+    }
+
+    public function setError(bool $error) : bool
+    {
+        $this->error = $error;
```
Majkl578: As scrutinizer points out, this will crash due to void return in non-void method.

Issue #648, opened 2018-05-07
 [1.7.0] Return value of Doctrine\DBAL\Migrations\Version::formatParameter() must be of the type string, null returned


MatthiasKuehneEllerhold:
Upgrading to 1.7.0 leads to this error when trying to insert a NULL-Value into a database table row.

```
TypeError: Return value of Doctrine\DBAL\Migrations\Version::formatParameter() must be of the type string, null returned in /x/vendor/doctrine/migrations/lib/Doctrine/DBAL/Migrations/Version.php:481
Stack trace:
#0 /x/vendor/doctrine/migrations/lib/Doctrine/DBAL/Migrations/Version.php(459): Doctrine\DBAL\Migrations\Version->formatParameter(NULL, 'string')
#1 /x/vendor/doctrine/migrations/lib/Doctrine/DBAL/Migrations/Version.php(433): Doctrine\DBAL\Migrations\Version->formatParamsForOutput(Array, Array)
#2 /x/vendor/doctrine/migrations/lib/Doctrine/DBAL/Migrations/Version.php(400): Doctrine\DBAL\Migrations\Version->outputSqlQuery(133, 'INSERT INTO `...')
#3 /x/vendor/doctrine/migrations/lib/Doctrine/DBAL/Migrations/Version.php(282): Doctrine\DBAL\Migrations\Version->executeRegisteredSql(false, false)
#4 /x/vendor/doctrine/migrations/lib/Doctrine/DBAL/Migrations/Migration.php(172): Doctrine\DBAL\Migrations\Version->execute('up', false, false)
#5 /x/vendor/doctrine/migrations/lib/Doctrine/DBAL/Migrations/Tools/Console/Command/MigrateCommand.php(132): Doctrine\DBAL\Migrations\Migration->migrate('065', false, false, Object(Closure))
#6 /x/vendor/symfony/console/Command/Command.php(252): Doctrine\DBAL\Migrations\Tools\Console\Command\MigrateCommand->execute(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#7 /x/vendor/symfony/console/Application.php(865): Symfony\Component\Console\Command\Command->run(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
[...]

PR #649, opened 2018-05-07
Fix the possibility to insert / update NULL values, add Unit-Test

MatthiasKuehneEllerhold:
Fix for https://github.com/doctrine/migrations/issues/648


jwage:
Thanks for reporting. I will release 1.7.1 today with this fix.

PR #652, opened 2018-05-07
Enhancement: Link to project documentation

localheinz:
This PR

* [x] links to the  project documentation in `composer.json`

Follows https://github.com/doctrine/migrations/pull/639#discussion_r186436705.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -3,7 +3,7 @@
     "type": "library",
     "description": "Database Schema migrations using Doctrine DBAL",
     "keywords": ["migrations", "database"],
-    "homepage": "https://www.doctrine-project.org",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -51,5 +51,10 @@
     },
     "bin": [
         "bin/doctrine-migrations"
-    ]
+    ],
+    "support": {
+        "issues": "https://github.com/doctrine/migrations/issues",
+        "source": "https://github.com/doctrine/migrations",
```
stof: these 2 are not necessary, as they are already auto-detected by Composer AFAIK.

jwage: If what @stof says above is true, then :+1:

localheinz:
Thank you, @jwage and @stof!

PR #653, opened 2018-05-07
Support Symfony 3.4 in Doctrine Migration 2.0 as 3.4 is a LTS release.

### LINKED ISSUE #651: Bring back support for Symfony 3.4
https://github.com/doctrine/migrations/pull/639 bumped old dependencies to newer versions, and in the process, Symfony components dependencies were bumped to `^4.0`.

However, since currently, only LTS version of Symfony is 3.4, and will be for the next year and a half, it would be nice for version 2.0 of Migrations to support it.

jwage:
Fixes #651

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
         "php": "^7.1",
         "doctrine/dbal": "^2.6",
         "ocramius/proxy-manager": "^2.0.2",
-        "symfony/console": "^4.0"
+        "symfony/console": "^3.4||^4.0"
     },
     "require-dev": {
         "doctrine/coding-standard": "^4.0",
--- a/composer.lock
+++ b/composer.lock
@@ -1,10 +1,10 @@
 {
     "_readme": [
         "This file locks the dependencies of your project to a known state",
-        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
+        "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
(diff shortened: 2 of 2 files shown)


emodric:
@jwage `symfony/console` dependency should also be updated :)

jwage:
@Majkl578 @emodric updated.

composer.json
```diff
@@ -14,7 +14,7 @@
         "php": "^7.1",
         "doctrine/dbal": "^2.6",
         "ocramius/proxy-manager": "^2.0.2",
-        "symfony/console": "^4.0"
+        "symfony/console": "^3.4|^4.0"
```
stof: I suggest using `||` as the OR operator rather than the older `|` to follow the composer recommendation (choosing the double pipe as the main operator was done because it is also the operator used by ruby's bundler and by npm, making the syntax consistent between package managers)

Majkl578: Doctrine uses single pipe everywhere currently. Could you point to the _recommendation_ and ideally to _deprecation_ of current format?

jwage: Updated.

emodric:
:tada:

composer.json
stof: deprecation is not planned (it would break all existing releases using it, so it is almost impossible to deprecate anything there). But the older operators have been removed from the composer doc in favor of new ones (this was done when updating the doc in https://github.com/composer/composer/pull/2596/commits/ebab9db27ac3887babc8404107dd448d974b2f43)

PR #660, opened 2018-05-08
Refactor monster Configuration class.

### LINKED ISSUE #544: Review responsibilities of the Configuration object
We have so many reasons to change that object that would be wise to split things.

jwage:
Fixes #544

- [x] Extract functionality out of Configuration in to separate classes and proxy to the new classes initially.
- [x] Extract MigrationRepository::getMigrationsToExecute method to a MigrationPlanCalculator class.
- [x] Move resolveVersionAlias to VersionAliasResolver class.
- [x] We need some kind of dependency/object factory. Right now all our objects and dependencies are in Configuration class and it is being used as a sort of service locator.

### DIFF
--- a/lib/Doctrine/Migrations/MigrationRepository.php
+++ b/lib/Doctrine/Migrations/MigrationRepository.php
@@ -1,0 +1,331 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Migrations;
+
+use Doctrine\DBAL\Connection;
+use Doctrine\Migrations\Configuration\Configuration;
+use Doctrine\Migrations\Exception\DuplicateMigrationVersion;
+use Doctrine\Migrations\Exception\MigrationClassNotFound;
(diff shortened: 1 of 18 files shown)


lib/Doctrine/Migrations/MigrationRepository.php
```diff
@@ -0,0 +1,339 @@
...
+        $versions = array_keys($this->migrations);
+        $latest   = end($versions);
+
+        return $latest !== false ? (string) $latest : '0';
+    }
+
+    /** @return Version[] */
+    public function getMigrationsToExecute(string $direction, string $to) : array
+    {
+        $migrationPlanCalculator = new MigrationPlanCalculator($this);
```
jwage: Whoops. This needs to move to `DependencyFactory`.

jwage:
If anyone has any feedback, leave comments on this PR and I will open up new PRs to address the issues.

PR #663, opened 2018-05-10
Fix use of tho to be to in docs

MasterOdin:
Fixes misspelling in the docs

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no


mikeSimonson:
@MasterOdin Thanks

PR #666, opened 2018-05-11
Fix --write-sql option so it can accept a path to write the sql file to.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #606

#### Summary

Allows the `--write-sql` to accept a value to write the sql file to. If you don't specify a value, it writes the file to the current working directory.

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
@@ -34,7 +34,8 @@ class MigrateCommand extends AbstractCommand
                 'write-sql',
                 null,
                 InputOption::VALUE_OPTIONAL,
-                'The path to output the migration SQL file instead of executing it. Default to current working directory.'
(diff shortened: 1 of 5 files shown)


lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
```diff
@@ -34,7 +34,8 @@ protected function configure() : void
                 'write-sql',
                 null,
                 InputOption::VALUE_OPTIONAL,
-                'The path to output the migration SQL file instead of executing it. Default to current working directory.'
+                'The path to output the migration SQL file instead of executing it. Default to current working directory.',
```
localheinz: Should this be `Defaults to` or `Default is` instead?

jwage: Updated to `Defaults`.

Issue #628, opened 2018-05-05
Extract constants to constant only classes


jwage:
https://github.com/doctrine/migrations/pull/617#discussion_r186265123

Majkl578:
Should follow DBAL pattern.

PR #672, opened 2018-05-12
Move constants to constant only classes.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | change
| BC Break     | yes
| Fixed issues | #628

#### Summary

Implement constant only classes to match the other Doctrine projects.


lib/Doctrine/Migrations/VersionDirection.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Migrations;
+
+final class VersionDirection
+{
+    public const DIRECTION_UP   = 'up';
+    public const DIRECTION_DOWN = 'down';
```
Majkl578: `VersionDirection::UP` + `VersionDirection::DOWN` is imho the way we want to go.

lib/Doctrine/Migrations/VersionState.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Migrations;
+
+final class VersionState
+{
+    public const STATE_NONE = 0;
```
Majkl578: dtto here, without `STATE_`

PR #673, opened 2018-05-12
Show more help text after generating a migration.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #173

#### Summary

Improve the developer experience by showing more help text after generating a new migration.

cc @weaverryan 

#### Output

```
$ ./migrations migrations:generate
Generated new migration class to "/data/doctrine/migration-test/lib/Doctrine/Test/Migrations/Version20180512034732.php"

To run just this migration for testing purposes, you can use migrations:execute --up 20180512034732

To revert the migration you can use migrations:execute --down 20180512034732
```

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/GenerateCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/GenerateCommand.php
@@ -52,7 +52,19 @@ EOT
             $this->procOpen($editorCommand, $path);
         }
 
-        $output->writeln(sprintf('Generated new migration class to "<info>%s</info>"', $path));
+        $output->writeln([
+            sprintf('Generated new migration class to "<info>%s</info>"', $path),
+            '',
+            sprintf(
+                'To run just this migration for testing purposes, you can use <info>migrations:execute --up %s</info>',
+                $versionNumber
+            ),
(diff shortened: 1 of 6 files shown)


lib/Doctrine/Migrations/Tools/Console/Command/GenerateCommand.php
```diff
@@ -53,6 +53,16 @@ public function execute(InputInterface $input, OutputInterface $output) : void
         }
 
         $output->writeln(sprintf('Generated new migration class to "<info>%s</info>"', $path));
+
+        $output->writeln(sprintf(
+            "\nTo run just this migration for testing purposes, you can use <info>migrations:execute --up %s</info>",
+            $versionNumber
+        ));
+
+        $output->writeln(sprintf(
```
localheinz: Instead of multiple consecutive calls to `writeln()`, and instead of prefixing messages with `"\n"`, what do you think about passing an array of message strings (and empty strings for new lines) to `writeln()` instead?

jwage: Changed.

Issue #331, opened 2015-08-05
Migrations should log when particular migration was executed


bostjan:
This is just a general suggestion, fishing for comments:)

Executed migrations are logged in database, in specified table. But nothing is logged about WHEN particular migration was executed.

I think this would be small but convenient addition to otherwise excellent software.

Please comment.

b.

mikeSimonson:
@bostjan Well I agree but I would like the advice of others. @stof ?

stof:
well, the hard part here is that currently, we just detect whether the migration table exists and we create it when it is not. But we would also have to check whether it is the uptodate one or the old one and create the necessary fields if it is old. We need to find an efficient way to do this as it will run at the beginning of each command.

Otherwise, this seems like a convenient feature.

bostjan:
@mikeSimonson @stof : So, if I am picking up the vibe correctly, it boils down to this question: if I provide a pull request with implementation, will it not go to waste, provided that new code does not totally suck? :)

bostjan:
Another question: datetime field would be preferable?
And another: using UTC (gmdate) would be preferable, but this issue is not yet resolved, right? https://github.com/doctrine/migrations/issues/330

b.

mikeSimonson:
It won't go to waste, but I am wondering if an approach were:
If the datetime field is not there it's not used and if it's there it's used ?

What do you think @stof ?

stof:
@mikeSimonson the issue is that this means that existing projects will never benefit from this feature. And it makes the read query harder as we would not be able to select fields explicitly

bostjan:
@stof you mean those existing projects that do not upgrade doctrine+migrations?

stof:
@bostjan if the suggestion of @mikeSimonson is applied, updating the library is not enough, if the migration table already exists:

> If the datetime field is not there it's not used and if it's there it's used

jwage:
Couldn't we offer a convenient way to migrate the table and add the new column for the datetime?

PR #675, opened 2018-05-12
Record when a migration was executed.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #331

#### Summary

Log when a migration was executed.


lib/Doctrine/Migrations/Configuration/AbstractFileConfiguration.php
```diff
@@ -23,6 +23,7 @@
         'migrations_namespace',
         'table_name',
         'column_name',
+        'created_at_column_name',
```
localheinz: Maybe `executed_at_column_name`, and consequently, a rename of the corresponding methods, would be more appropriate?

After all, this reflects the time of when the migration was executed, rather then when the migration (file) was created.

What do you think?

lib/Doctrine/Migrations/MigrationTableManager.php
```diff
@@ -0,0 +1,224 @@
...
+        $table->setPrimaryKey([$this->columnName]);
+
+        // remove columns from the table definition that we don't care about
+        // so we don't try to drop those columns
+        foreach ($table->getColumns() as $column) {
+            if (in_array($column->getName(), $columnNames, true)) {
+                continue;
+            }
+
+            $table->dropColumn($column->getName());
```
localheinz: If I understood correctly, then on the happy path we would not need to drop any columns from the definition - what do you think about modifying to

```diff
diff --git a/lib/Doctrine/Migrations/MigrationTableManager.php b/lib/Doctrine/Migrations/MigrationTableManager.php
index 175c2da..721116c 100644
--- a/lib/Doctrine/Migrations/MigrationTableManager.php
+++ b/lib/Doctrine/Migrations/MigrationTableManager.php
@@ -189,11 +189,9 @@ class MigrationTableManager
         // remove columns from the table definition that we don't care about
         // so we don't try to drop those columns
         foreach ($table->getColumns() as $column) {
-            if (in_array($column->getName(), $columnNames, true)) {
-                continue;
+            if (!in_array($column->getName(), $columnNames, true)) {
+                $table->dropColumn($column->getName());
             }
-
-            $table->dropColumn($column->getName());
         }

         return $table;
```

?

Majkl578: That goes against early-exit rule in conditions/loops in Doctrine CS. :)

lib/Doctrine/Migrations/Configuration/Configuration.php
```diff
@@ -42,6 +42,9 @@ class Configuration
     /** @var int */
     private $migrationsColumnLength = 255;
 
+    /** @var string */
+    private $migrationsExecutedAtColumnName = 'executedAt';
```
Majkl578: Should probably be `executed_at` to match table name and also avoid case sensitivity mess.

lib/Doctrine/Migrations/DependencyFactory.php
```diff
@@ -79,12 +81,18 @@ public function getMigrationRepository() : MigrationRepository
         });
     }
 
-    public function getMigrationTableCreator() : MigrationTableCreator
+    public function getMigrationTableManager() : MigrationTableManager
     {
-        return $this->getDependency(MigrationTableCreator::class, function () {
-            return new MigrationTableCreator(
+        return $this->getDependency(MigrationTableManager::class, function () {
```
Majkl578: Closure needs a return type.

lib/Doctrine/Migrations/MigrationFileBuilder.php
```diff
@@ -53,15 +64,24 @@ private function getVersionUpdateQuery(string $version, string $direction) : str
     {
         if ($direction === VersionDirection::DOWN) {
             $query = "DELETE FROM %s WHERE %s = '%s';\n";
```
Majkl578: Unnecessary variable.

lib/Doctrine/Migrations/MigrationFileBuilder.php
```diff
@@ -53,15 +64,24 @@ private function getVersionUpdateQuery(string $version, string $direction) : str
...
+
+            return sprintf(
+                $query,
+                $this->tableName,
+                $this->columnName,
+                $version
+            );
         }
 
+        $query = "INSERT INTO %s (%s, %s) VALUES ('%s', %s);\n";
```
Majkl578: Unnecessary variable.

lib/Doctrine/Migrations/MigrationTableManager.php
```diff
@@ -0,0 +1,224 @@
...
+use Doctrine\DBAL\Schema\Table;
+use Doctrine\DBAL\Types\Type;
+use Doctrine\Migrations\Configuration\Configuration;
+use function array_keys;
+use function in_array;
+
+/**
+ * @internal
+ */
+class MigrationTableManager
```
Majkl578: Any better name? :worried: 
https://stackoverflow.com/questions/1866794/naming-classes-how-to-avoid-calling-everything-a-whatevermanager

lib/Doctrine/Migrations/MigrationTableManager.php
```diff
@@ -0,0 +1,224 @@
...
+                $this->migrationTableCreated = false;
+            }
+        }
+
+        return $this->migrationTableCreated;
+    }
+
+    public function isMigrationTableUpToDate() : bool
+    {
+        if ($this->migrationTableUpToDate === null) {
```
Majkl578: Inverting the condition will decrease the depth of this method.

lib/Doctrine/Migrations/MigrationTableManager.php
```diff
@@ -0,0 +1,224 @@
...
+    public function createMigrationTable() : bool
+    {
+        $this->configuration->validate();
+
+        if ($this->configuration->isDryRun()) {
+            return false;
+        }
+
+        if ($this->isMigrationTableCreated()) {
+            if ($this->isMigrationTableUpToDate() === false) {
```
Majkl578: Strict comparison seems unnecessary for boolean-only method.

lib/Doctrine/Migrations/Version.php
```diff
@@ -194,4 +199,17 @@ public function markVersion(string $direction) : void
             );
         }
     }
+
+    private function getExecutedAt() : DateTime
```
Majkl578: DateTimeImmutable please

lib/Doctrine/Migrations/Version.php
```diff
@@ -194,4 +199,17 @@ public function markVersion(string $direction) : void
...
+    {
+        return new DateTime();
+    }
+
+    private function getExecutedAtDatabaseValue() : string
+    {
+        return Type::getType('datetime')->convertToDatabaseValue(
```
Majkl578: datetime_immutable please

lib/Doctrine/Migrations/MigrationTableManager.php
```diff
@@ -0,0 +1,224 @@
...
+            Type::getType(self::MIGRATION_COLUMN_TYPE),
+            ['length' => $this->columnLength]
+        );
+    }
+
+    public function getExecutedAtColumn() : Column
+    {
+        return new Column(
+            $this->executedAtColumnName,
+            Type::getType('datetime'),
```
Majkl578: datetime_immutable please

lib/Doctrine/Migrations/MigrationTableManager.php
```diff
@@ -0,0 +1,224 @@
...
+    private $schemaManager;
+
+    /** @var Connection */
+    private $connection;
+
+    /** @var AbstractPlatform */
+    private $platform;
+
+    /** @var string */
+    private $tableName;
```
lcobucci: I think it might be time to have a class to represent the migration table, which encapsulates: name, column name (needs a better name), column length (better name as well?), and executed at column name.

jwage: Good idea. I will make that change.

lib/Doctrine/Migrations/MigrationTableManager.php
```diff
@@ -0,0 +1,224 @@
...
+
+                    break;
+                }
+            }
+        }
+
+        return $this->migrationTableUpToDate;
+    }
+
+    public function createMigrationTable() : bool
```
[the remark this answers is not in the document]
jwage: Possibly. I will make an issue to revisit this and maybe implement exceptions for these cases instead of returning bool.

lib/Doctrine/Migrations/Configuration/Configuration.php
```diff
@@ -140,6 +143,24 @@ public function getMigrationsColumnLength() : int
...
+    }
+
+    public function getMigrationsExecutedAtColumnName() : string
+    {
+        return $this->migrationsExecutedAtColumnName;
+    }
+
+    public function getQuotedMigrationsExecutedAtColumnName() : string
+    {
+        return $this->getDependencyFactory()
```
Majkl578: To be honest, this thing is just a service locator now... Wiring everything should imho be done in the SF Bundle / Zend Module and/or in a factory/builder object.

lib/Doctrine/Migrations/MigrationTableStatus.php
```diff
@@ -0,0 +1,72 @@
...
+        if ($this->upToDate !== null) {
+            return $this->upToDate;
+        }
+
+        $table = $this->schemaManager->listTableDetails($this->migrationTable->getName());
+
+        $this->upToDate = true;
+
+        foreach ($this->migrationTable->getColumnNames() as $columnName) {
+            if (! $table->hasColumn($columnName)) {
```
Majkl578: can be inverted with `continue`

lib/Doctrine/Migrations/MigrationTableUpdater.php
```diff
@@ -0,0 +1,86 @@
...
+    {
+        $fromTable = $this->getFromTable();
+        $toTable   = $this->migrationTable->getDBALTable();
+
+        $fromSchema = $this->createSchema([$fromTable]);
+        $toSchema   = $this->createSchema([$toTable]);
+
+        $queries = $fromSchema->getMigrateToSql($toSchema, $this->platform);
+
+        foreach ($queries as $query) {
```
Majkl578: Shouldn't this actually run inside transaction sice there are multiple queries?

lib/Doctrine/Migrations/Configuration/Configuration.php
jwage: It can be done, but not without a big BC break. The only place it is being used like a service locator is in places where I wanted to maintain BC. Like here. Also, these are all internal services so it feels weird to put the burden on the end user to wire all these classes up that they should not even know about or have access to.

jwage: To be clear, I am referring to developers who don't use Symfony, Zend, etc. and have their app integrated with Doctrine DBAL and Migrations. Today all that is required to configure migrations is setting up one service that is passed your DBAL connection:

```
$configuration = new Configuration($connection);
// configure your migrations
$configuration->...;
```

In 2.0 this is the same. All the existing setups will just continue to work. All the services that make up the internals are not meant to be public facing services. If I put the wiring in the framework integrations, wouldn't this mean that when I refactor the internals, it'd be a BC break in the framework integrations due to the wiring would needing to be updated?

lib/Doctrine/Migrations/MigrationTableUpdater.php
Majkl578: From my experience, PostgreSQL definitely does, MySQL definitely does not.
More vendors documented here: https://wiki.postgresql.org/wiki/Transactional_DDL_in_PostgreSQL:_A_Competitive_Analysis

jwage: I added the transaction, it won't hurt :+1:

lib/Doctrine/Migrations/MigrationTableUpdater.php
```diff
@@ -0,0 +1,97 @@
...
+    {
+        $fromTable = $this->getFromTable();
+        $toTable   = $this->migrationTable->getDBALTable();
+
+        $fromSchema = $this->createSchema([$fromTable]);
+        $toSchema   = $this->createSchema([$toTable]);
+
+        $queries = $fromSchema->getMigrateToSql($toSchema, $this->platform);
+
+        $this->connection->beginTransaction();
```
stof: this should be done only if the platform supports transational DDL IMO (which DBAL knows)

jwage: DBAL doesn't know this currently. It only has the supportsTransactions() method. We discussed submitting an enhancement to the DBAL for this though. For now though, we figured having this here for everyone wouldn't hurt. Thoughts?

stof: ah indeed, I though it was something known by DBAL and used by doctrine/migrations, but doctrine/migrations always wraps the migration in a transaction, even if it would not be able to rollback it fully due to non transactional DDL (and migrations can ask to disable the usage of the transaction).

So yeah, keep it like that.

Issue #647, opened 2018-05-07
Use Symfony Stopwatch component for internal timing


jwage:
Right now we use microtime. Swap this out with the Symfony Stopwatch component.

PR #681, opened 2018-05-16
Use Symfony Stopwatch for timing and memory.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #647

#### Summary

Use the Symfony stopwatch for timing and memory monitoring.


lib/Doctrine/Migrations/Stopwatch.php
```diff
@@ -0,0 +1,18 @@
...
+namespace Doctrine\Migrations;
+
+use Symfony\Component\Stopwatch\Stopwatch as SymfonyStopwatch;
+use Symfony\Component\Stopwatch\StopwatchEvent;
+
+class Stopwatch
+{
+    public function start(string $section) : StopwatchEvent
+    {
+        $stopwatch = new SymfonyStopwatch(true);
```
stof: why creating a new SymfonyStopwatch each time ?

jwage: @stof This may have been a misunderstanding on my part of how Stopwatch behaves. Looking at the code now I see that I can have one instance of it. I will make a PR to change it.

Issue #594, opened 2018-01-25
Possibility to make the whole migrate command transactional


stof:
Currently, each migration gets wrapped in a transaction, ensuring that we don't execute half of the migration in case of failure (for platforms supporting to put DDL statements in a transaction, which is the case of PostgreSQL but was not the case of MySQL last time I used it).
However, if you have multiple unexecuted migrations, it could be useful to ask the `migrations:migrate` command to wrap the whole execution in a transaction, so that the whole execution of the command is atomic. My use case is to be able to have a all-or-nothing migration during a deployment process, even in case multiple features shipping a DB migrations are deployed in the same deployment rather than multiple ones.

What do you think about this feature ?

PowerKiKi:
[Atomic DDL](https://dev.mysql.com/doc/refman/8.0/en/atomic-ddl.html) just landed in the brand new  MySQL 8. Unfortunately it seems it is [not supported in MariaDB](https://mariadb.com/kb/en/library/start-transaction/#ddl-statements) yet. But that would still be awesome to support it in this project.

stof:
Atomic DDL still does not allow to perform DDL statements in a transaction involving other SQL statements (which is almost always required as soon as your migrations apply on a non-empty database)

lcobucci:
@stof I think that it's useful and it could be set in the configuration file for the project, so the user can specify the granularity level of the transaction.

jwage:
@stof Do you think it is best to make this an option on the `migrate` command or a configuration option? Or both?

Majkl578:
Info about support by other vendors: https://wiki.postgresql.org/wiki/Transactional_DDL_in_PostgreSQL:_A_Competitive_Analysis

stof:
@jwage I would like to be able to provide this in the configuration, to avoid having to tell each of our devs to use it in the CLI. but I'm fine if it available in both places.

stof:
Note that DBAL already has a platform API to know whether the platform supports transactional DDL. And doctrine/migrations already uses this API to wrap each migration in a transaction when supported.
This feature request is about being able to switch between `transaction per migration` to `single transaction per run` (and it would stay on `no transaction` when unsupported, like today)

PowerKiKi:
I would rather have no configuration at all and always use the highest level of atomicity as possible. This would be consistent with existing behavior, less things to implement, and you would still be able to execute a single migration with the `migrations:execute`command should you need to have a partially migrated db.

PR #683, opened 2018-05-17
Implement all or nothing transaction strategy for migrations.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | yes
| Fixed issues | #594

#### Summary

Implement all or nothing transaction strategy for migrations.

### DIFF
--- a/docs/en/reference/installation.rst
+++ b/docs/en/reference/installation.rst
@@ -70,6 +70,8 @@ file like the following:
 
         <migrations-directory>/path/to/migrations/classes/DoctrineMigrations</migrations-directory>
 
+        <all-or-nothing>1</all-or-nothing>
+
     </doctrine-migrations>
 
 Of course you could do the same thing with a *configuration.yml* file:
@@ -82,6 +84,7 @@ Of course you could do the same thing with a *configuration.yml* file:
     column_name: version
     executed_at_column_name: executed_at
     migrations_directory: /path/to/migrations/classes/DoctrineMigrations
+    all_or_nothing: true
 
 With the above example, the migrations tool will search the ``migrations_directory``
 recursively for files that begin with ``Version`` followed one to 255 characters
--- a/docs/en/reference/introduction.rst
+++ b/docs/en/reference/introduction.rst
@@ -63,7 +63,7 @@ commands to our `Doctrine Command Line Interface <http://doctrine-orm.readthedoc
         new \Doctrine\DBAL\Migrations\Tools\Console\Command\VersionCommand()
     ));
 
-Additionally you have to make sure the 'db' and 'dialog' Helpers are added to your Symfony
+Additionally you have to make sure the ``db`` and ``dialog`` Helpers are added to your Symfony
 Console HelperSet.
 
 .. code-block:: php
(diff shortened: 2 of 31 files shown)


lib/Doctrine/Migrations/Configuration/XML/configuration.xsd
```diff
@@ -24,6 +24,7 @@
                     </xs:simpleType>
                 </xs:element>
                 <xs:element type="xs:string" name="migrations-directory" minOccurs="0" maxOccurs="1"/>
+                <xs:element type="xs:string" name="all-or-nothing" minOccurs="0" maxOccurs="1"/>
```
stof: shoudln't it be `xs:bool` ?

docs/en/reference/installation.rst
```diff
@@ -70,6 +70,8 @@ file like the following:
 
         <migrations-directory>/path/to/migrations/classes/DoctrineMigrations</migrations-directory>
 
+        <all-or-nothing>1</all-or-nothing>
```
OskarStark: Here it is 1...

docs/en/reference/introduction.rst
```diff
@@ -185,18 +185,87 @@ the migrations are executed in the correct order.
...
+
+            <name>Doctrine Sandbox Migrations</name>
+
+            <migrations-namespace>DoctrineMigrations</migrations-namespace>
+
+            <table name="doctrine_migration_versions" />
+
+            <migrations-directory>/path/to/migrations/classes/DoctrineMigrations</migrations-directory>
+
+            <all-or-nothing>true</all-or-nothing>
```
OskarStark: ... here it is true.

Should be consistent

jwage: The docs were updated and fixed at some point in the future. FYI, we take the string from the XML and pass it through this class https://github.com/doctrine/migrations/blob/master/lib/Doctrine/Migrations/Tools/BooleanStringFormatter.php here https://github.com/doctrine/migrations/blob/master/lib/Doctrine/Migrations/Configuration/XmlConfiguration.php#L89

OskarStark: Thanks for the feedback

OskarStark: Thanks for the feedback

docs/en/reference/introduction.rst
```diff
@@ -185,18 +185,87 @@ the migrations are executed in the correct order.
...
 If you specify your own migration classes (like `DoctrineMigrations\NewMigration` in the previous
 example) you will need an autoloader unless all those classes begin with the prefix Version*,
 for example path/to/migrations/classes/VersionNewMigration.php.
+
+All or Nothing Migrations
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you want your migrations to be all or nothing then you can configure your migrations behave that way.
+This means when you executed multiple migrations in a row, the whole migration will be wrapped
+in a single migration and if one of the migrations fails, the transaction will be rolled back.
```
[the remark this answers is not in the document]
jwage: Looks like it. Can you check if it still exists in master and record an issue to fix it if so?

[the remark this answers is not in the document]
apfelbox: @jwage oh, didn't realize that the PR is so old 😅 

The whole paragraph is gone in 2.0 and I couldn't find it in 1.8 either (that's where this PR landed).

So probably already fixed then. 👍 
Sorry for the noise.

PR #684, opened 2018-05-17
Add shorter console command aliases to improve developer experience.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Add shorter console command aliases to improve the developer experience.

Before:

    $ ./migrations migrations:generate

After:

    $ ./migrations generate


weaverryan:
I like it. It's already obvious these commands are for "migrations" because they're in a `doctrine-migrations` executable.

PR #685, opened 2018-05-17
Update console runner to match other Doctrine libs.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

I noticed that this was different on the other Doctrine projects so I updated it to match.


bin/doctrine-migrations.php
```diff
@@ -77,5 +77,6 @@
     ? include 'migrations-output.php'
```
stof: these input and output extension points are not supported anymore (so should be dropped)

jwage: @stof removed here #686

PR #686, opened 2018-05-17
Remove unused input and output extension points from bin/doctrine-mig…

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | https://github.com/doctrine/migrations/pull/685/files#r189024397

#### Summary

Removes unused input and output extension points in bin/doctrine-migrations.php

### DIFF
--- a/bin/doctrine-migrations.php
+++ b/bin/doctrine-migrations.php
@@ -69,14 +69,6 @@ if(class_exists('\Symfony\Component\Console\Helper\QuestionHelper')) {
     $helperSet->set(new DialogHelper(), 'dialog');
 }
 
-$input = file_exists('migrations-input.php')
-    ? include 'migrations-input.php'
-    : null;
-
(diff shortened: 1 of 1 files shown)


stof:
Note that I'm not the only one finding these extension points useless (even if we forget that your previous PR broke them entirely): https://github.com/doctrine/migrations/commit/b79452ae784f490d580ddf0b5a7299abb06cd3d8#commitcomment-277496

jwage:
@stof I am fine with removing them to keep things consistent with the other Doctrine projects. They can always create their own console script if they want custom input/output.

stof:
thus, I don't see what the use case for replacing the Input and Output objects are when running the CLI.
Use cases for customizing the Output that I know of are related to other usages of the API (collecting the output in a BufferOutput to send it elsewhere for instance).

Issue #624, opened 2018-05-05
Upgrade to Box 3


theofidry:
Before I start to try to replace the existing Box2 with Box3, I would like to know if there is any e2e tests to ensure the generated PHAR is working.

/cc @mikeSimonson

jwage:
I don't think there are unfortunately.

mikeSimonson:
The e2e test is me running some version and status command on my local machine ;p

theofidry:
Haha fair, ok follow up in #270 I guess

jwage:
Can we do this for 2.0?

theofidry:
Sure

jwage:
@theofidry I am going to start work on this.

theofidry:
Cool. Don't hesitate to ping me on Slack or Twitter if you have an issue

PR #691, opened 2018-05-18
Upgrade to Box 3 for phar generation.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #624

#### Summary

Upgrade to Box 3 for PHAR generation.


box.json
```diff
@@ -1,23 +1,11 @@
 {
     "alias": "doctrine-migrations.phar",
-    "compactors": [
```
theofidry: I don't think it's gonna make a big difference, but you can keep the compactors, I would replace `Composer` with `Php` though

build-phar.sh
```diff
@@ -8,10 +8,4 @@ fi
 
 composer install --no-dev --optimize-autoloader
```
theofidry: you shouldn't need that line: Box dump the optimized autoloader and already skips the dev deps

jwage:
@theofidry I committed the box.phar since the tests rely on it. I guess the test could download it each time but I am not sure what is worse, relying on the internet for a test or checking in box.phar :)

composer.json
```diff
@@ -53,5 +53,8 @@
     },
     "bin": [
         "bin/doctrine-migrations"
-    ]
+    ],
+    "platform": {
+        "php": "7.1.x"
```
Majkl578: @jwage Why exactly was this added?
a) it's invalid here, it must be in `"config"` node (has no effect here)
b) using `7.1.x` is invalid, it must be a specific version (i.e. `7.1.3`)
c) having this here will likely cause issues with SF4 since Migrations require `^7.1` and platform override therefore should be `7.1.0`

.gitignore
```diff
@@ -5,7 +5,6 @@ reports/
 dist/
 tests/phpunit.xml
 vendor/
-box.phar
```
stof: If you commit it in the repo, you should at least ignore it from the archive (using `.gitattributes`) so that users of the package don't get the box phar downloaded in their projects.

Majkl578: This was already reverted, see #694 + #700.

composer.json
jwage: It was suggested by @theofidry but i must have gotten it wrong. Can you record an issue and I will fix today.

Majkl578: #701

PR #692, opened 2018-05-18
Upgrade composer dependencies.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Upgrade composer dependencies with `composer update`.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -1343,16 +1343,16 @@
         },
         {
             "name": "nette/bootstrap",
-            "version": "v2.4.5",
+            "version": "v2.4.6",
             "source": {
                 "type": "git",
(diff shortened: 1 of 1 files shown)


Majkl578:
Btw the point of commited lock file is to keep PHPCS, PHPStan and related at fixed version, it's irrelevant for the rest, so not needed to update unless you are upgrading CS/SA or to fix a bug.

jwage:
@Majkl578 Got it ok. I just saw that some things had new versions so I figured I would update.

Majkl578:
Yeah, it can be quite annoying to see outdated composer.lock. 😆 
Different approaches were discussed in https://github.com/doctrine/coding-standard/pull/30 and this one won.

Issue #348, opened 2015-08-24
Recursive regex finder doesn't allow for multiple namespaces


liquorvicar:
This may be a feature rather than a bug per se but it confused the hell of out of me for a while. If you pass a namespace into the RecursiveRegexFinder, then all classes that it finds will be prefixed with that namespace. In practice this doesn't currently cause any issues as the files are included at the point they are identified in the file system so the classes should always be loaded.

If you don't pass in a namespace, the classes get added to the global namespace.

It's a bit annoying stylistically as to keep compatibility with the PSRs I'd like to namespace migrations in different folders properly.

Could we perhaps grab the namespace from the file (if there is one) if no namespace is passed in?

mikeSimonson:
What would be your use case for that and won't it add to many different behaviour ?

liquorvicar:
No real use case at the moment (as I said it works fine); it just bothers me that the migration classes don't follow PSR-compatible namespaces if you use multiple dirs. It did cause me some headaches when trying to fix something else as what worked in my app in vagrant didn't work in the tests. In the app the migration classes are loaded fine as the file is required so the class is loaded. In the tests it couldn't find all the migration classes.

I think the behaviour should be fairly straightforward and should retain BC, i.e.
- if you pass a namespace in then it uses that,
- otherwise, if you pass in null, it tries to get the namespace from the migration file itself
- if it can't get the namespace, it uses the global namespace

It would mean a change to the bundle to allow setting the namespace to null in the config but otherwise should work ok.

mikeSimonson:
@liquorvicar How do you get the namespace from the file itself ?

liquorvicar:
@mikeSimonson grep/preg etc Unless there's a clever way in php to get the meta info of the last class loaded/compiled etc

chrisguitarguy:
@mikeSimonson I think the best approach here would be to not define the namespace in configuration. Instead, recursively search the directory (or directories) and find all files that match the pattern and include them. Then [`get_declared_classes`](http://php.net/manual/en/function.get-declared-classes.php) and check if the classes are in one of the directories. This is how some of the [doctrine persistence stuff works](https://github.com/doctrine/common/blob/773c56e51fc26ae07a47146fae31b320dd6cada3/lib/Doctrine/Common/Persistence/Mapping/Driver/AnnotationDriver.php#L193).

jwage:
[...]

PR #693, opened 2018-05-20
Allow Migration Classes to be in Multiple Namespaces

chrisguitarguy:
|      Q       |   A
|------------- | -----------
| Type         | Improvement
| BC Break     | yes
| Fixed issues | #348

#### Summary

This changes the various `MigrationFinder` implementations to use reflection and `get_declared_classes` to load the various migration classes. The code is similar to the [doctrine persistence stuff](https://github.com/doctrine/common/blob/de19a0b00b7a3e5f8820141f11ddf26ee0a7d26e/lib/Doctrine/Common/Persistence/Mapping/Driver/AnnotationDriver.php#L175-L237)

The end result is that all the code for loading migrations no longer cares about the namespace in the configuration -- it's only for generating new migrations now. Users can nest namespaces as they wish.

### DIFF
--- a/lib/Doctrine/Migrations/Finder/Finder.php
+++ b/lib/Doctrine/Migrations/Finder/Finder.php
@@ -5,14 +5,16 @@ declare(strict_types=1);
 namespace Doctrine\Migrations\Finder;
 
 use InvalidArgumentException;
+use ReflectionClass;
 use const PHP_EOL;
 use const SORT_STRING;
-use function basename;
+use function get_declared_classes;
+use function in_array;
 use function is_dir;
+use function ksort;
 use function realpath;
 use function sprintf;
 use function substr;
-use function uasort;
 
 abstract class Finder implements MigrationFinder
 {
@@ -50,7 +52,7 @@ abstract class Finder implements MigrationFinder
             $includedFiles[] = realpath($file);
         }
 
-        $classes = $this->loadMigrationClasses($includedFiles);
+        $classes  = $this->loadMigrationClasses($includedFiles);
         $versions = [];
         foreach ($classes as $class) {
             $version = substr($class->getShortName(), 7);
@@ -69,11 +71,18 @@ abstract class Finder implements MigrationFinder
         return $versions;
     }
 
+    /**
+     * Look up all declared classes and find those classes contained
+     * in the give `$files` array.
+     *
+     * @param string[] $files The set of files that were `required`
+     * @return ReflectionClass[] the classes in `$files`
+     */
     protected function loadMigrationClasses(array $files) : array
     {
         $classes = [];
         foreach (get_declared_classes() as $class) {
-            $ref = new \ReflectionClass($class);
+            $ref = new ReflectionClass($class);
             if (in_array($ref->getFileName(), $files)) {
                 $classes[] = $ref;
             }
--- a/lib/Doctrine/Migrations/Finder/Finder.php
+++ b/lib/Doctrine/Migrations/Finder/Finder.php
@@ -71,7 +71,7 @@ abstract class Finder implements MigrationFinder
 
     /**
      * Look up all declared classes and find those classes contained
-     * in the give `$files` array.
+     * in the given `$files` array.
      *
      * @param string[] $files The set of files that were `required`
      * @param string|null $namespace If not null only classes in this namespace will be returned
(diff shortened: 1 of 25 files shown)


lib/Doctrine/Migrations/Finder/Finder.php
```diff
@@ -39,35 +42,52 @@ protected function getRealPath(string $directory) : string
...
                     $version
                 ));
             }
-
-            $migrations[$version] = sprintf('%s\\%s', $namespace, $className);
+            $versions[$version] = $class->getName();
         }
 
-        return $migrations;
+        ksort($versions, SORT_STRING);
```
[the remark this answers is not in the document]
jwage: Probably not needed. I would make sure we have test coverage for the sorting and then remove it and see if anything breaks.

chrisguitarguy:
> Couldn't this change be made BC? Basically, keep on allowing a namespace to be passed optionally and if it is passed, filter the migrations it finds, to only the ones in that namespace.

Probably? It's not really a BC break as it is now: whatever migrations a person has that already exist before this change will still load correctly.

Maybe a better way to make it more BC friendly is to filter out anything that doesn't extend the `AbstractMigration` so nothing unexpected gets loaded. Maybe also filter out anything that's an abstract class.

jwage:
@chrisguitarguy The reason I ask is because, I think as it is right now, the migration namespace setting goes unused. So we either need to deprecate and remove that setting or keep it there. I think we should keep it there and enforce the filtering with that setting and make sure it is optional everywhere. This way things are guaranteed to just keep working the way they did before.

chrisguitarguy:
> I think as it is right now, the migration namespace setting goes unused

It's used in [code generation](https://github.com/chrisguitarguy/migrations/blob/9c52cdd139d398fd56b8926ca43986b4e1890cda/lib/Doctrine/Migrations/MigrationGenerator.php#L82). So another path could be renaming the setting to `default_namespace` or something like that.

As long as existing migrations load correctly (which they already won't do to `void` typehints added to `AbstractMigration::{up,down}`), then I wouldn't add another flag options to enable or disable this stuff.

chrisguitarguy:
Unless you want to target something in a 1.7.x release that deprecate things -- similar to how you have to opt into behavior into new behavior symfony upgrades (thinking specifically of some form stuff).

jwage:
I would prefer to keep it all BC and not remove or rename anything.

lib/Doctrine/Migrations/Finder/Finder.php
```diff
@@ -39,35 +42,52 @@ protected function getRealPath(string $directory) : string
...
+
+        return $versions;
     }
 
-    protected function getFileSortCallback() : callable
+    /**
+     * Look up all declared classes and find those classes contained
+     * in the give `$files` array.
```
jwage: Typo `give` should be `given`

lib/Doctrine/Migrations/Finder/Finder.php
```diff
@@ -39,35 +42,52 @@ protected function getRealPath(string $directory) : string
...
+     * @return ReflectionClass[] the classes in `$files`
+     */
+    protected function loadMigrationClasses(array $files) : array
     {
-        return function (string $a, string $b) {
-            return (basename($a) < basename($b)) ? -1 : 1;
-        };
+        $classes = [];
+        foreach (get_declared_classes() as $class) {
+            $ref = new ReflectionClass($class);
```
jwage: We like to avoid using abbreviations. Please use `$reflectionClass` as the variable name here.

lib/Doctrine/Migrations/Finder/GlobFinder.php
```diff
@@ -12,12 +12,12 @@
     /**
      * @return string[]
      */
-    public function findMigrations(string $directory, ?string $namespace = null) : array
+    public function findMigrations(string $directory) : array
```
jwage: As discussed, please reintroduce the `$namespace` argument as optional and if it is provided, filter the classes returned to only be the ones in that namespace.

chrisguitarguy:
@jwage changes made! Sorry it took me so long to get back to this.

Do we need to add the configuration option here to enable multiple namespaces?

jwage:
@chrisguitarguy We could do that in another PR to allow multiple namespaces to be specified for the filtering. I think this is probably fine for this PR.

Issue #701, opened 2018-05-30
PHP platform in composer.json root node is set incorrectly


Majkl578:
https://github.com/doctrine/migrations/pull/691#discussion_r191726668

> a) it's invalid here, it must be in "config" node (has no effect here)
> b) using 7.1.x is invalid, it must be a specific version (i.e. 7.1.3)
> c) having this here will likely cause issues with SF4 since Migrations require ^7.1 and platform override therefore should be 7.1.0

PR #704, opened 2018-05-30
Fix composer.json platform.php setting.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #701

#### Summary

This was something that @theofidry suggested we add to composer.json for the box build process. I admit I don't fully understand why. Maybe someone can chime in here on why it is needed and what the value should be.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,10 @@
         }
     },
     "config": {
-        "sort-packages": true
+        "sort-packages": true,
+        "platform": {
+            "php": "7.1.0"
+        }
     },
     "extra": {
         "branch-alias": {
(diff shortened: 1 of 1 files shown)


theofidry:
>This in the end affects testing of lowest/latest dependencies too and will always i.e. reject Symfony 4 completely on CI.

Not if you remove it:

>>Note that this config should be unset for the build not building the PHAR with for example with composer config --unset platform.php

If you don't set it by default you must not forget to add it when building the PHAR locally, so maybe it can be added to the build script instead.

Majkl578:
See alternative apporach in #705.

jwage:
@Majkl578 I like that approach better :+1: closing this PR.

Issue #702, opened 2018-05-30
Migrations executedAt should either preserve timezone or always use UTC


Majkl578:
Time of execution is currently logged without timezone and without conversion to standardized timezone (UTC).
This may lead to surprising behavior if timezone ever changes.

Two possible approaches:
a) always log execution time in UTC time
b) log execution time with timezone/offset (using `datetimetz_immutable` type)

Approach a) is simpler, but approach b) does not cause data loss, so which one is better? :)

Either way, the output should be in _current_ timezone (as per `date_default_timezone_get()`).

PR #706, opened 2018-05-30
Always store executed_at in UTC and render using date_default_timezone()

jwage:
|      Q       |   A
|------------- | -----------
| Type         | enhancement
| BC Break     | no
| Fixed issues | #702

#### Summary

Always store executed_at in UTC and render using date_default_timezone()

### DIFF
--- a/lib/Doctrine/Migrations/Version.php
+++ b/lib/Doctrine/Migrations/Version.php
@@ -5,12 +5,14 @@ declare(strict_types=1);
 namespace Doctrine\Migrations;
 
 use DateTimeImmutable;
+use DateTimeZone;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Types\Type;
 use Doctrine\Migrations\Configuration\Configuration;
 use Doctrine\Migrations\Exception\MigrationNotConvertibleToSql;
 use function assert;
 use function count;
+use function date_default_timezone_get;
 use function in_array;
 use function str_replace;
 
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Migrations/Version.php
```diff
@@ -221,7 +226,7 @@ private function getWriteSqlFileMigratorConfig() : MigratorConfig
     private function getExecutedAtDatabaseValue() : string
     {
         return Type::getType(MigrationTable::MIGRATION_EXECUTED_AT_COLUMN_TYPE)->convertToDatabaseValue(
-            new DateTimeImmutable(),
+            (new DateTimeImmutable('now', new DateTimeZone(date_default_timezone_get())))->setTimezone(new DateTimeZone('UTC')),
```
Majkl578: You can drop `new DateTimeZone(date_default_timezone_get()))`, it's default behavior.

tests/Doctrine/Migrations/Tests/VersionTest.php
```diff
@@ -599,6 +605,48 @@ public function testRunWithInsertNullValue() : void
...
     }
 
+    /**
+     * @dataProvider getExecutedAtTimeZones
+     */
+    public function testExecutedAtTimeZone(string $timeZone) : void
+    {
+        $originalTimeZone = date_default_timezone_get();
+
+        date_default_timezone_set($timeZone);
```
Majkl578: I'd put whole test after this statement inside `try { ... }` and then `date_default_timezone_set($originalTimeZone);` inside `finally`, otherwise when the test fails, the timezone will stay messed up.

tests/Doctrine/Migrations/Tests/VersionTest.php
```diff
@@ -599,6 +605,48 @@ public function testRunWithInsertNullValue() : void
...
+    }
+
+    /**
+     * @return string[][]
+     */
+    public function getExecutedAtTimeZones() : array
+    {
+        return [
+            ['America/New_York'],
+            ['Indian/Chagos'],
```
Majkl578: Would maybe also add  `'UTC'`? :)

tests/Doctrine/Migrations/Tests/VersionTest.php
morozov: FIWI, there's `TestCase::iniSet()` which could be use to simplify the cleanup.

Majkl578: Cool, didn't know it exists. 👍

Issue #682, opened 2018-05-17
Reorganize classes in root Doctrine\Migrations namespace in to folders


jwage:
We have a few logical groupings that have surfaced through the refactoring so I think we can do a reorg of the classes before releasing 2.0

PR #707, opened 2018-05-31
Reorganize classes for 2.0

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #682

#### Summary

As we refactored, several new classes were created and put in the root namespace. Move the logical groupings in to folders.


.gitignore
```diff
@@ -9,3 +9,5 @@ phpunit.xml
 /phpcs.xml
 /.phpcs-cache
 /box.phar
+/.idea
```
morozov: IMO, this line belongs to a personal `.gitignore` file but not to the project, unless the project is tightly coupled to the IDE.

jwage: I didn't know you could do that. I will remove it and do that. Thanks!

jwage: Removed here https://github.com/doctrine/migrations/commit/d3d7a8ce0e5133c7c48b46cbaf9ee0ee48191492

Majkl578: > IMO, this line belongs to a personal `.gitignore` file 

:+1: 

There is nice article from GitHub: https://help.github.com/articles/ignoring-files/#create-a-global-gitignore

PR #709, opened 2018-05-31
CS on return types

Remg:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | /

#### Summary

This PR only removes an extra space before semicolons on generated return types.

https://github.com/php-fig/fig-standards/blob/master/proposed/extended-coding-style-guide.md#45-method-and-function-arguments
> When you have a return type declaration present there MUST be one space after the colon followed by the type declaration. The colon and declaration MUST be on the same line as the argument list closing parentheses with no spaces between the two characters.


Majkl578:
The generated code follows Doctrine Coding Standard which requires space before colon.
See discussion here: https://github.com/doctrine/migrations/pull/577#discussion_r155620732

Issue #710, opened 2018-05-31
Mark classes internal that are internal


jwage:
There really isn't much that the outside should be depending on in this library/tool so mark everything as @internal that nobody should be depending on.

stof:
well, I would say that the `Tools/Console` aprt should not be considered as internal, otherwise DoctrineMigrationsBundle reusing these commands to register them directly in the Symfony CLI would have to be considered as invalid (from the bundle perspective, we need to be able to rely on backward compatibility for these parts in non-major releases, and that means not being internal). 
The Configuration part should also be considered public API (even the built-in runner still relies on receiving a configuration)

jwage:
Yes, the commands, configuration, etc. wouldn't be considered internal.

stof:
well, the initial description says `mark everything as @internal`

jwage:
Updated the description to be more accurate.

PR #711, opened 2018-05-31
Mark classes as internal that nobody should be depending on.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #710

#### Summary

Mark classes as internal that nobody should be depending on. The only thing I see that people should be depending on are:

- Doctrine\Migrations\Configuration\Configuration
- Doctrine\Migrations\Events
- Doctrine\Migrations\Finder\MigrationFinder
- Doctrine\Migrations\Provider\SchemaProviderInterface
- Doctrine\Migrations\Provider\StubSchemaProvider
- Doctrine\Migrations\AbstractMigration
   - isTransactional
   - getDescription
   - warnIf
   - abortIf
   - skipIf
   - preUp
   - postUp
   - preDown
   - postDown
   - up
   - down
   - addSql
   - write
   - throwIrreversibleMigrationException
- Doctrine\Migrations\Tools\Console\Command\AbstractCommand
- Doctrine\Migrations\Tools\Console\Command\DiffCommand
- Doctrine\Migrations\Tools\Console\Command\DumpSchemaCommand
- Doctrine\Migrations\Tools\Console\Command\ExecuteCommand
- Doctrine\Migrations\Tools\Console\Command\GenerateCommand
- Doctrine\Migrations\Tools\Console\Command\LatestCommand
- Doctrine\Migrations\Tools\Console\Command\MigrateCommand
- Doctrine\Migrations\Tools\Console\Command\RollupCommand
- Doctrine\Migrations\Tools\Console\Command\StatusCommand
- Doctrine\Migrations\Tools\Console\Command\UpToDateCommand
- Doctrine\Migrations\Tools\Console\Command\VersionCommand

### DIFF
--- a/lib/Doctrine/Migrations/Configuration/ArrayConfiguration.php
+++ b/lib/Doctrine/Migrations/Configuration/ArrayConfiguration.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace Doctrine\Migrations\Configuration;
 
+/**
+ * @internal
+ */
 class ArrayConfiguration extends AbstractFileConfiguration
 {
     /** @inheritdoc */
--- a/lib/Doctrine/Migrations/Provider/OrmSchemaProvider.php
+++ b/lib/Doctrine/Migrations/Provider/OrmSchemaProvider.php
@@ -10,6 +10,9 @@ use Doctrine\ORM\Tools\SchemaTool;
 use UnexpectedValueException;
 use function count;
 
+/**
+ * @internal
+ */
 final class OrmSchemaProvider implements SchemaProviderInterface
 {
     /** @var EntityManagerInterface */
--- a/lib/Doctrine/Migrations/Provider/SchemaDiffProviderInterface.php
+++ b/lib/Doctrine/Migrations/Provider/SchemaDiffProviderInterface.php
@@ -6,6 +6,9 @@ namespace Doctrine\Migrations\Provider;
 
 use Doctrine\DBAL\Schema\Schema;
 
+/**
+ * @internal
+ */
 interface SchemaDiffProviderInterface
 {
     public function createFromSchema() : Schema;
--- a/lib/Doctrine/Migrations/Tools/Console/Helper/MigrationDirectoryHelper.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Helper/MigrationDirectoryHelper.php
@@ -6,7 +6,6 @@ namespace Doctrine\Migrations\Tools\Console\Helper;
 
 use Doctrine\Migrations\Configuration\Configuration;
 use InvalidArgumentException;
-use Symfony\Component\Console\Helper\Helper;
 use const DIRECTORY_SEPARATOR;
 use function date;
 use function file_exists;
@@ -15,7 +14,10 @@ use function mkdir;
 use function rtrim;
 use function sprintf;
 
-class MigrationDirectoryHelper extends Helper
+/**
+ * @internal
+ */
+class MigrationDirectoryHelper
 {
     /** @var Configuration */
     private $configuration;
--- a/lib/Doctrine/Migrations/Configuration/AbstractFileConfiguration.php
+++ b/lib/Doctrine/Migrations/Configuration/AbstractFileConfiguration.php
@@ -16,6 +16,9 @@ use function in_array;
 use function realpath;
 use function strcasecmp;
 
+/**
+ * @internal
+ */
 abstract class AbstractFileConfiguration extends Configuration
 {
     /** @var array */
--- a/lib/Doctrine/Migrations/Configuration/Connection/ConnectionLoaderInterface.php
+++ b/lib/Doctrine/Migrations/Configuration/Connection/ConnectionLoaderInterface.php
@@ -6,6 +6,9 @@ namespace Doctrine\Migrations\Configuration\Connection;
 
 use Doctrine\DBAL\Connection;
 
+/**
+ * @internal
+ */
 interface ConnectionLoaderInterface
 {
     /**
--- a/lib/Doctrine/Migrations/Configuration/Connection/Loader/ArrayConnectionConfigurationLoader.php
+++ b/lib/Doctrine/Migrations/Configuration/Connection/Loader/ArrayConnectionConfigurationLoader.php
@@ -11,6 +11,9 @@ use InvalidArgumentException;
 use function file_exists;
 use function is_array;
 
+/**
+ * @internal
+ */
 class ArrayConnectionConfigurationLoader implements ConnectionLoaderInterface
 {
     /** @var null|string */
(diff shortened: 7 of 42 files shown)


jwage:
@Majkl578 @stof Can you take a look? Any classes I marked as internal that shouldn't be?

lib/Doctrine/Migrations/Configuration/ArrayConfiguration.php
```diff
@@ -4,6 +4,9 @@
 
 namespace Doctrine\Migrations\Configuration;
 
+/**
+ * @internal
```
stof: @jwage should these configuration classes really be internal ? How are they used then ?

lib/Doctrine/Migrations/Events.php
```diff
@@ -4,6 +4,9 @@
 
 namespace Doctrine\Migrations;
 
+/**
+ * @internal
```
stof: this must not be internal. The events are extension points of the library, so the event names need to be public (otherwise they are not usable).

lib/Doctrine/Migrations/Finder/MigrationFinder.php
```diff
@@ -4,6 +4,9 @@
 
 namespace Doctrine\Migrations\Finder;
 
+/**
+ * @internal
```
stof: Aren't these meant to be an extension point of the library ? Otherwise, how would people choose a different finder ?

lib/Doctrine/Migrations/Tools/Console/Helper/MigrationDirectoryHelper.php
```diff
@@ -15,6 +15,9 @@
 use function rtrim;
 use function sprintf;
 
+/**
+ * @internal
```
stof: console helpers should not be internal either, as they must be registered in the CLI Application too (`cli-config.php` is expected to instantiate a HelperSet with the appropriate helpers in it)

lib/Doctrine/Migrations/Provider/SchemaDiffProviderInterface.php
```diff
@@ -6,6 +6,9 @@
 
 use Doctrine\DBAL\Schema\Schema;
 
+/**
+ * @internal
```
stof: isn't this meant to be an extension point ?

lib/Doctrine/Migrations/Configuration/ArrayConfiguration.php
jwage: They don't get used directly, they are a part of the internals of the console commands but are never used directly. They produce a `Configuration` instance that then gets used in the console commands.

lib/Doctrine/Migrations/Finder/MigrationFinder.php
jwage: Yes, this shouldn't be internal. Changing it.

lib/Doctrine/Migrations/Tools/Console/Helper/MigrationDirectoryHelper.php
jwage: To be honest I am not sure why this is a symfony console helper. It should not be. It is only used directly here as an internal dependency https://github.com/doctrine/migrations/blob/31051801be855a45a1d34dcad3a92b37abebda24/lib/Doctrine/Migrations/Generator/Generator.php#L91

lib/Doctrine/Migrations/Provider/SchemaDiffProviderInterface.php
jwage: No, only `SchemaProviderInterface` is an extension point that can be passed through here https://github.com/doctrine/migrations/blob/master/lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php

I think the interface might have been created to ease unit testing whenever it was added.

lib/Doctrine/Migrations/Events.php
jwage: Fixing.

lib/Doctrine/Migrations/Tools/Console/Helper/MigrationDirectoryHelper.php
stof: well, this one should not be a console helper then. But other console helpers should not be internal.

lib/Doctrine/Migrations/Configuration/ArrayConfiguration.php
stof: should these classes actually be extending Configuration then ?

lib/Doctrine/Migrations/Provider/SchemaDiffProviderInterface.php
stof: the interface has been added because the lazy one is implemented through decoration

lib/Doctrine/Migrations/Provider/OrmSchemaProvider.php
```diff
@@ -10,6 +10,9 @@
 use UnexpectedValueException;
 use function count;
 
+/**
+ * @internal
```
stof: should not be internal. ORM users should be able to use this implementation of the non-internal `SchemaProviderInterface`

lib/Doctrine/Migrations/Finder/MigrationFinder.php
stof: implementations should not be internal either. Otherwise, the only way to use the extension point is to write your own implementation from scratch, even if you want the same behavior than one of the built-in implementations (as you cannot use them due to being internal and so not covered by BC).

lib/Doctrine/Migrations/Provider/OrmSchemaProvider.php
jwage: This gets used internally here. I don't think it should be public https://github.com/doctrine/migrations/blob/master/lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php#L136

jwage: If they don't provide their own schema provider that implements the `SchemaProviderInterface` then it uses the OrmSchemaProvider.

lib/Doctrine/Migrations/Provider/SchemaDiffProviderInterface.php
jwage: Understood. So I think this should not be public still, right?

lib/Doctrine/Migrations/Configuration/ArrayConfiguration.php
jwage: @stof arguably no, they should not. A better design would be to inject the Configuration in to ArrayConfiguration, etc. and then call the public getters on Configuration and have no inheritance at all. I could try tackling that refactoring in another PR.

lib/Doctrine/Migrations/Provider/SchemaDiffProviderInterface.php
stof: yeah, this one is OK. I confused it with `SchemaProviderInterface`

sserbin:
@jwage was `Doctrine\Migrations\Version\Version` accidentally marked as `@internal`? (it's also not in a list in op)

It seems like it shouldn't have, as `Doctrine\Migrations\Configuration::getMigrations()` is public (as well as a few others Configuration's methods, accepting or returning `Version`s )

lcobucci:
@sserbin the implementation has been rightfully marked as `@internal`. It looks like we forgot to use the interface for those operations, though.

PR #719, opened 2018-06-04
Fix command return type hints so they are consistent and match what S…

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Working on updating DoctrineMigrationsBundle and I noticed the return type hints for the commands were not consistent and inline with what the Symfony console return type is defined as.

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/StatusCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/StatusCommand.php
@@ -47,7 +47,7 @@ EOT
         parent::configure();
     }
 
-    public function execute(InputInterface $input, OutputInterface $output) : void
+    public function execute(InputInterface $input, OutputInterface $output) : ?int
     {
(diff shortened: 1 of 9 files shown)


lib/Doctrine/Migrations/Tools/Console/Command/StatusCommand.php
```diff
@@ -47,7 +47,7 @@ protected function configure() : void
         parent::configure();
     }
 
-    public function execute(InputInterface $input, OutputInterface $output) : void
+    public function execute(InputInterface $input, OutputInterface $output) : ?int
```
Majkl578: The return types shouldn't be nullable, you never return null.
Also if upstream ever adds `?int` return type, removing nullability is valid covariant change.

jwage: I made it match what the Symfony console doc block has currently. You are correct that in our commands we never return null, so we could change this here.

PR #725, opened 2018-06-05
Throw only specialized package-only exceptions

Majkl578:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Stop throwing around SPL exceptions, only throw our own.

BC is kept through inheritance.

### DIFF
--- a/lib/Doctrine/Migrations/Finder/Exception/NameIsReserved.php
+++ b/lib/Doctrine/Migrations/Finder/Exception/NameIsReserved.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Migrations\Finder\Exception;
+
+use InvalidArgumentException;
+use const PHP_EOL;
+use function sprintf;
+
+final class NameIsReserved extends InvalidArgumentException implements FinderException
+{
+    public static function new(string $version) : self
+    {
+        return new self(sprintf(
+            'Cannot load a migrations with the name "%s" because it is a reserved number by Doctrine Migrations.'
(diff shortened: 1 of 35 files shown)


lib/Doctrine/Migrations/Finder/Exception/FInderException.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Migrations\Finder\Exception;
+
+interface FInderException
```
greg0ire: case issue

lib/Doctrine/Migrations/Finder/Exception/NameIsReserved.php
```diff
@@ -0,0 +1,22 @@
...
+use function sprintf;
+
+final class NameIsReserved extends InvalidArgumentException implements FInderException
+{
+    public static function new(string $version) : self
+    {
+        return new self(sprintf(
+            'Cannot load a migrations with the name "%s" because it is a reserved number by Doctrine Migrations.'
+            . PHP_EOL
+            . "It's used to revert all migrations including the first one.",
```
greg0ire: `It is`, and simple quotes?

Majkl578: @greg0ire Some of the messages would indeed benefit from rewording, mind to send a PR? ;)

greg0ire: Will do!

PR #733, opened 2018-06-12
Add phpstan-phpunit to composer.json

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Add phpstan/phpstan-phpunit to composer.json

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -4,4 +4,5 @@ parameters:
         - '#ProxyManager\\Proxy\\VirtualProxyInterface#'
 
 includes:
+    - vendor/phpstan/phpstan-phpunit/extension.neon
     - vendor/phpstan/phpstan-strict-rules/rules.neon
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
         "jdorn/sql-formatter": "^1.1",
         "mikey179/vfsStream": "^1.6",
         "phpstan/phpstan": "^0.9.2",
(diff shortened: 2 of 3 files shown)


localheinz:
Does this also need a modification of [`phpstan.neon`](https://github.com/doctrine/migrations/blob/master/phpstan.neon)?

For reference, see https://github.com/phpstan/phpstan-phpunit#usage:

```
includes:
	- vendor/phpstan/phpstan-phpunit/extension.neon
```

Majkl578:
@localheinz Yes.

jwage:
Whoops. Forgot that. It is updated now.

phpstan.neon
```diff
@@ -5,3 +5,4 @@ parameters:
 
 includes:
     - vendor/phpstan/phpstan-strict-rules/rules.neon
+    - vendor/phpstan/phpstan-phpunit/extension.neon
```
localheinz: Probably not that important, but in case more rules will be imported, I recommend to keep includes sorted, perhaps using the [Lines Sorter](https://plugins.jetbrains.com/plugin/5919-lines-sorter) plugin for PhpStorm.

localheinz:
Build failure appears to be unrelated; restarting the build might fix it.

phpstan.neon
jwage: Fixing. Good call.

PR #740, opened 2018-06-26
Fix: No need to update composer itself twice

localheinz:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | n/a

#### Summary

This PR

* [x] removes a redundant run of `composer self-update` on Travis

💁‍♂️ For reference, see

* https://travis-ci.org/doctrine/migrations/jobs/396429439#L473-L476
* https://travis-ci.org/doctrine/migrations/jobs/396429439#L501-L502

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ cache:
 
 before_install:
   - mv ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini{,.disabled} || echo "xdebug not available"
-  - travis_retry composer self-update
 
 install:
   - rm composer.lock


jwage:
Where is it calling composer self-update already?

localheinz:
@jwage 

See https://travis-ci.org/doctrine/migrations/jobs/396429439#L473-L476, Travis itself ensures that `composer` is self-updated.

jwage:
Ah cool. Thanks for clarifying.

localheinz:
Thank you, @jwage!

PR #745, opened 2018-08-18
Passing custom AbstractMigration class through CLI --parent

jszoja:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary
This comment allows passing --parent CLI parameter to migrations:generate.
In this argument you are defining the full name of the class, which will be used as the parent one for all migrations instead of the default one 'Doctrine\DBAL\Migrations\AbstractMigration'.

I'm working with PHP 5.6, hence I did it on the branch based on v1.5.0, but you may want to integrate it to the next release. If you are interested in having it, I could find out how to set the custom class in config files, eg xml. Otherwise I'm happy to leave it as it is for my use.

I created the sample project using this feature in jszoja/multidb-doctrine-migrations.
Thank for your time.


alcaeus:
I'm going to close this PR: for one, 1.5 is waaaay old and no longer supported. More importantly, 1.6.0 introduced the ability to change the template used to generate migrations, allowing for much more ways to customise the generated migration than just the parent class. Thank you for the contribution!

PR #746, opened 2018-08-30
workaround for libXML bug - https://bugs.php.net/bug.php?id=62577

Piskvor:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

If libxml_disable_entity_loader(true), simplexml_load_file() fails;
this avoids the limitation by avoiding file operations in libXML.

The only difference here is that the file operations are done via file_get_contents() and the resulting string is passed to simplexml_load_string() from PHP, instead of simplexml_load_file() doing these two things internally.

### DIFF
--- a/lib/Doctrine/Migrations/Configuration/XmlConfiguration.php
+++ b/lib/Doctrine/Migrations/Configuration/XmlConfiguration.php
@@ -11,9 +11,10 @@ use SimpleXMLElement;
 use const DIRECTORY_SEPARATOR;
 use const LIBXML_NOCDATA;
 use function assert;
+use function file_get_contents;
 use function libxml_clear_errors;
 use function libxml_use_internal_errors;
-use function simplexml_load_file;
+use function simplexml_load_string;
 
 /**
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Migrations/Configuration/XmlConfiguration.php
```diff
@@ -41,7 +42,11 @@ protected function doLoad(string $file) : void
             throw XmlNotValid::failedValidation();
         }
 
-        $xml = simplexml_load_file($file, SimpleXMLElement::class, LIBXML_NOCDATA);
+        $xmlString = file_get_contents($file);
+        $xml       = false;
+        if ($xmlString !== false) {
```
Majkl578: This should be an assertion instead: `assert($xmlString !== false)`. Also please name the variable i.e. `$rawXml`.

Piskvor: Sure, changed that. Thanks for your time!

jwage:
:+1: It looks good. I am going to rebase/squash and then merge.

PR #747, opened 2018-08-31
Add doctrine-project.json to root of the project.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Add doctrine-project.json to the root of the project for the website build process to read.

Related to https://github.com/doctrine/doctrine-website/pull/218

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -1,0 +1,25 @@
+{
+    "active": true,
+    "name": "Migrations",
+    "slug": "migrations",
+    "docsSlug": "doctrine-migrations",
+    "versions": [
+        {
+            "name": "2.0",
+            "branchName": "master",
+            "slug": "latest",
+            "upcoming": true
+        },
+        {
+            "name": "1.7",
+            "branchName": "1.7",
+            "slug": "1.7",
+            "aliases": [
+                "current",
+                "stable"
+            ],
+            "maintained": true,
+            "current": true
+        }
+    ]
+}
--- a/composer.json
+++ b/composer.json
@@ -1,8 +1,8 @@
 {
     "name": "doctrine/migrations",
     "type": "library",
-    "description": "Database Schema migrations using Doctrine DBAL",
-    "keywords": ["migrations", "database"],
+    "description": "PHP Doctrine Migrations project offer additional functionality on top of the database abstraction layer (DBAL) for versioning your database schema and easily deploying changes to it. It is a very easy to use and a powerful tool.",
+    "keywords": ["php", "database", "migrations", "dbal"],
     "homepage": "https://www.doctrine-project.org/projects/migrations.html",
     "license": "MIT",
     "authors": [


doctrine-project.json
```diff
@@ -0,0 +1,36 @@
+{
+    "active": true,
+    "name": "Migrations",
+    "shortName": "Migrations",
+    "slug": "migrations",
+    "docsSlug": "doctrine-migrations",
+    "composerPackageName": "doctrine/migrations",
```
Majkl578: Can't this be inferred from `composer.json`?

doctrine-project.json
```diff
@@ -0,0 +1,36 @@
...
+    "docsRepositoryName": "migrations",
+    "docsPath": "/docs",
+    "description": "PHP Doctrine Migrations project offer additional functionality on top of the database abstraction layer (DBAL) for versioning your database schema and easily deploying changes to it. It is a very easy to use and a powerful tool.\n",
```
Majkl578: Can't this be inferred from `composer.json`?

doctrine-project.json
```diff
@@ -0,0 +1,36 @@
...
+    "keywords": [
```
Majkl578: Can't this be inferred from `composer.json`?

Majkl578:
Does this mean `master` always has to hold authoritative information about all versions?

Majkl578:
I'd prefer automating both inferring data from `composer.json` and making autodiscovery of the versions (from existing branches, from `branch-alias`). This adds one more burden for new releases. :/

jwage:
We can definitely work towards that as a goal. Should be a relatively easy enhancement.

stof:
the drawback of automatic discovery of branches is that the doc website cares about *maintained* branches AFAIK, and github does not have a way to mark branches as unmaintained.

stof:
I'm clearing the milestone, as this is not really tied to any release of the package

Majkl578:
> and github does not have a way to mark branches as unmaintained.

That is true. We could solve this for example by `.not-maintained` placeholder file.
With the approach here, we would end up with different json in each branch and also additional thing to bump every minor/major release in every project.

stof:
@Majkl578 the project file does not only list the branches. It also defines a stable version, and marks some versions as upcoming. All these might be hard to read from composer.json (and even harder when you consider it comes from *multiple* composer.json, which then might disagree).
Considering that the file in the default branch of the repo is the one taken into account by the website makes sense to me, as the website is working with the repo, not with a given revision of the repo).
The placeholder file might not be good for performance of the builder though, as you would have to check all git branches for this file when building it, instead of knowing upfront which branch you care about.

jwage:
@stof @Majkl578 It is definitely a challenge to have a different doctrine-project.json for each version. We already have to update the doctrine-website repo when we release new things, so this is at least a step forward to where each project can make the change in their own repo instead of having to make the change in the website code.

PR #757, opened 2018-10-08
Bump Box

theofidry:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

- Update Box to the latest stable release (3.1.0)
- Remove redundant settings
- Rename `box.json` to `box.json.dist` and add `box.json` to the ignored files to allow one to tweak
  the Box settings locally easily
- Tweak the Box compilation script to make it more resilient in case of failure

### DIFF
--- a/build-phar.sh
+++ b/build-phar.sh
@@ -1,11 +1,19 @@
 #!/usr/bin/env bash
 
+set -euf -o pipefail
+
 ./download-box.sh
+
+function restorePlatform {
+    composer config --unset platform
+    mv -f composer.lock.back composer.lock || true
+}
 
 # lock PHP to minimum allowed version
 composer config platform.php 7.1.0
+cp composer.lock composer.lock.back || true
 composer update
 
 php box.phar compile -vv
 
-composer config --unset platform
+trap restorePlatform exit
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,7 +1,7 @@
 build/*                 export-ignore
 /tests                  export-ignore
 /.*                     export-ignore
-box.json                export-ignore
+box.json.dist           export-ignore
 build.properties.dev    export-ignore
 build.xml               export-ignore
 /phpunit.xml.dist       export-ignore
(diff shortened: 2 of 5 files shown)


build-phar.sh
```diff
@@ -1,11 +1,19 @@
...
+
 # lock PHP to minimum allowed version
 composer config platform.php 7.1.0
+cp composer.lock composer.lock.back || true
 composer update
 
 php box.phar compile -vv
 
-composer config --unset platform
+trap restorePlatform exit
```
Majkl578: This trap should be before running `compile`, otherwise it has basically no effect here at the end.

build-phar.sh
```diff
@@ -1,11 +1,19 @@
 #!/usr/bin/env bash
 
+set -euf -o pipefail
+
 ./download-box.sh
 
+function restorePlatform {
```
Majkl578: Hmm, this is a rare signature for bash functions IMO.

build-phar.sh
```diff
@@ -1,11 +1,19 @@
...
+    composer config --unset platform
+    mv -f composer.lock.back composer.lock || true
+}
+
 # lock PHP to minimum allowed version
 composer config platform.php 7.1.0
+cp composer.lock composer.lock.back || true
```
Majkl578: Why ".back"? Standard extension for backups is ".bak" or more preferably ".orig".

build-phar.sh
theofidry: ha indeed :/

build-phar.sh
theofidry: ¯\_(ツ)_/¯

build-phar.sh
theofidry: wdym?

Majkl578: Usually bash functions are written as such:
```bash
functionName() {
    body
}
```

theofidry: Ah right, it's been a while since I've done some bash so I just went with whatever though was correct and my IDE didn't complain either... Can be renamed I guess

PR #758, opened 2018-10-12
Added missing commands in integrations doc

arneee:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | Corrected docs

#### Summary

Added missing command in integration chapter of doctrine-migrations


jwage:
Thanks!

PR #765, opened 2018-11-05
Phpstan tests

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Also enabling PHPStan in tests/ on L7 + strict.

### DIFF
--- a/lib/Doctrine/Migrations/OutputWriter.php
+++ b/lib/Doctrine/Migrations/OutputWriter.php
@@ -12,10 +12,15 @@ class OutputWriter
     /** @var callable */
     private $callback;
 
+    /** @var string */
+    private $lastMessage = '';
+
     public function __construct(?callable $callback = null)
     {
         if ($callback === null) {
-            $callback = static function ($message) : void {
+
+            $callback = static function (string $message) : void {
+                $this->lastMessage = $message;
             };
         }
 
@@ -30,5 +35,10 @@ class OutputWriter
     public function write(string $message) : void
     {
         ($this->callback)($message);
+    }
+
+    public function getLastMessage() : string
+    {
+        return $this->lastMessage;
     }
 }
--- a/lib/Doctrine/Migrations/OutputWriter.php
+++ b/lib/Doctrine/Migrations/OutputWriter.php
@@ -18,8 +18,7 @@ class OutputWriter
     public function __construct(?callable $callback = null)
     {
         if ($callback === null) {
-
-            $callback = static function (string $message) : void {
+            $callback = function (string $message) : void {
                 $this->lastMessage = $message;
             };
         }
--- a/lib/Doctrine/Migrations/Version/Executor.php
+++ b/lib/Doctrine/Migrations/Version/Executor.php
@@ -26,7 +26,7 @@ use function ucfirst;
  *
  * @internal
  */
-final class Executor implements ExecutorInterface
+class Executor implements ExecutorInterface
 {
     /** @var Configuration */
     private $configuration;
--- a/tests/Doctrine/Migrations/Tests/Configuration/ConfigurationTest.php
+++ b/tests/Doctrine/Migrations/Tests/Configuration/ConfigurationTest.php
@@ -167,7 +167,7 @@ class ConfigurationTest extends MigrationTestCase
     {
         $connection = $this->createMock(MasterSlaveConnection::class);
 
-        $connection->expects($this->once())
+        $connection->expects(self::once())
             ->method('connect')
             ->with('master')
             ->willReturn(true);
@@ -258,7 +258,7 @@ class ConfigurationTest extends MigrationTestCase
 
         $schemaManager = $this->createMock(AbstractSchemaManager::class);
 
-        $conn->expects($this->any())
+        $conn->expects(self::any())
             ->method('getSchemaManager')
             ->willReturn($schemaManager);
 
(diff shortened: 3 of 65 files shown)


tests/Doctrine/Migrations/Tests/Version/VersionTest.php
```diff
@@ -178,10 +186,12 @@ public function testAddSql() : void
...
-        self::assertNull($version->addSql('SELECT * FROM foo'));
-        self::assertNull($version->addSql('SELECT * FROM foo'));
-        self::assertNull($version->addSql('SELECT * FROM foo WHERE id = ?', [1]));
-        self::assertNull($version->addSql('SELECT * FROM foo WHERE id = ?', [1], [PDO::PARAM_INT]));
+        $version->addSql('SELECT * FROM foo');
+        $version->addSql('SELECT * FROM foo');
+        $version->addSql('SELECT * FROM foo WHERE id = ?', [1]);
+        $version->addSql('SELECT * FROM foo WHERE id = ?', [1], [PDO::PARAM_INT]);
+
+        self::markTestIncomplete('Does not perform any real assertion.');
```
lcobucci: For these cases we should use `$this->addToAssertionCount(1)`, more info: https://github.com/sebastianbergmann/phpunit/pull/3348

lib/Doctrine/Migrations/Version/Executor.php
```diff
@@ -26,7 +26,7 @@
  *
  * @internal
  */
-final class Executor implements ExecutorInterface
+class Executor implements ExecutorInterface
```
lcobucci: We shouldn't remove `final` here

tests/Doctrine/Migrations/Tests/Version/VersionTest.php
```diff
@@ -180,18 +180,15 @@ public function testAddSql() : void
     {
         $configuration = $this->getSqliteConfiguration();
 
-        $version = $this->createTestVersion(
-            $configuration,
-            '003',
-            VersionDummy::class
-        );
+        $versionExecutor = $this->createMock(Executor::class);
```
lcobucci: We should mock the interface instead

lib/Doctrine/Migrations/OutputWriter.php
```diff
@@ -12,10 +12,15 @@ class OutputWriter
...
 
+    /** @var string */
+    private $lastMessage = '';
+
     public function __construct(?callable $callback = null)
     {
         if ($callback === null) {
-            $callback = static function ($message) : void {
+
+            $callback = static function (string $message) : void {
```
lcobucci: Static closures don't get any object bound to `$this`, we should then change this to:

```suggestion
            $callback = function (string $message) : void {
```

tests/Doctrine/Migrations/Tests/Configuration/ConfigurationTest.php
```diff
@@ -122,7 +133,10 @@ public function testVersionsTryToGetLoadedIfNotAlreadyLoadedWhenAccessingMethodT
         );
         $migrator->migrate('3Test');
 
-        $result = call_user_func_array([$configuration, $method], $args);
+        /** @var callable $callable */
+        $callable = [$configuration, $method];
+
+        $result = call_user_func_array($callable, $args);
```
lcobucci: I think this would be better (readability and performance) and would allow us to remove that `/** @var callable $callable */` which is never reliable:

```php
$result = $configuration->$method(...$args);
```

Majkl578:
```php
[$configuration, $method](...$args)
```
This should work too.

tests/Doctrine/Migrations/Tests/Functional/FunctionalTest.php
```diff
@@ -524,7 +524,7 @@ public function testMigrationExecutedAt() : void
     }
 
     /**
-     * @return Connection|Configuration[]
+     * @return object[] [Connection, Configuration]
```
lcobucci: Is this correct? Didn't see this syntax before :stuck_out_tongue:

tests/Doctrine/Migrations/Tests/OutputWriterTest.php
```diff
@@ -19,7 +19,9 @@ public function testDefaultCallback() : void
     {
         $outputWriter = new OutputWriter();
 
-        self::assertNull($outputWriter->write('test message'));
+        $outputWriter->write('test message');
+
+        self::markTestIncomplete('Does not perform any real assertion.');
```
lcobucci: `$this->addToAssertionCount(1)`, ref: https://github.com/sebastianbergmann/phpunit/pull/3348

lcobucci: Nevermind, outdated

tests/Doctrine/Migrations/Tests/Functional/FunctionalTest.php
Majkl578: Yeah, looks weird, imho has no meaning in PHPStan context.

jwage: I am not familiar with this either. I didn't make the change. It was in the first commits by @Majkl578 https://github.com/doctrine/migrations/pull/765/commits/bf980f3dc2ff16badb7fda9cb03a3b7442936e87

jwage: I think it should be this:

```
@return Connection[]|Configuration[]
```

Majkl578: Afaik it was intended to indicate a tuple consisting of Connection and Configuration, not array of those.

jwage: phpstan didn't complain about the syntax. Should I change it back?

PR #773, opened 2019-01-02
Change the column length of the version column to default to 14

### LINKED ISSUE #772: Change default migrations table version column length from 255 to 14
Related https://github.com/doctrine/DoctrineMigrationsBundle/issues/247

jwage:
fixes #772

The version column by default is only 14 characters so we don't need to create a varchar(255) by default. This also causes issues with MariaDB see https://github.com/doctrine/migrations/issues/514 and https://github.com/doctrine/migrations/pull/538 and https://github.com/doctrine/DoctrineMigrationsBundle/issues/247

### DIFF
--- a/lib/Doctrine/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/Migrations/Configuration/Configuration.php
@@ -21,6 +21,7 @@ use Doctrine\Migrations\OutputWriter;
 use Doctrine\Migrations\QueryWriter;
 use Doctrine\Migrations\Version\Version;
 use function str_replace;
+use function strlen;
 
 /**
  * The Configuration class is responsible for defining migration configuration information.
@@ -43,7 +44,7 @@ class Configuration
     private $migrationsColumnName = 'version';
 
     /** @var int */
-    private $migrationsColumnLength = 255;
+    private $migrationsColumnLength;
 
     /** @var string */
     private $migrationsExecutedAtColumnName = 'executed_at';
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Migrations/Configuration/Configuration.php
```diff
@@ -43,7 +43,7 @@ class Configuration
     private $migrationsColumnName = 'version';
 
     /** @var int */
-    private $migrationsColumnLength = 255;
+    private $migrationsColumnLength = 14;
```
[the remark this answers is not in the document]
jwage: What is your reasoning for wanting to do it that way?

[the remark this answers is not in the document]
greg0ire: Magic numbers are code smells I try to avoid by always showing what they represent, be it by using a constant, or showing what computation gave that result. For instance, I would use  `60 * 60 * 24` *  instead of `86400`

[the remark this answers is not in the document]
jwage: Ok, I made the change.

lib/Doctrine/Migrations/Configuration/Configuration.php
```diff
@@ -91,11 +92,12 @@ public function __construct(
...
-        $this->outputWriter      = $outputWriter;
-        $this->migrationFinder   = $migrationFinder;
-        $this->queryWriter       = $queryWriter;
-        $this->dependencyFactory = $dependencyFactory;
+        $this->connection             = $connection;
+        $this->outputWriter           = $outputWriter;
+        $this->migrationFinder        = $migrationFinder;
+        $this->queryWriter            = $queryWriter;
+        $this->dependencyFactory      = $dependencyFactory;
+        $this->migrationsColumnLength = strlen((new DateTime())->format(self::VERSION_FORMAT));
```
Majkl578: Please use DateTimeImmutable.

lib/Doctrine/Migrations/Configuration/Configuration.php
```diff
@@ -318,7 +320,7 @@ public function createMigrationTable() : bool
     public function getDateTime(string $version) : string
     {
         $datetime = str_replace('Version', '', $version);
-        $datetime = DateTime::createFromFormat('YmdHis', $datetime);
+        $datetime = DateTime::createFromFormat(self::VERSION_FORMAT, $datetime);
```
Majkl578: Please use DateTimeImmutable.

jwage: Should I be making unrelated changes in the same PR? I thought about this when I started but decided not to do it in this PR.

Majkl578: Up to you here, but the other occurence is new code.

jwage: I kept it the same as everything else so it is consistent in this sha. I will send another PR to change it everywhere.

jwage: Created an issue for this here https://github.com/doctrine/migrations/issues/774

PR #775, opened 2019-01-03
Use DateTimeImmutable instead of DateTime.

jwage:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #774

#### Summary

Related to conversation here https://github.com/doctrine/migrations/pull/773#discussion_r244876039

### DIFF
--- a/lib/Doctrine/Migrations/FileQueryWriter.php
+++ b/lib/Doctrine/Migrations/FileQueryWriter.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Migrations;
 
-use DateTime;
+use DateTimeImmutable;
 use DateTimeInterface;
 use Doctrine\Migrations\Generator\FileBuilder;
 use function file_put_contents;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Migrations/FileQueryWriter.php
```diff
@@ -41,7 +41,7 @@ public function write(
         array $queriesByVersion,
         ?DateTimeInterface $now = null
     ) : bool {
-        $now = $now ?? new DateTime();
+        $now = $now ?? new DateTimeImmutable();
```
SenseException: Does it make sense to use `new DateTimeImmutable('now', new DateTimeZone('UTC'));` like above in Configuration.php? Can a timezone with daylight saving time an issue here?

jwage: Probably should be changed to match elsewhere. Practically I don't think it would ever cause an issue though.

SenseException: I think it needs good timing. But this is for another time and PR to find out.

jwage: I will send another PR to fix that.

jwage: Issue logged here https://github.com/doctrine/migrations/issues/776

PR #787, opened 2019-02-01
upgrade guide: document bc-break with exception classes rename

sserbin:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

document bc-break with exception classes rename


jwage:
Thanks!

PR #790, opened 2019-02-06
Update docs to reflect the deafult value of column_length is 14

ntzm:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues |


jwage:
Thanks!

PR #796, opened 2019-03-13
Support multiple versions in ExecuteCommand

enumag:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | 

#### Summary

I have a use-case where I need to run many migration versions at once but need to specify them all manually. I can't simply run all until the last migration in the list because some existing migrations might be missing in the list I need to run.

Running them all using the execute command one-by-one is slow because of the overhead of running the command so I'm using a custom command to run multiple specific versions at once. With this PR I could get rid of the customized command in the future.

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
@@ -23,12 +23,12 @@ class ExecuteCommand extends AbstractCommand
             ->setName('migrations:execute')
             ->setAliases(['execute'])
             ->setDescription(
-                'Execute a single migration version up or down manually.'
+                'Execute given migration versions up or down manually.'
             )
             ->addArgument(
-                'version',
-                InputArgument::REQUIRED,
-                'The version to execute.',
+                'versions',
+                InputArgument::IS_ARRAY,
+                'The versions to execute.',
                 null
             )
             ->addOption(
@@ -63,9 +63,13 @@ class ExecuteCommand extends AbstractCommand
                 'Time all the queries individually.'
             )
             ->setHelp(<<<EOT
-The <info>%command.name%</info> command executes a single migration version up or down manually:
+The <info>%command.name%</info> command executes given migration versions up or down manually:
 
     <info>%command.full_name% YYYYMMDDHHMMSS</info>
+    
+You can execute multiple versions at once:
+
+    <info>%command.full_name% YYYYMMDDHHMMSS YYYYMMDDHHMMSS YYYYMMDDHHMMSS ...</info>
 
 If no <comment>--up</comment> or <comment>--down</comment> option is specified it defaults to up:
 
@@ -90,7 +94,7 @@ EOT
 
     public function execute(InputInterface $input, OutputInterface $output) : ?int
     {
-        $version        = $input->getArgument('version');
+        $versions       = $input->getArgument('versions');
         $timeAllQueries = (bool) $input->getOption('query-time');
         $dryRun         = (bool) $input->getOption('dry-run');
         $path           = $input->getOption('write-sql');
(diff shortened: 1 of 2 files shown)


enumag:
Note: need to add a test case for the new feature into ExecuteCommandTest. For now please give me feedback if this feature is wanted.

jwage:
I would accept this enhancement for 2.1 if it remains BC.

alcaeus:
> I would accept this enhancement for 2.1 if it remains BC.

I second that. The preferred course of action would be to:
* Add a new `versions` argument that can accept multiple versions
* Deprecate the `version` argument for removal in 3.0
* Ensure that calls that currently work will still work after this change

enumag:
I can do that. The only issue is that for my own needs accepting multiple versions as arguments turned out to not be enough because my list can be quite long and could potentially exceed the maximum command length. I solved that by allowing passing the list of versions to STDIN (one version per line) as well which is used if no versions are provided as arguments. Would that be acceptable here?

alcaeus:
I think if we get to that point, we should probably think about other ways to organise migrations: having a list of migrations that exceeds the maximum command length is way less than optimal, regardless of whether the command can take `STDIN` all alternative or not.

enumag:
> having a list of migrations that exceeds the maximum command length is way less than optimal

Why? It seems perfectly fine to me. I can't really think of a way to organize migrations that would help me with my use-case either.

Basically I have database backup I need to restore, each table as one CSV file. So I first need to run all the migrations from `_db_version.csv`, then restore the data and then run the rest of the migrations.

I can't simply run everything until the last version in `_db_version.csv` because of stable / beta / dev branching. It would be rare but occasionally there can be a backup from beta that didn't have certain migration from stable applied to it yet but the migration from stable can be older than the last one from beta. In which case I should only run this migration after restoring the data, not before.

Just running the list of migrations here is alright in my opinion. I just pipe the `_db_versions.csv` into the ExecuteCommand and get exactly the same database structure as I had when the backup was generated. Then I just apply any migrations which didn't exist at the time and do some magic to fix sequence numbers.

alcaeus:
And that's what I mean with "think about other ways to organise migrations": you basically have two sets of migrations that you need to run: one set that runs before database import, one set that runs after. I don't want to suggest manually keeping track of which versions to run at which point as a viable solution to the problem.

enumag:
> you basically have two sets of migrations that you need to run: one set that runs before database import, one set that runs after

Which migration belongs to which set is different for every backup - depending on which version of the application the backup comes from. It's not something I can solve by reorganizing the migrations.

lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
```diff
@@ -90,20 +94,27 @@ protected function configure() : void
...
         $dryRun         = (bool) $input->getOption('dry-run');
         $path           = $input->getOption('write-sql');
         $direction      = $input->getOption('down') !== false
             ? Direction::DOWN
             : Direction::UP;
 
-        $version = $this->migrationRepository->getVersion($version);
+        $versions = array_map(
+            function (string $version) {
+                return $this->migrationRepository-->getVersion($version);
```
lcobucci:
```suggestion
                return $this->migrationRepository->getVersion($version);
```

lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
```diff
@@ -90,20 +94,27 @@ protected function configure() : void
...
         $timeAllQueries = (bool) $input->getOption('query-time');
         $dryRun         = (bool) $input->getOption('dry-run');
         $path           = $input->getOption('write-sql');
         $direction      = $input->getOption('down') !== false
             ? Direction::DOWN
             : Direction::UP;
 
-        $version = $this->migrationRepository->getVersion($version);
+        $versions = array_map(
+            function (string $version) {
```
greg0ire: Please add a return type hint

SenseException: And making it `static` would be nice too :-)

tests/Doctrine/Migrations/Tests/Tools/Console/Command/ExecuteCommandTest.php
```diff
@@ -31,7 +31,7 @@ public function testWriteSqlCustomPath() : void
 
         $input->expects(self::once())
             ->method('getArgument')
-            ->with('version')
+            ->with('versions')
```
localheinz: What do you think about adding tests that demonstrate that this actually works with

- one specified version
- multiple specified versions

?

goetas:
done in https://github.com/doctrine/migrations/pull/891

PR #797, opened 2019-03-17
Update to doctrine/coding-standard 6.0

patrickjahns:
|      Q       |   A
|------------- | -----------
| Type         |  improvement
| BC Break     | no
| Fixed issues | -

#### Summary
 Updated the coding-standard package in composer.json and used `phpcbf` to fix the errors automatically

### DIFF
--- a/lib/Doctrine/Migrations/Configuration/AbstractFileConfiguration.php
+++ b/lib/Doctrine/Migrations/Configuration/AbstractFileConfiguration.php
@@ -141,9 +141,9 @@ abstract class AbstractFileConfiguration extends Configuration
 
     private function setMigrationOrganization(string $migrationOrganization) : void
     {
-        if (strcasecmp($migrationOrganization, static::VERSIONS_ORGANIZATION_BY_YEAR) === 0) {
+        if (strcasecmp($migrationOrganization, self::VERSIONS_ORGANIZATION_BY_YEAR) === 0) {
             $this->setMigrationsAreOrganizedByYear();
-        } elseif (strcasecmp($migrationOrganization, static::VERSIONS_ORGANIZATION_BY_YEAR_AND_MONTH) === 0) {
+        } elseif (strcasecmp($migrationOrganization, self::VERSIONS_ORGANIZATION_BY_YEAR_AND_MONTH) === 0) {
             $this->setMigrationsAreOrganizedByYearAndMonth();
         } else {
             throw UnknownConfigurationValue::new('organize_migrations', $migrationOrganization);
--- a/lib/Doctrine/Migrations/Provider/LazySchemaDiffProvider.php
+++ b/lib/Doctrine/Migrations/Provider/LazySchemaDiffProvider.php
@@ -49,7 +49,8 @@ class LazySchemaDiffProvider implements SchemaDiffProviderInterface
         return $this->proxyFactory->createProxy(
             Schema::class,
             static function (&$wrappedObject, $proxy, $method, array $parameters, &$initializer) use ($originalSchemaManipulator) {
-                $initializer   = null;
+                $initializer = null;
+
                 $wrappedObject = $originalSchemaManipulator->createFromSchema();
 
                 return true;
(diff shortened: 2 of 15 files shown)


lib/Doctrine/Migrations/Provider/LazySchemaDiffProvider.php
```diff
@@ -49,7 +49,8 @@ public function createFromSchema() : Schema
         return $this->proxyFactory->createProxy(
             Schema::class,
             static function (&$wrappedObject, $proxy, $method, array $parameters, &$initializer) use ($originalSchemaManipulator) {
-                $initializer   = null;
+                $initializer = null;
+
```
patrickjahns:
```
 49 | ERROR | [x] Expected 1 lines after "return", found 0.
    |       |     (SlevomatCodingStandard.ControlStructures.ControlStructureSpacing.IncorrectLinesCountAfterControlStructure)
```

Is also causing trouble here

lib/Doctrine/Migrations/Configuration/AbstractFileConfiguration.php
```diff
@@ -141,9 +141,9 @@ private function loadMigrations(array $migrations) : void
 
     private function setMigrationOrganization(string $migrationOrganization) : void
     {
-        if (strcasecmp($migrationOrganization, static::VERSIONS_ORGANIZATION_BY_YEAR) === 0) {
+        if (strcasecmp($migrationOrganization, self::VERSIONS_ORGANIZATION_BY_YEAR) === 0) {
```
patrickjahns:
```
  97 | ERROR | [x] Late static binding for constants is disallowed.
     |       |     (SlevomatCodingStandard.Classes.DisallowLateStaticBindingForConstants.DisallowedLateStaticBindingForConstant)
```

🤔 - thinking of it, this is not so innocent. Should we force the change here or should we rather exclude the rule for this file?

jwage: What is the issue? The change seems fine unless I'm missing something.

patrickjahns: Due to the fact it's a abstract class and intended to be extended, when switching from `static::VERSIONS_ORGANIZATION_BY_YEAR ` to `self::VERSIONS_ORGANIZATION_BY_YEAR ` we change behavior here.

With `static::VERSIONS_ORGANIZATION_BY_YEAR ` the constant from the subClass would be used - whereas with `self::VERSIONS_ORGANIZATION_BY_YEAR ` - the constant from `AbstractFileConfiguration` would always be used ( from my understanding ).

Please correct me if I am wrong

jwage: I understand what you mean now. I don't think it matters in practice. These classes are internal and not intended for the end user to extend so the constant should not ever be overridden.

lib/Doctrine/Migrations/Provider/LazySchemaDiffProvider.php
jwage: Can this not be fixed?

patrickjahns: I apologize - should have been more verbose on what the issue actually is. 
Identified the following bug/misbehavior while rolling out the new coding-standard - please see: https://github.com/slevomat/coding-standard/issues/635

Would need to fix it upstream to not receive a false positive here as well. The other option would be to exclude this sniff for the time being. When fixed upstream we can reenable the sniff again

jwage: No worries. Thanks for explaining. Should we disable the rule for now?

patrickjahns: I am fine with it - however in other repositories in the doctrine project "we" decided to merge it with this change and let the rule stay for now.

Either way is fine for me and I prepare the change 👍

jwage: I'm fine with that as long as the build is green.

PR #799, opened 2019-03-20
handle unregistered migrations in up-to-date check

dbu:
|      Q       |   A
|------------- | -----------
| Type         | bug & improvement
| BC Break     | no
| Fixed issues | -

#### Summary

We need to handle the case where there are extra migrations in doctrine. without this change, the command outputs something like `"Out-of-date! 18446744073709551613 migration is available to execute."` we convert the signed integer value to unsigned.

with this fix, we explicitly handle the case. the default is to report the issue but not fail, same as the migrate command that only outputs a warning but does not refuse to run. i also add a flag in case people want a strict check.

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/UpToDateCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/UpToDateCommand.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\Migrations\Tools\Console\Command;
 
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use function count;
 use function sprintf;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Migrations/Tools/Console/Command/UpToDateCommand.php
```diff
@@ -43,12 +45,26 @@ public function execute(InputInterface $input, OutputInterface $output) : ?int
...
-            $availableMigrations,
-            $availableMigrations > 1 ? 's are' : ' is'
+            '<error>You have %1$u previously executed migration%3$s in the database that %2$s registered migration%3$s.</error>',
+            $extraMigrations,
+            $extraMigrations > 1 ? 'are not' : 'is not a',
+            $extraMigrations > 1 ? 's' : ''
         ));
 
-        return 1;
+        return true === $input->getOption('fail-on-unregistered') ? 2 : 0;
```
dbu: i originally had `$input->getOption('fail-on-unregistered') ? 2 : 0` but then the code sniffer complained that the return value of getOption is mixed. i tried to fix it like this, but now the code style check complains about yoda style comparison. how should i do the comparison?

SenseException: `$input->getOption('fail-on-unregistered') === true`

dbu:
looks like scrutinizer failed talking to github or something.

is that ok as is, should i adjust anything with the behaviour?

jwage:
Usually we just restart the build and try it again.

dbu:
i don't think i have permission to do that.

content-wise, do you need anything more to have this mergeable? changelog entry maybe? and do you agree with the actual feature addition?

PR #801, opened 2019-03-21
Commands: allow to reset migrationConfiguration property and reuse the same Command

sergey-shambir:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

After pull request #219 the method setMigrationConfiguration changes nothing when migrationConfiguration was initialized
So we can't reset migrationConfiguration after first run in the same process
This change allows to re-use migration commands whith another configuration on the same PHP process


sergey-shambir:
This change targeted for 1.8 (since we're using this version in our products), but applicable to 2.0 (with minor merge conflicts).

sergey-shambir:
There are too much other issues when reusing MigrationsMigrateDoctrineCommand, so it's better to restart PHP script each time.

PR #802, opened 2019-03-27
don't run with 64-bit data on 32-bit build

remicollet:
ON 32-bit computer

```
There was 1 error:
1) Doctrine\Migrations\Tests\Tools\BytesFormatterTest::testFormatBytes
TypeError: Argument 1 passed to Doctrine\Migrations\Tools\BytesFormatter::formatBytes() must be of the type int, float given, called in /builddir/build/BUILD/migrations-0101f5bd7f4e5043bf8630db2930f8fd7da552b6/tests/Doctrine/Migrations/Tests/Tools/BytesFormatterTest.php on line 17
/builddir/build/BUILDROOT/php-doctrine-migrations-2.0.0-1.fc31.noarch/usr/share/php/Doctrine/Migrations/Tools/BytesFormatter.php:20
/builddir/build/BUILD/migrations-0101f5bd7f4e5043bf8630db2930f8fd7da552b6/tests/Doctrine/Migrations/Tests/Tools/BytesFormatterTest.php:17

```

PR #803, opened 2019-03-27
allow "float" in formatBytes to work on 32-bit computers

remicollet:
Alternative to pr #802 which could be better.

As float also allow int, I don't think this is a BC break, especially as the class is final


jwage:
Thanks @remicollet for this PR and everything else you do! I will merge this manually in to the 2.0 branch and release 2.0.1

Issue #795, opened 2019-03-08
Handle deleted version classes


jwage:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

If a *Version.php class is deleted and that version has already been migrated and exists in the database, the `php bin/console doctrine:migrations:version 20190226081917 --delete` command will fail because the version class does not exist.

We need a way to cleanup orphaned versions. See https://github.com/doctrine/DoctrineMigrationsBundle/issues/259

PR #807, opened 2019-04-05
Handle deleted migration classes #795

jschaedl:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #795 

#### Summary

With the current implementation adding and deleting migration versions is strongly coupled to an existing migration class file and there is no easy way to decouple this dependency. (at least I didn't see one :-))  

So when a User is trying to delete an already migrated version with a missing migration class file, I would  suggest to give him more useful information and to ask him, if he would like to create an empty migration file so that he can proceed with the deletion.

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/VersionCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/VersionCommand.php
@@ -146,7 +146,7 @@ EOT
             $availableVersions = $this->migrationRepository->getAvailableVersions();
 
             foreach ($availableVersions as $version) {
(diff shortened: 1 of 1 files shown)


jwage:
Thanks for taking a stab at fixing this. However, I don't think this is the right approach. I think we just need a way to directly delete the version from the table without loading the migration version class. Then we have a clean way to handle this scenario.

jschaedl:
@jwage This would be indeed possible. What do you think: https://github.com/doctrine/migrations/pull/808

jwage:
Closing in favor of #808

PR #808, opened 2019-04-05
Handle deleted migration classes #795

jschaedl:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #795

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/VersionCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/VersionCommand.php
@@ -146,7 +146,7 @@ EOT
             $availableVersions = $this->migrationRepository->getAvailableVersions();
 
             foreach ($availableVersions as $version) {
-                $this->mark($output, $version, true);
+                $this->mark($input, $output, $version, true);
             }
         } elseif ($rangeFromOption !== null && $rangeToOption !== null) {
             $availableVersions = $this->migrationRepository->getAvailableVersions();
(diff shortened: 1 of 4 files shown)


jschaedl:
> Looks good. Can you move that delete functionality to a service and add a unit test for it?

Sure I can do that. But not sure if we need a whole new Service class like `MigrationVersionRemover` 🤔

As an alternative we could add a new method to the `MigrationRepository` and name it `deleteMigrationVersionFromDatabase(string $version): void`.

WDYT?

jwage:
MigrationRepository sounds good.

lib/Doctrine/Migrations/Tools/Console/Command/VersionCommand.php
```diff
@@ -168,10 +168,29 @@ private function markVersions(InputInterface $input, OutputInterface $output) :
      * @throws VersionDoesNotExist
      * @throws UnknownMigrationVersion
      */
-    private function mark(OutputInterface $output, string $version, bool $all = false) : void
+    private function mark(InputInterface $input, OutputInterface $output, string $version, bool $all = false) : void
     {
         if (! $this->migrationRepository->hasVersion($version)) {
-            throw UnknownMigrationVersion::new($version);
+            if ((bool) $input->getOption('delete') === false) {
```
jwage: I'm not sure if VersionCommand already has a test or not but we need to test this functionality/logic here too.

jwage:
Looks good. Can you squash? and then I think we are ready to merge.

jschaedl:
@jwage I squashed, so ready to merge :-)

jwage:
Looks like CS is failing. Run vendor/bin/phpcbf to fix.

jschaedl:
@jwage I think we are ready to merge. :-)

jwage:
Thanks for your work on this!

Issue #806, opened 2019-04-01
Migrations with --write-sql does not respect tables that are not created in database


raph1mm:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.0.0

#### Summary

In case when there are two migrations, which first creates table and the second one - changes it - using --dry-run crashes migrations.

#### Current behavior

Migrations cannot "see" tables that are created only "virtually" during dry run. There is error:

> There is no table with name 'my_db.my_table' in the schema.

#### How to reproduce

1. Create first migration with up() method:
`$schema->createTable('my_table');`
2. Create second migration with up() method:
`$schema->getTable('my_table');`
3. Run 
`bin/console doctrine:migrations:migrate --write-sql='./my.sql'`

#### Expected behavior

The expected behavior would be sql script with dumped all migrations - without any errors.

PR #809, opened 2019-04-08
Fix for referencing tables created during a dry run

jwage:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #806 

#### Summary

When you execute a dry run of a migration, you cannot reference tables created in earlier migrations in later migrations due to the from Schema always being fetched from the current state of the database. Since it is a dry run, nothing is committed to the database. The fix is to use the same `Schema` object throughout the whole chain of migrations when executing a dry run.

Added a test to cover this scenario that fails before the fix:

```
1) Doctrine\Migrations\Tests\Functional\FunctionalTest::testDryRunWithTableCreatedWithSchemaInFirstMigration
Doctrine\DBAL\Schema\SchemaException: There is no table with name 'public.foo' in the schema.
/home/travis/build/doctrine/migrations/vendor/doctrine/dbal/lib/Doctrine/DBAL/Schema/SchemaException.php:30
/home/travis/build/doctrine/migrations/vendor/doctrine/dbal/lib/Doctrine/DBAL/Schema/Schema.php:168
/home/travis/build/doctrine/migrations/tests/Doctrine/Migrations/Tests/Stub/Functional/DryRun/DryRun2.php:14
/home/travis/build/doctrine/migrations/lib/Doctrine/Migrations/Version/Executor.php:209
/home/travis/build/doctrine/migrations/lib/Doctrine/Migrations/Version/Executor.php:131
/home/travis/build/doctrine/migrations/lib/Doctrine/Migrations/Version/Version.php:187
/home/travis/build/doctrine/migrations/lib/Doctrine/Migrations/Migrator.php:180
/home/travis/build/doctrine/migrations/lib/Doctrine/Migrations/Migrator.php:144
/home/travis/build/doctrine/migrations/tests/Doctrine/Migrations/Tests/Functional/FunctionalTest.php:201
```

### DIFF
--- a/lib/Doctrine/Migrations/Migrator.php
+++ b/lib/Doctrine/Migrations/Migrator.php
@@ -179,6 +179,13 @@ class Migrator
             foreach ($migrationsToExecute as $version) {
                 $versionExecutionResult = $version->execute($direction, $migratorConfiguration);
 
+                // capture the to Schema for the migration so we have the ability to use
+                // it as the from Schema for the next migration when we are running a dry run
+                // $toSchema may be null in the case of skipped migrations
+                if (! $versionExecutionResult->isSkipped()) {
+                    $migratorConfiguration->setFromSchema($versionExecutionResult->getToSchema());
+                }
+
                 $sql[$version->getVersion()] = $versionExecutionResult->getSql();
                 $time                       += $versionExecutionResult->getTime();
             }
--- a/lib/Doctrine/Migrations/MigratorConfiguration.php
+++ b/lib/Doctrine/Migrations/MigratorConfiguration.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace Doctrine\Migrations;
 
+use Doctrine\DBAL\Schema\Schema;
+
 /**
  * The MigratorConfiguration class is responsible for defining the configuration for a migration.
  *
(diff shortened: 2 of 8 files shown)


lib/Doctrine/Migrations/Migrator.php
```diff
@@ -179,6 +179,15 @@ private function executeMigration(
             foreach ($migrationsToExecute as $version) {
                 $versionExecutionResult = $version->execute($direction, $migratorConfiguration);
 
+                $toSchema = $versionExecutionResult->getToSchema();
+
+                // capture the to Schema for the migration so we have the ability to use
+                // it as the from Schema for the next migration when we are running a dry run
+                // $toSchema may be null in the case of skipped migrations
+                if ($toSchema !== null) {
```
SenseException: Can you use

``` php
if (! $versionExecutionResult->isSkipped()) {
        $toSchema = $versionExecutionResult->getToSchema();
        $migratorConfiguration->setFromSchema($toSchema);
}
```

here?

jwage: I could, but then I think PHPStan will still complain because getToSchema() return type hint can be null. I could remove the optional return type hint and throw an exception when the getter is called and toSchema is null. This should only happen when isSkipped is true.

jwage: This is fixed now.

PR #819, opened 2019-05-10
Generator Template: Accept Properties

Finleth:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

My company and I are currently implementing this doctrine migrations library and have set it up such that we can pass a custom option to our custom generate command to use different default <up> and <down> content based off of the parameter given. However along these same lines we've found the need to pass in along default properties to the generated migration version.

This could be worked around (at my work I can just write my own Generator.php file that my custom generator command uses for now) but I thought this would be fairly simple to implement and could be beneficial for anybody else looking to use their own customizable templates.

### DIFF
--- a/lib/Doctrine/Migrations/Generator/Generator.php
+++ b/lib/Doctrine/Migrations/Generator/Generator.php
@@ -39,6 +39,8 @@ use Doctrine\Migrations\AbstractMigration;
  */
 final class Version<version> extends AbstractMigration
 {
+<properties>
+
     public function getDescription() : string
     {
         return '';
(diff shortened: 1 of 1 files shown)


jwage:
This class is marked as internal. I am not sure how I feel about opening this up at the moment.

Finleth:
Would there be any other options for setting up default generated migrations with properties that can, through the use of either a custom command or any other command line options, take in a value for the properties?

PR #823, opened 2019-05-17
Fix DiffGenerator

enumag:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/DoctrineBundle/issues/971

#### Summary

The `getFilterSchemaAssetsExpression` method is deprecated and with DoctrineBundle 1.11 it no longer returns the filter used in Symfony configuration. Because of that the schema filter was not applied and migrations generated extra tables in our system.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
     ],
     "require": {
         "php": "^7.1",
-        "doctrine/dbal": "^2.6",
+        "doctrine/dbal": "^2.9",
         "ocramius/package-versions": "^1.3",
         "ocramius/proxy-manager": "^2.0.2",
         "symfony/console": "^3.4||^4.0",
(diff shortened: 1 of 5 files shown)


enumag:
Btw. any reason for having a lock file commited in the repo? Afaik it's not considered good practice for libraries.

weaverryan:
@enumag The Doctrine team decided to commit the lock file. There are reasons for it - so it's not by accident :)

enumag:
@weaverryan Obviously. What are the reasons?

jwage:
@enumag We have some information about that here https://www.doctrine-project.org/policies.html

enumag:
@jwage I don't necessarily agree with the policy since you won't get notified about test failures caused by newer versions of other libraries. Tools like phpstan could be version-locked in the composer.json instead. But anyway thanks for the link.

bendavies:
The same issue exists in doctrine-fixtures, FYI.

enumag:
So... can anyone do a review for me here?

jwage:
Thanks!

PR #824, opened 2019-05-22
return annotation for Configuration::getMigratedVersions is fixed

boris-brtan:
- https://github.com/doctrine/migrations/blob/master/lib/Doctrine/Migrations/Configuration/Configuration.php#L414

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | in 1.x version there is wrong return type annotation which can mislead developers. In last 2.x version this is fixed.

#### Summary

return type annotation for Configuration::getMigratedVersions was corrected from `Version[]` into `string[]`

### DIFF
--- a/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/DBAL/Migrations/Configuration/Configuration.php
@@ -544,7 +544,7 @@ class Configuration
     /**
      * Returns all migrated versions from the versions table, in an array.
      *
-     * @return Version[]
+     * @return string[]
      */
     public function getMigratedVersions()
     {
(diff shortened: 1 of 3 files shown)


jwage:
I don't currently have any plans to release a new 1.8.x version except for security fixes.

boris-brtan:
OK then, i will try to find some security issues.

SenseException:
@jwage Should we mention this in the contribution guide about 1.8 only be open for security fixes? It would make sense.

jwage:
@SenseException sure

jwage:
Thanks @boris-brtan !

Issue #788, opened 2019-02-03
Updating migrations table using Oracle fails


bobvandevijver:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.0

#### Summary

The new `executed_at` column as introduced in !675 (awesome feature btw!) unfortunately breaks migrations on Oracle databases as the SQL statement generated to update the migrations table is incorrect. 
```
ALTER TABLE MIGRATION_VERSIONS ADD (version VARCHAR2(14) NOT NULL, executed_at TIMESTAMP(0) DEFAULT NULL NULL)
```

As you can see, it tries to re-add the already existing table (which is missing the `executed_at` column), which obviously then fails with:

```
ORA-01430: column being added already exists in table
```

Note we've also just updated several MariaDB and Postgres servers without any issue, so it seems that only Oracle is affected.

#### Current behavior

Migrations are broken when using an Oracle database (until manual intervention)

#### How to reproduce

Remove the `executed_at` table if you have one, and run (using the Symfony migrations bundle for integration):

```
php bin/console doctrine:migrations:status
```

#### Expected behavior

No errors 😄

jwage:
Why does that column already exist? Did you add it manually. I don't have a way to test against Oracle easily. Can you help triage the problem?

bobvandevijver:
I'm afraid you read the message wrong: Oracle warns about the `version` column already being there (which is and should be, as we upgraded from 1.8). The generated SQL to add the `executed_at` column also tries to add the `version` column, which is indeed invalid.

I can test, but I unfortunately do not have much time to spare to find the actual cause.

SenseException:
@bobvandevijver Which version of Oracle DB is in use?

bobvandevijver:
I've now seen on these two versions:
* Oracle Database 12c Enterprise Edition Release 12.1.0.2.0 - 64bit Production
* Oracle Database 11g Express Edition Release 11.2.0.2.0 - 64bit Production

jwage:
If you manually add the new columns to the table, does Doctrine stop trying to update the table?

bobvandevijver:
Yes, if I manually add the `executed_at` column it stops complaining and it then works as I would expect. Tried mentioning that in the original post, but that might not have been as clear as I thought :)

hmonglee:
Hello I have the same issue with Oracle. I made a PR.

PR #828, opened 2019-06-07
Fix doctrine migration schema update

hmonglee:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #788

#### Summary

Fix doctrine migration schema update when colums name is not in the same case

### DIFF
--- a/lib/Doctrine/Migrations/Tracking/TableUpdater.php
+++ b/lib/Doctrine/Migrations/Tracking/TableUpdater.php
@@ -11,6 +11,7 @@ use Doctrine\DBAL\Schema\Schema;
 use Doctrine\DBAL\Schema\Table;
 use Throwable;
 use function in_array;
+use function strtolower;
 
 /**
  * The TableUpdater class is responsible for updating the tracking table schema if it needs to be updated.
@@ -88,7 +89,7 @@ class TableUpdater
         // remove columns from the table definition that we don't care about
         // so we don't try to drop those columns
         foreach ($table->getColumns() as $column) {
-            if (in_array($column->getName(), $columnNames, true)) {
+            if (in_array(strtolower($column->getName()), $columnNames, true)) {
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Migrations/Tracking/TableUpdater.php
```diff
@@ -88,7 +89,7 @@ private function getFromTable() : Table
         // remove columns from the table definition that we don't care about
         // so we don't try to drop those columns
         foreach ($table->getColumns() as $column) {
-            if (in_array($column->getName(), $columnNames, true)) {
+            if (in_array(strtolower($column->getName()), $columnNames, true)) {
```
SenseException: How does this behave on a different naming strategy?

tests/Doctrine/Migrations/Tests/Tracking/TableUpdaterTest.php
```diff
@@ -55,7 +55,7 @@ public function testUpdateMigrationTable() : void
             ->willReturn($table);
 
         $versionColumn = new Column(
-            'version',
+            'VERSION',
```
SenseException: Please don't adapt an existing test but create a new one.

SenseException (changes requested):

lib/Doctrine/Migrations/Tracking/TableUpdater.php
hmonglee: @SenseException Maybe strtolower both side ? I know it's not the best choice. What you will recommend for that ?

SenseException: Maybe you should describe the problem first, that you want to fix, including an example. You can edit your summary above for that example and describe why names aren't in the correct case. This example may also work as a test, so it is also okay to put the example into a test, that will fail first, before it can be fixed in your PR.

alcaeus:
@hmonglee can you please explain the problem you're trying to fix? I'm not sure how we can get to this point, and as @SenseException pointed out the fix seems a bit dodgy. Thanks!

PR #830, opened 2019-06-20
[DOCS] Added use statement for `Events`

evertharmeling:
To clarify the usage example

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

In the example in the docs the 'use statement' for the used class `Events` was missing.

### DIFF
--- a/docs/en/reference/events.rst
+++ b/docs/en/reference/events.rst
@@ -20,6 +20,7 @@ listens for all possible migrations events.
     use Doctrine\Common\EventSubscriber;
     use Doctrine\Migrations\Event\MigrationsEventArgs;
     use Doctrine\Migrations\Event\MigrationsVersionEventArgs;
+    use Doctrine\Migrations\Events;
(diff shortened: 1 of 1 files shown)


greg0ire:
Link for the lazy: https://github.com/doctrine/migrations/blob/master/lib/Doctrine/Migrations/Events.php

docs/en/reference/events.rst
```diff
@@ -18,6 +18,7 @@ listens for all possible migrations events.
     <?php
 
     use Doctrine\Common\EventSubscriber;
+    use Doctrine\Migrations\Events;
```
localheinz: Should this come last, i.e., after 

```php
use Doctrine\Migrations\Event\MigrationsVersionEventArgs;
```

?

alcaeus:
Thanks @evertharmeling!

PR #839, opened 2019-07-15
Lazy console commands

vudaltsov:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

Since Symfony `3.4` console commands can be lazy thanks to [a static name property](https://github.com/symfony/symfony/pull/23887).
This change does not break BC because the BC layer is already [in the base command's constructor](https://github.com/symfony/symfony/pull/23887/files#diff-8e80595663798837db1b033187835535R75).

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
@@ -22,6 +22,9 @@ use function sprintf;
  */
 class DiffCommand extends AbstractCommand
 {
+    /** @var string */
+    protected static $defaultName = 'migrations:diff';
+
     /** @var SchemaProviderInterface|null */
     protected $schemaProvider;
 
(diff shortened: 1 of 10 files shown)


vudaltsov:
I will do this in other doctrine repos soon.

lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
```diff
@@ -22,6 +22,9 @@
  */
 class DiffCommand extends AbstractCommand
 {
+    /** @var string */
```
localheinz: Is the comment really necessary?

alcaeus: Yes: while the initial type could be inferred from the initial assignment, this isn't reliable enough to satisfy PHPStan. Thus, the type hint still needs to be included.

greg0ire:
```
Since Symfony 3.4 console commands can be lazy thanks to a static name property.
This change does not break BC because the BC layer is already in the base command's constructor.
```

This explanation is perfect, and should be included in your commit message.

alcaeus:
Thanks @vudaltsov!

Issue #833, opened 2019-06-25
Diff command throws Exception instead of message


jawira:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | v1.8

#### Summary

**In Migration v1.8, `diff command` throws Exception instead of message. This is a problem for pipelines (CI) and also for build tools (make, Phing, ...) since they rely on return values.**

This is new for me since, in older projects, `diff command` only returned a nice message (same kind of message that `migrate command` returns).

Digging into the code I found out where this change was made: https://github.com/doctrine/migrations/commit/cbe1b7f2ffc6441a151197497becb81b73b1a51d

As you can see, in this commit the following code was removed...:

```php
// lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
if ($up === '' && $down === '') {
	$output->writeln('No changes detected in your mapping information.');
	return;
}
```

and was moved as an exception:

```php
// lib/Doctrine/Migrations/MigrationDiffGenerator.php
if ($up === '' && $down === '') {
    throw new RuntimeException('No changes detected in your mapping information.');
}
```

IMHO a refactoring should not transform a message into an Exception.

Later, in commit d9c1b0cfe3e1479c28fa8264e8c41b3227d3c134 this exception was refactored as:

```php
// lib/Doctrine/Migrations/MigrationDiffGenerator.php
if ($up === '' && $down === '') {
	throw NoChangesDetected::new();
}
```

#### Previous behaviour

When there's no changes is mapping, the `diff command` displays a message as output:

```
No changes detected in your mapping information.
```

#### Current behaviour

When there's no changes is mapping, the `diff command` throws an Exception:

```
In NoChangesDetected.php line 13:
No changes detected in your mapping information.  
```                                                

#### How to reproduce

Run `diff command` when you have no changes in your mapping:

```
// I'm using symfony console
$ bin/console doctrine:migrations:diff
```

----

I though in two possible solutions to solve this:

1. Either `NoChangesDetected` exception is removed and a message is displayed (this was the behaviour before v1.8).
2. Or, it's also possible to add an option `--allow-no-diff` to mute the `NoChangesDetected` exception. Please note that `migration command` has a `--allow-no-migration` option which has a similar behaviour.

I'm more than happy to help with this if you can give some advices and tell me what to do and where to start :)

jwage:
This change was made in 2.0 right? Your version says 1.8 but I thought it was done in 2.0 refactoring.

[...]

PR #840, opened 2019-07-16
Add option allow-empty-diff for diff command

jawira:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #833

#### Summary

`--allow-no-diff` is new option for `DiffCommand`, it allow to display a message instead of an Exception when there's no changes in mapping information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ phpunit.xml
 /.phpcs-cache
 /box.phar
 /box.json
+
+# Please don't add your IDE's files here, use global gitignore instead.
--- a/lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\Migrations\Tools\Console\Command;
 
 use Doctrine\Migrations\Generator\DiffGenerator;
+use Doctrine\Migrations\Generator\Exception\NoChangesDetected;
 use Doctrine\Migrations\Provider\OrmSchemaProvider;
 use Doctrine\Migrations\Provider\SchemaProviderInterface;
 use Doctrine\Migrations\Tools\Console\Exception\InvalidOptionUsage;
@@ -81,6 +82,12 @@ EOT
                 InputOption::VALUE_OPTIONAL,
                 'Check Database Platform to the generated code.',
                 true
+            )
+            ->addOption(
+                'allow-empty-diff',
+                null,
+                InputOption::VALUE_NONE,
+                'Do not throw an exception when no changes are detected.'
             );
     }
 
@@ -94,6 +101,7 @@ EOT
         $filterExpression = $input->getOption('filter-expression') ?? null;
         $formatted        = (bool) $input->getOption('formatted');
         $lineLength       = (int) $input->getOption('line-length');
+        $allowEmptyDiff   = (bool) $input->getOption('allow-empty-diff');
         $checkDbPlatform  = filter_var($input->getOption('check-database-platform'), FILTER_VALIDATE_BOOLEAN);
 
         if ($formatted) {
@@ -106,13 +114,22 @@ EOT
 
         $versionNumber = $this->configuration->generateVersionNumber();
 
-        $path = $this->createMigrationDiffGenerator()->generate(
-            $versionNumber,
-            $filterExpression,
-            $formatted,
-            $lineLength,
-            $checkDbPlatform
-        );
+        try {
+            $path = $this->createMigrationDiffGenerator()->generate(
+                $versionNumber,
+                $filterExpression,
+                $formatted,
+                $lineLength,
+                $checkDbPlatform
+            );
+        } catch (NoChangesDetected $exception) {
+            if ($allowEmptyDiff) {
+                $output->writeln($exception->getMessage());
+
+                return 0;
+            }
+            throw $exception;
+        }
 
         $editorCommand = $input->getOption('editor-cmd');
 
--- a/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
@@ -59,7 +59,7 @@ class MigrateCommand extends AbstractCommand
                 'allow-no-migration',
                 null,
                 InputOption::VALUE_NONE,
-                'Don\'t throw an exception if no migration is available (CI).'
+                'Do not throw an exception if no migration is available.'
             )
             ->addOption(
                 'all-or-nothing',
--- a/tests/Doctrine/Migrations/Tests/Tools/Console/Command/DiffCommandTest.php
+++ b/tests/Doctrine/Migrations/Tests/Tools/Console/Command/DiffCommandTest.php
@@ -49,10 +49,15 @@ final class DiffCommandTest extends TestCase
 
         $input->expects(self::at(3))
             ->method('getOption')
-            ->with('check-database-platform')
+            ->with('allow-empty-diff')
             ->willReturn(true);
 
         $input->expects(self::at(4))
+            ->method('getOption')
+            ->with('check-database-platform')
+            ->willReturn(true);
+
+        $input->expects(self::at(5))
             ->method('getOption')
             ->with('editor-cmd')
             ->willReturn('mate');


.gitignore
```diff
@@ -11,3 +11,4 @@ phpunit.xml
 /.phpcs-cache
 /box.phar
 /box.json
+/.idea
```
lcobucci: This should be present in your global gitignore. Different people might use different tools and we definitely don't want to add all possible configuration files here.

lib/Doctrine/Migrations/Generator/DiffGenerator.php
```diff
@@ -98,7 +94,7 @@ static function ($assetName) use ($filterExpression) {
         );
 
         if ($up === '' && $down === '') {
-            throw NoChangesDetected::new();
```
lcobucci: We shouldn't change this, a try/catch block can be added to the command. That isolates the change and doesn't introduce a new type in the signature of the method (null).

.gitignore
stof: see https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore for the doc about that.

@lcobucci I suggest you to include this link in your answers about that point in the future. It makes it easier for people not experienced with git to solve it :smile:

alcaeus:
I agree with @lcobucci here: I don't think we want to remove the exception from `DiffGenerator`, nor do we want to add a command line switch to control behaviour. Instead, the `DiffCommand` should handle the exception gracefully:
* catch it,
* let the user know that there is no diff for which a migration should be generated by printing a nice message,
* exit with a non-zero exit code

The non-zero exit code is necessary to ensure tooling can know the result of a command call, e.g. whether a migration was generated, none was necessary or another error occurred. Parsing the command output is generally not a good idea in these scenarios: we don't want people to rely on the wording of a success/error message in the output. That's what exit codes are for.

One could argue that the current exception has a similar result, but an uncaught exception always feels like an unexpected error. Not being able to generate a migration because the migrations are up-to-date with the database is not such an error case. That's why the command should gracefully handle the exception, with tooling handling the resulting return code and moving on if no special handling is necessary when no migration could be generated.

.gitignore
jawira: I added a comment instead, hope you like it

lib/Doctrine/Migrations/Generator/DiffGenerator.php
jawira: I restored this

jawira:
> Thanks for your contribution, I think it needs some minor adjustment and would love to hear the reason why this is important to you.

Well, about why I need this. To avoid any error I regularly execute three commands in a row: "migrate - diff - migrate". Because typing these three commands every time is a pain in the ass, I use buildtools to do it (i.e. Phing or Make). Thing is, buildtools rely on returned value, and when Migration throw an error all execution is stopped. 

You can found some more details in original issue https://github.com/doctrine/migrations/issues/833, but I am here to answer any questions you may have.

lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
```diff
@@ -81,6 +82,12 @@ protected function configure() : void
                 InputOption::VALUE_OPTIONAL,
                 'Check Database Platform to the generated code.',
                 true
+            )
+            ->addOption(
+                'allow-no-diff',
```
greg0ire: How about `allow-empty-diff` instead? `allow-no-diff` can be interpreted as "Not a single diff will be allowed".

lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
```diff
@@ -81,6 +82,12 @@ protected function configure() : void
...
+                null,
+                InputOption::VALUE_NONE,
+                "Don't throw an exception when no changes detected (CI)."
```
greg0ire: I'd avoid abbreviations, or specifying context (this option might be helpful outside CI, in a pre-push hook for instance).
```suggestion
                "Do not throw an exception when no changes are detected."
```

lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
jawira: Hi, thanks for your comments. 
I understand what you saying, but I would prefer to use `allow-no-changes`, what do you think? the rationale is that `allow-empty-diff` may be interpreted as "An empty diff file will be created" and - correct me if I'm wrong - that's what `GenerateCommand` does. Instead this command only displays a message instead of an Exception.

lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
jawira: Well, to create this option I used another option as reference. The option in question is `allow-no-migration` from `MigrateCommand`, as you can see the option's description is:

https://github.com/doctrine/migrations/blob/0069f40df5d7fbe54c32f03843bb36a1c7123bc9/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php#L62

The effect of this option is the same as the options I have written: display a message instead of throwing an exception. I have no problem changing the option's description, but I suggest to also change the message for `allow-no-migration` option, to keep some coherence.

lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
greg0ire: I think `allow-no-changes` presents the same issue as `allow-no-diff`, and that `allow-empty-diff` will hardly be interpreted as "An empty diff file will be created"  when the context is the diff command.

lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
greg0ire: 👍 please also change the `allow-no-migration` option then.

lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
jawira: Ok :)

lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
jawira: Done.

lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
```diff
@@ -59,7 +59,7 @@ protected function configure() : void
                 'allow-no-migration',
                 null,
                 InputOption::VALUE_NONE,
-                'Don\'t throw an exception if no migration is available (CI).'
+                'Do not throw an exception if no migration is available.'
```
alcaeus: `Don't` definitely rings nicer, I don't think you should change that.
```suggestion
                'Don\'t throw an exception if no migration is available.'
```

lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
```diff
@@ -81,6 +82,12 @@ protected function configure() : void
                 InputOption::VALUE_OPTIONAL,
                 'Check Database Platform to the generated code.',
                 true
+            )
+            ->addOption(
+                'allow-empty-diff',
+                null,
+                InputOption::VALUE_NONE,
+                'Do not throw an exception when no changes are detected.'
```
alcaeus:
```suggestion
                'Don\'t throw an exception when no changes are detected.'
```

jawira: It wasn't my idea. Please see https://github.com/doctrine/migrations/pull/840#discussion_r307963742

lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
jawira: It was not my idea. Please see https://github.com/doctrine/migrations/pull/840#discussion_r307963742

alcaeus: Oh, never mind then. In that case, just squash commits please :)

jawira: done

jawira:
Hi guys, Is there anything I can do for you to accept this PR? Thanks :)

SenseException:
For now, it has 3 approvals and is waiting to be merged or someone else might find something in another review. Sorry, but the waiting part is the hardest part of a PR. I also know that well. Please be patient and thank you for your contribution and the time you've spent into it. It is appreciated even though it isn't merged yet.

alcaeus:
Thanks @jawira!

goetas:
The tests for this contribution were quite weak. It was testing only the happy path, not the case when the exception is thrown.

didacrios:
Just a suggestion @jawira could you replace in the summary the description of the command:

It says `--allow-no-diff` while the command is `--allow-empty-diff` and maybe somebody (like a friend of mine 🙄)  tries this command and went nuts not understanding why this command is not working 

Thanks!

jawira:
> Just a suggestion @jawira could you replace in the summary the description of the command:
> 
> It says `--allow-no-diff` while the command is `--allow-empty-diff` and maybe somebody (like a friend of mine roll_eyes) tries this command and went nuts not understanding why this command is not working
> 
> Thanks!

Done! :)

PR #841, opened 2019-07-23
Fixed default name in the documentation: Doctrine Database Migrations

webimpress:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

Default name for is `Doctrine Database Migrations`, not as it is said in the documentation `Doctrine Migrations`.

Ref:
https://github.com/doctrine/migrations/blob/941231fde05b95678f4fc9fdc1368acb42c174e2/lib/Doctrine/Migrations/Tools/Console/Command/AbstractCommand.php#L100
https://github.com/doctrine/migrations/blob/c369ce8f1f41275faed0e57c786ce291d9d83164/lib/Doctrine/Migrations/Tools/Console/Helper/MigrationStatusInfosHelper.php#L46

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -90,29 +90,29 @@ Please note that if you want to use the YAML configuration option, you will need
 
 Here are details about what each configuration option does:
 
-+----------------------------+-------------+----------------------------+----------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


alcaeus:
Thanks @webimpress!

PR #842, opened 2019-07-24
Fixed docs for required params in the configuration

webimpress:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

It looks like `migrations_namespace` and `migrations_directory` are
mandatory to use any commands. Otherwise we are getting exception.

Ref:

- in AbstractCommand on initialize we validate the configuration:
https://github.com/doctrine/migrations/blob/0069f40df5d7fbe54c32f03843bb36a1c7123bc9/lib/Doctrine/Migrations/Tools/Console/Command/AbstractCommand.php#L67-L75

- and here is the validate method of configuration:
https://github.com/doctrine/migrations/blob/0069f40df5d7fbe54c32f03843bb36a1c7123bc9/lib/Doctrine/Migrations/Configuration/Configuration.php#L264-L273

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -90,29 +90,29 @@ Please note that if you want to use the YAML configuration option, you will need
 
 Here are details about what each configuration option does:
 
-+----------------------------+------------+----------------------------+----------------------------------------------------------------------------------+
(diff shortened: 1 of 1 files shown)


alcaeus:
Thanks @webimpress!

PR #854, opened 2019-09-06
Fix migration direction calculation fails.

jamesleesaunders:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | N/K

#### Summary

When using a timestamp for version numbers the direction calculation fails due to overflow of integer. This PR removed the (int) to allow timestamps to be used.

### DIFF
--- a/lib/Doctrine/Migrations/MigrationPlanCalculator.php
+++ b/lib/Doctrine/Migrations/MigrationPlanCalculator.php
@@ -54,14 +54,12 @@ final class MigrationPlanCalculator
         string $to,
         array $migrated
     ) : bool {
-        $to = (int) $to;
-
         if ($direction === Direction::DOWN) {
             if (! in_array($version->getVersion(), $migrated, true)) {
                 return false;
             }
 
-            return (int) $version->getVersion() > $to;
+            return $version->getVersion() > $to;
         }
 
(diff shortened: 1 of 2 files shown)


alcaeus:
What platform are you running PHP on? I don't know what year you use for your versions, but a string like `20161202011555` will not cause an int overflow on 64-bit systems. This being 2019, I'm very inclined to ignore issues due to 32-bit systems.

jamesleesaunders:
Hi @alcaeus Unfortunately we are running on Solaris 64 X86.
Examples:
* Version20190903081429.php
* Version20190904080206.php
Appreciate this is a little in the dark ages.

lcobucci:
@jamesleesaunders @alcaeus perhaps if we use `DateTimeImmutable` objects to perform the comparison we might accommodate all needs? 

It might be a bit wasteful to create an object only for that but at least it won't break things.

alcaeus:
That would be one option, given that we're sure that the version is always going to be in the expected format.

goetas:
This is not a problem anymore with 3.0 since there migrations are always considered as a string

goetas:
Starting from https://github.com/doctrine/migrations/pull/858, migrations are considered as strings, thus this problem does not existing anymore.

Issue #853, opened 2019-09-06
Migration dependencies


goetas:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | ?? (hopefully no)

#### Summary

Hi, 
I'm working on an application that has some kind of plugins. Each plugin can have dependencies with other plugins. Each plugin has its-own set of tables, (and migrations).
Tables have foreign keys to other tables  (created by plugins on which the current plugin depends).

This scenario requires to load migrations from multiple paths, and to have a different execution order from the current default migrations sorting and the possibility to load migrations from multiple folders.

From what I could see, this is not yet supported by doctrine migrations. I'm planning to work on something to implement this kind of feature.
The idea is to introduce an interface `MigrationsSorter` used as migration sorting strategy. The default implementation will use the current sorting algorithm, but it will be possible to provide different sorting mechanism.

Is something as this interesting for this project?

alcaeus:
> Is something as this interesting for this project?

Absolutely. I needed something like this myself when working on an application that uses Sylius. We needed to run all migrations from Sylius but also add our own migrations, which wasn't possible by default. The only way we achieved this was by symlinking the Sylius migrations and making sure that our migrations use the same namespace as Sylius migrations, which was hacky to say the least.

I'm not sure what we need to do to fully support this, but I think loading migrations from other packages (like we load mappings for our object managers) would definitely be a great improvement for this library!

goetas:
great to hear that! As soon as I have a more clear road map, will give you an update

goetas:
Started to work on this in https://github.com/doctrine/migrations/pull/855

slince:
Hello, I'm having a similar problem. I'm trying to extend these classes myself, but the dependencies seem too complex; Many of the properties are private, which is so hard to extend that I have to re-inherit almost all the classes. Is there any plan to re-implement this?

goetas:
See https://github.com/doctrine/migrations/pull/858

goetas:
Fixed in https://github.com/doctrine/migrations/pull/858

Issue #852, opened 2019-09-06
Store version description into migrations table


alOneh:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

It could be a nice feature to have the description we provide in the Migration class generated filled also in the migration_versions table.

The idea is to have a new column `description` in the `migration_versions` table to see what happens without going to the source code. Could be interesting for dba or release manager.

alcaeus:
What's the motivation for keeping that? Note that I'm not saying we shouldn't do it, I'm just trying to understand the use-case.

alOneh:
To be honest, at the first time I discuss about that with a colleague who wants to separate logically the migration classes by the sprint number (agile mode) to know what db changes we made during the given sprint number. But for me it was really weird, so I dig into the code and see that we can use the description for that but the inconvenient is that it is not persisted in the db.

So the motivation for keeping that is only to keep track of that information, this description is only use when you run the migrate command but if you want to know what a version does without having access to the source code the migration_versions table does not tell you that and you only have the reference to the version number.

Hope it helps.

jwage:
My initial reaction is I don't feel this should be a part of the migrations tool. I think it should probably be accomplished outside of the production database and the database migrations tool. Most likely somewhere in the project management tooling.

PR #856, opened 2019-09-07
Allow multiple migration directories and namespaces

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | yes/no
| Fixed issues | #853

#### Summary
This is a WIP pr for #853, is allows to define multiple paths (and namespaces) for migrations.

The generation and reporting are not yet implemented, I'm publishing the PR as work-in-progress to get early feedback.

### DIFF
--- a/lib/Doctrine/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/Migrations/Configuration/Configuration.php
@@ -49,6 +49,9 @@ class Configuration
     /** @var string */
     private $migrationsExecutedAtColumnName = 'executed_at';
 
+    /** @var string[][] */
+    private $migrationsDirectories = [];
+
     /** @var string|null */
     private $migrationsDirectory;
 
@@ -101,6 +104,19 @@ class Configuration
         $this->queryWriter            = $queryWriter;
         $this->dependencyFactory      = $dependencyFactory;
         $this->migrationsColumnLength = strlen($this->createDateTime()->format(self::VERSION_FORMAT));
+    }
+
+    public function addMigrationsPath(string $path, ?string $namespace = null)
+    {
+        $this->migrationsDirectories[(string)$namespace][$path] = $path;
+    }
+
+    public function getMigrationDirectories() : array
+    {
+        if ($this->migrationsDirectory) {
+            $this->migrationsDirectories[(string)$this->migrationsNamespace][$this->migrationsDirectory] = $this->migrationsDirectory;
+        }
+        return $this->migrationsDirectories;
     }
 
     public function setName(string $name) : void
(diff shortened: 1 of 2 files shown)


goetas:
The difficould part of this PR is that `version` contains only the "version" number, but when migrations will have multiple namespaces, version number might clash...

alcaeus:
> The difficould part of this PR is that `version` contains only the "version" number, but when migrations will have multiple namespaces, version number might clash...

That is correct. I believe that the version number no longer needs to be unique, but it must be unique within the migration namespace. Would you agree?

goetas:
> That is correct. I believe that the version number no longer needs to be unique, but it must be unique within the migration namespace. Would you agree?

I agree, but would this kind of change considered a BC break? Or the `getVersion()` is something "internal" not covered but BC promise?

I would like to know this because if is a BC break then I would choose a different approach and target it to a possible 3.0...

goetas:
Another question: the "version" is used to decide with entry store in the migrations table, if we change that behavior, then the "migrations_table" will need to be migrated to a new format... 

And this can be an even bigger headache even for 3.0

alcaeus:
Yes, that is always going to be a headache. I was already thinking about this when looking at #852: technically, it's an easy change, but changing the migrations table makes it very expensive and dangerous to do so. It would be nice if we could introduce migrations for the migrations table itself, but it's not easy to build.

@jwage any idea here? Did we already have situations where we changed the migration table structure?

goetas:
If we are going in direction to fix some of this issues, would be nice to have some feedback on https://github.com/doctrine/migrations/issues/857 (@jwage ?)

I plan to be able to allocate some time to work on this...

alcaeus:
> If we are going in direction to fix some of this issues in 3.0, would be nice to have some feedback on #857 (@jwage ?)

Absolutely.

> I plan to be able to allocate some time to work on this...

That is awesome, thank you very much! ❤️
I don't have much time to take care of the migrations library, but I'll try to keep feedback coming. Feel free to ping me in Slack if I miss some GitHub notifications!

goetas:
@alcaeus I remember that 1-3 years ago @jwage worked on a 2.0 migrations branch that was also considering getting rid of thew `down()` migration direction. But than that got discarded in what become the current 2.0... Is there some "written" reason on that decision? I'm a firm believer that was a good idea...

alcaeus:
Not that I know of - 2.0 was mainly the work of @jwage. I'm not very involved in this one.

lib/Doctrine/Migrations/MigrationRepository.php
```diff
@@ -391,15 +391,23 @@ public function getNextVersion() : ?string
...
-        $migrationsDirectory = $this->configuration->getMigrationsDirectory();
+        $migrationDirectories = $this->configuration->getMigrationDirectories();
 
-        if (count($this->versions) !== 0 || $migrationsDirectory === null) {
+        if (count($this->versions) !== 0 || count($migrationDirectories) === 0) {
             return;
         }
 
-        $this->registerMigrationsFromDirectory($migrationsDirectory);
+        foreach ($migrationDirectories as $namespace => $paths) {
```
lcobucci: I think that exposes the internals of `Configuration` a bit too much, maybe we can provide a method there to encapsulate things in a better way.  Like:

```php
class Configuration
{
    /**
     * @return iterable<array<string|null>>
     */
    public function iterateMigrationDirectories() : iterable
    {
        foreach ($this->migrationsDirectories as $namespace => $paths) {
            foreach ($paths as $path) {
                yield [$path, $namespace ?: null];
            }
        }
    }
}
```

Then we could do this here:

```php
foreach ($this->configuration->iterateMigrationDirectories() as $directoryConfiguration) {
    $this->registerMigrations(
        $this->migrationFinder->findMigrations(...$directoryConfiguration)
    );
}
```

We would also be able to drop `count($migrationDirectories) === 0` too since we would have an empty iterator when nothing is configured.

lcobucci: If `Configuration#getMigrationDirectories()` is only used by this method we can simply no create it too :smile:

lib/Doctrine/Migrations/Configuration/Configuration.php
```diff
@@ -103,6 +106,19 @@ public function __construct(
         $this->migrationsColumnLength = strlen($this->createDateTime()->format(self::VERSION_FORMAT));
     }
 
+    public function addMigrationsPath(string $path, ?string $namespace = null)
```
lcobucci: I'd suggest always requiring the namespace.

goetas:
Moving development to #858

@lcobucci Thanks for the review. I have addressed most of your suggestions and started  a bigger set of changes in #858

Issue #834, opened 2019-06-26
Less verbose output for the migrate command


stof:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | yes
| BC Break    | no (unless some tools parse the text output of the `migrate` command)

#### Summary

Currently, running the `migrate` commands in normal verbosity displays a lot of output (it shows each SQL query plus some details about each migration execution). And the only lower verbosity is `--quiet` which won't display anything at all (an uncaught exception would still be rendered by the Console component, but that's all).
In my project, this made us face an issue due to the size of this output: our Travis CI setup runs migrations (from a blank state) to initialize the DB (which helps us ensuring that migrations are correct at least regarding schema changes and the validity of other data-migrating queries). The project currently contains 538 DB migrations. And the `migrate` command alone generates ~7600 lines of output. This makes our CI output reach the Travis log limit, which truncates the logs (and so hides the output about which test is failing...). I know that I could  dump the schema and rollup migrations to limit how many we have (and I will at some point, but I need to freeze development changing the DB schema to perform it, and find a way to hook the `rollup` command in my development process at that point and then revert to `migrate`) but that would happen again in the future.

Would you consider moving the existing output to the `verbose` level, and having a much more compact progress report at `normal` level (maybe with a dot progress, as done by PHPUnit when running tests ?) ?

jwage:
I would be fine with this change. I faced the same problem in one of my projects in the past and I would periodically rollup the starting point with a schema.sql file that I load first, and then run migrations from that point. Running all the migrations from 10 years of development really took a long time so we made it a regular process/practice to rollup the migrations.

stof:
Well, the fact that such pattern took a very long time in the past was the reason why I contributed https://github.com/doctrine/migrations/pull/323 (which was then replaced by https://github.com/doctrine/migrations/pull/407 as a better way to achieve the same optimization result). Time stays under control now (running these 538 migrations takes 10s on Travis)
Migrations 2.0 now supports this dump+rollup feature, but changing my deployment process to run `rollup` instead of `migrate` for that particular deployment is something I haven't figured out yet.

On a side note, this has another benefit for Travis: when loading the logs of a build, the collapsing of the setup command happens only when the full output of that command is received (as the end marker is at the end of this output) and a 7600 lines output covers many chunks and so stays uncollapsed long enough to annoy someone starting to scroll past it. A more compact output will help here.

jwage:
[...]

PR #858, opened 2019-09-15
Migrations 3.0

goetas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | https://github.com/doctrine/migrations/issues/857 https://github.com/doctrine/migrations/issues/853 https://github.com/doctrine/migrations/issues/726 https://github.com/doctrine/migrations/issues/805 https://github.com/doctrine/migrations/issues/742 https://github.com/doctrine/migrations/pull/854

#### Summary
This PR proposes what could become doctrine/migrations 3.0. (Some code and ideas are borrowed from https://github.com/doctrine/migrations/pull/162)

This is a "work in progress", so any feedback is appreciated. 

Exiting user-facing features are have been kept as they are ( some BC breaks are present of course).

Summary of changes:
-  The "version" is not anymore a number or date, but the FQCN of the migration class.
-  Multiple namespaces are allowed (and paths). This allows to provide third party migrations, plugin-provided migrations and and dependent migrations as well (as described in  #853)
- Extracted the "metadata table"  logic in its own interface as in https://github.com/doctrine/migrations/pull/162 
- Most of the logs/prints not use psr/log 
- Used dependency injection in most of the "core" code parts (v1 and v2 have the god classes as `Configuration`, `MigrationRepository` and `Version` that were doing all possible kind of things and having circular dependencies between each other )

Tests have not been updated, but the provided prototype has most of the features already working. 
Rollup and Diff commands have not been updated yet.

I realize that is a huge PR and hard to review, but the configuration and migration-repository class was passed all over the places, and as soon as I started to remove methods from it, each other class had to be changed...

Todo:
- [x] migrate all functionalities
- [x] update tests
- [x] phpstan
- [x] prepare migration path for 3.0 update (see https://github.com/goetas/migrations/pull/1)
- [x] document migration path

PR #859, opened 2019-09-15
enable verbosity for OutputWrite

mssimi:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #834 

#### Summary

`migrations:migrate` now shows less info, to show detailed output use option  `-v` (verbose).

Example output:

97v8wEy

### DIFF
--- a/lib/Doctrine/Migrations/OutputWriter.php
+++ b/lib/Doctrine/Migrations/OutputWriter.php
@@ -31,9 +31,9 @@ class OutputWriter
         $this->callback = $callback;
     }
 
-    public function write(string $message) : void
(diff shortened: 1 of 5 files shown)


stof:
I'm not sure using the Symfony Console verbosity constant in the API of the OutputWriter is the right choice, given that this API does not depend on symfony/console at all currently. I think we rather need an API exposed for OutputWriter explicitly, and deal with the Symfony Console verbosity constant only in the glue layer.

lcobucci:
This class is being removed in #858 so we shouldn't keep this. Thanks @mssimi for your work

PR #860, opened 2019-09-30
Fix the installation instruction

stof:
|      Q       |   A
|------------- | -----------
| Type         | doc
| BC Break     | no
| Fixed issues | n/a

#### Summary

We should not recommend an exact match on 2.0.

I used double quotes to make this command work as is cross-platform (the `^` char needs to be escaped in Windows as it has a special meaning)


alcaeus:
Thanks @stof!

Issue #861, opened 2019-10-01
Remove PHP 7.4 from `allow_failures`


carusogabriel:
PHP 7.4 will be released [soon](https://wiki.php.net/todo/php74) and we need to ensure that `Doctrine` is compatible with it.

All we need to do is to remove these following lines

https://github.com/doctrine/migrations/blob/50c7d160c249b9acbc25090a6a8eb2ee8f66e959/.travis.yml#L26-L27

 and verify that everything is working.

This is a very good issue for those who want to participate in https://hacktoberfest.digitalocean.com.

WilChrist:
I Can handle this.

greg0ire:
@WilChrist you can have a look at the Travis build: https://travis-ci.org/doctrine/migrations/builds/592060520?utm_source=github_status&utm_medium=notification (right now it's not working, sorry)

WilChrist:
yes @greg0ire but the errors seems to be in the test, caused by deprecated functions ? should i replace them by currently used functions ?

greg0ire:
Yes, absolutely, please do that!

WilChrist:
@greg0ire how could i test localy if it work ?

greg0ire:
You can try running `./vendor/bin/phpunit`. You will find more information about how to contribute here: https://www.doctrine-project.org/contribute/index.html

goetas:
Fixed in https://github.com/doctrine/migrations/pull/869

PR #862, opened 2019-10-01
allow_failures: - php: 7.4snapshot removed

WilChrist:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #861 

#### Summary

I just remove these two lines ```  allow_failures:
    - php: 7.4snapshot``` + the jobs one also (because i think it's not usefull anymore for now)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,10 +22,6 @@ install:
 script:
   - ./vendor/bin/phpunit
 
-jobs:
-  allow_failures:
-    - php: 7.4snapshot
-
   include:
     - stage: Test
       env: DEPENDENCIES=low
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ install:
   - rm composer.lock
   - travis_retry composer update --prefer-dist
 
+jobs:
+
 script:
   - ./vendor/bin/phpunit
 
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -22,10 +22,6 @@ install:
 script:
   - ./vendor/bin/phpunit
 
-jobs:
```
alcaeus: The `jobs` line is still necessary as the resulting YAML file will be invalid otherwise.

WilChrist: Ok, i'll get it back

.travis.yml
```diff
@@ -19,13 +19,11 @@ install:
   - rm composer.lock
   - travis_retry composer update --prefer-dist
 
+jobs:
```
stof: adding it here is wrong. `include` was a child of `jobs` (sibling of `allow_failures`). Moving the top-level key withut moving its content makes no sense (`include` is not nested inside `script`, in an invalid way)

WilChrist: Yes you're right i missed that

WilChrist:
@stof and now ?

WilChrist:
@carusogabriel Thanks for guidance. I'm pretty new in contributing to open source, even in using doctrine, i used it just something like one or two times in projets... and i never look at its internal untill now.

alcaeus:
Superseded by #869. Thanks for your work @WilChrist!

PR #867, opened 2019-11-10
Merge release 2.1.2 into master

doctrinebot:
Release [2.1.2](https://github.com/doctrine/migrations/milestone/28)

2.1.2
=====

- Total issues resolved: **0**
- Total pull requests resolved: **1**
- Total contributors: **1**

Documentation
-------------

 - [860: Fix the installation instruction](https://github.com/doctrine/migrations/pull/860) thanks to @stof

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -38,7 +38,7 @@ Now, in the root of your project place a file named ``migrations.php``, ``migrat
     .. code-block:: yaml
 
         name: "My Project Migrations"
-        migrations_namespace: "MyProject\Migrations"
+        migrations_namespace: "MyProject\\Migrations"
         table_name: "doctrine_migration_versions"
         column_name: "version"
         column_length: 14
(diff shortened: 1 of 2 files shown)


alcaeus:
Appveyor failure is unrelated to this PR, would override.

@Ocramius not sure how to proceed with the requirement of having the PR up-to-date with the base branch: that kinda defeats the purpose of doing these merge-ups. Shall we disable that protected branch rule in our repos?

Ocramius:
I don't think these merge-up branches will ever be mergeable via UI anyway

alcaeus:
> I don't think these merge-up branches will ever be mergeable via UI anyway

With branch protection enabled, they can only be merged via UI since we can't push directly to other branches. Hence the idea to disable the up-to-date rule.

I agree that they will almost never be mergeable without manual work, but that's also the reason why we decided to create PRs from intermediate branches instead of merging directly in the tool.

PR #868, opened 2019-11-12
Escape backslash in YAML example

czarpino:
Single backslash in double quotes causes malformed YAML error

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | ~

#### Summary

Escape backslash inside double quote in YAML configuration example. It was causing a YamlNotValid exception.


alcaeus:
Thanks @czarpino!

PR #869, opened 2019-11-13
Update supported PHP versions

alcaeus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

* ~Drop PHP 7.1~ skipped for now to help out Symfony users
* Require build for PHP 7.4 to pass

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,9 +23,6 @@ script:
   - ./vendor/bin/phpunit
 
 jobs:
-  allow_failures:
-    - php: 7.4snapshot
-
   include:
     - stage: Test
       env: DEPENDENCIES=low
--- a/composer.lock
+++ b/composer.lock
@@ -1095,27 +1095,29 @@
         },
         {
             "name": "doctrine/instantiator",
-            "version": "1.1.0",
+            "version": "1.3.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/instantiator.git",
-                "reference": "185b8868aa9bf7159f5f953ed5afb2d7fcdc3bda"
+                "reference": "ae466f726242e637cebdd526a7d991b9433bacf1"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/instantiator/zipball/185b8868aa9bf7159f5f953ed5afb2d7fcdc3bda",
-                "reference": "185b8868aa9bf7159f5f953ed5afb2d7fcdc3bda",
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -28,13 +28,15 @@ jobs:
...
       env: DEPENDENCIES=low
+      php: 7.2
       install:
         - rm composer.lock
         - travis_retry composer update --prefer-dist --prefer-lowest
         - ./download-box.sh
 
     - stage: Test
       env: COVERAGE
+      php: 7.3
```
carusogabriel: Why not `7.2` like the other jobs?

.travis.yml
```diff
@@ -23,19 +22,18 @@ script:
   - ./vendor/bin/phpunit
 
 jobs:
-  allow_failures:
-    - php: 7.4snapshot
-
   include:
     - stage: Test
       env: DEPENDENCIES=low
+      php: 7.2
```
lcobucci: Not needed since `7.2` is the lowest version already.

.travis.yml
```diff
@@ -23,19 +22,18 @@ script:
...
       install:
         - rm composer.lock
         - travis_retry composer update --prefer-dist --prefer-lowest
         - ./download-box.sh
 
     - stage: Test
       env: COVERAGE
+      php: 7.3
```
lcobucci: What's the need for this?

.travis.yml
```diff
@@ -47,10 +45,12 @@ jobs:
 
     - stage: Code Quality
       env: CODING_STANDARDS
+      php: 7.2
```
lcobucci: Not needed since `7.2` is the lowest version already.

.travis.yml
```diff
@@ -47,10 +45,12 @@ jobs:
...
       install: travis_retry composer install --prefer-dist
       script: ./vendor/bin/phpcs
 
     - stage: Code Quality
       env: STATIC_ANALYSIS
+      php: 7.2
```
lcobucci: Not needed since `7.2` is the lowest version already.

.travis.yml
alcaeus: Because typo :(

.travis.yml
alcaeus: Wasn't sure what version it picks, but if we can be sure it's 7.2 I'll drop it.

composer.lock
```diff
@@ -4059,7 +4073,7 @@
     "prefer-stable": false,
     "prefer-lowest": false,
     "platform": {
-        "php": "^7.1"
+        "php": "^7.2"
```
lcobucci: This should be `^7.1` then :+1:

PR #872, opened 2019-11-20
Fix of branch name for doctrine website build

SenseException:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

This should fix the migrations version infos and the build error `error: pathspec 'origin/2.1' did not match any file(s) known to git` on Doctrine website build.


alcaeus:
Thanks @SenseException!

PR #876, opened 2019-11-30
Make configuration and dependency injection container immutable

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | yes
| Fixed issues | -

#### Summary

This PR makes the configuration and dependency injection container immutable. (frozen state)

This avoids having other services hacking into the DI container or config object at runtime (avoiding inconsistent retrieval of services

### DIFF
--- a/lib/Doctrine/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/Migrations/Configuration/Configuration.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Migrations\Configuration;
 
+use Doctrine\Migrations\Configuration\Exception\FrozenConfiguration;
 use Doctrine\Migrations\Configuration\Exception\MissingNamespaceConfiguration;
 use Doctrine\Migrations\Configuration\Exception\UnknownConfigurationValue;
 use Doctrine\Migrations\Exception\MigrationException;
@@ -49,8 +50,27 @@ final class Configuration
     /** @var MetadataStorageConfiguration */
     private $metadataStorageConfiguration;
 
+    /** @var bool */
+    private $frozen = false;
+
+    public function freeze() : void
+    {
+        if (! $this->frozen) {
+            $this->validate();
+        }
+        $this->frozen = true;
+    }
+
+    private function assertNotFrozen() : void
+    {
+        if ($this->frozen) {
+            throw FrozenConfiguration::new();
+        }
+    }
+
     public function setMetadataStorageConfiguration(MetadataStorageConfiguration $metadataStorageConfiguration) : void
     {
+        $this->assertNotFrozen();
         $this->metadataStorageConfiguration = $metadataStorageConfiguration;
     }
 
@@ -64,6 +84,7 @@ final class Configuration
 
     public function addMigrationClass(string $className) : void
     {
+        $this->assertNotFrozen();
         $this->migrationClasses[] = $className;
     }
 
@@ -74,6 +95,7 @@ final class Configuration
 
     public function addMigrationsDirectory(string $namespace, string $path) : void
     {
+        $this->assertNotFrozen();
         $this->migrationsDirectories[$namespace] = $path;
     }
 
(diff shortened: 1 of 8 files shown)


lib/Doctrine/Migrations/Configuration/Configuration.php
```diff
@@ -169,11 +196,14 @@ public function setCheckDatabasePlatform(bool $checkDbPlatform) : void
 
     public function isDatabasePlatformChecked() : bool
     {
+        $this->checkFrozenConfiguration();
```
Koc: This methodis not mutate anything. Do we really need check here?

lib/Doctrine/Migrations/DependencyFactory.php
```diff
@@ -71,6 +75,19 @@ public function __construct(Configuration $configuration, Connection $connection
         $this->em            = $em;
     }
 
+    public function freeze() : void
+    {
+        $this->frozen = true;
+        $this->configuration->freeze();
+    }
+
+    private function checkFrozenDependencies() : void
```
Koc: Private methods usually defined after public. Does Doctrine follows same conversation? Also what do you think about start this method name with assert prefix, like assertNotFrozen?

tests/Doctrine/Migrations/Tests/Tools/Console/Command/DoctrineCommandTest.php
```diff
@@ -0,0 +1,37 @@
...
+{
+    public function testCommandFreezes() : void
+    {
+        $dependencyFactory = $this->getMockBuilder(DependencyFactory::class)
+            ->disableOriginalConstructor()
+            ->setMethods(['freeze'])
+            ->getMock();
+
+        $dependencyFactory
+            ->expects($this->once())
```
greg0ire: This should fix the build:
```suggestion
            ->expects(self::once())
```

lib/Doctrine/Migrations/DependencyFactory.php
greg0ire: We don't AFAIK, we usually enforce our conventions. Love the `assertNotFrozen` suggestion 👍

goetas:
@greg0ire @Koc thanks for the review. all suggestions have been implemented

lib/Doctrine/Migrations/DependencyFactory.php
```diff
@@ -197,9 +215,13 @@ public function getMigrationRepository() : MigrationRepository
         });
     }
 
-    public function setMetadataStorageConfiguration(MetadataStorageConfiguration $metadataStorageConfigration) : void
+    /**
+     * @param mixed $service
```
greg0ire: I think you can't have anything but an object here, right?

```suggestion
     * @param object $service
```

goetas: I think `callable` is a possible value for the migrations sorter... of course can be wrapped by an object via `__invoke`....

greg0ire: If that's the case, the phpdoc at the beginning of the file is wrong, so please use the same thing in both places (either `mixed` or `object|callable` or `object`)

goetas: You raised a good point. Actually there is a `setSorter` method responsible for setting a custom sorter. Does it make sense to have `setSorter`? we could use `setService` and drop `setSorter` ... but we do not have an interface for the sorter, since is just any callback...

Ideas?

greg0ire: I'd either leave `sorter` alone and use `object`, or add an interface and provide a `FromCallableSorter` implementation

goetas: What about this approach https://github.com/doctrine/migrations/pull/876/commits/a89d28916fe0ce495438bb59a890611239a9445d ?

greg0ire: Looks good 👍

goetas: good!

PR #879, opened 2019-12-01
Drop php 7.1 and allow phpunit 8

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | -

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -15,7 +15,6 @@ environment:
   matrix:
     - php: 7.2.23
     - php: 7.3.10
-    - php: 7.1.32
 
 init:
   - SET PATH=C:\Program Files\OpenSSL;c:\tools\php;C:\tools\composer;%PATH%
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -13,8 +13,8 @@ cache:
 
 environment:
   matrix:
-    - php: 7.2.23
-    - php: 7.3.10
+    - php: 7.2.25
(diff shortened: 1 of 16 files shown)


lcobucci:
Let's just bump up php version and require phpunit 8. Much easier for everyone 😁

goetas:
phpunit 8 requires php 7.2, while the minimum version for doctrine migrations is 7.1. Is it ok to increase the min version to 7.2 for this library too?

greg0ire:
>  Is it ok to increase the min version to 7.2 for this library too?

7.1 is EOL, I think it is ok.

goetas:
>> Is it ok to increase the min version to 7.2 for this library too?

> 7.1 is EOL, I think it is ok.

Happy to hear that. Build is green!

greg0ire:
FWIW, I think we could even do this on 2.2.x , not sure what others think.

goetas:
appveyor is having some issues https://status.chocolatey.org/issues/2019-11-29-slow-and-intermittent-connection-issues/

alcaeus:
@greg0ire I don’t see a reason to drop 7.1 in 2.x.

greg0ire:
Allowing new features like the `object` type declaration in new code could be a reason, but maybe that's something we want to do if and when someone makes a PR that would benefit from such features.

alcaeus:
We can always do that when the new API shows up.

PR #880, opened 2019-12-01
Use executeUpdate instead of executeQuery for write operation

goetas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

Use executeUpdate for migration queries as it is a write operation


greg0ire:

greg0ire:
Also, since there is no BC-break, this should probably target 2.2.x , can another contributor confirm this?

alcaeus:
@greg0ire that is correct - this could easily go into 2.3.0. @goetas do you see any reason why this shouldn't be in 2.3?

goetas:
I think that this could perfectly go into 2.x too.

alcaeus:
Sounds good - can you rebase this onto 2.3.x please? Thanks!

greg0ire:
Oooh it's a duplicate in fact, closing see #888

PR #881, opened 2019-12-01
Allow dump schema when migrations in other namespaces are present

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | yes/no
| Fixed issues | -

#### Summary

Being able to define migrations in multiple namespaces/folders, means that probably they belong to different part of the application, thus should be possible to dump those parts of the schema independently

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/DumpSchemaCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/DumpSchemaCommand.php
@@ -12,10 +12,10 @@ use Symfony\Component\Console\Output\OutputInterface;
 use function addslashes;
 use function assert;
 use function class_exists;
-use function count;
 use function is_string;
 use function key;
 use function sprintf;
+use function strpos;
 
 /**
  * The DumpSchemaCommand class is responsible for dumping your current database schema to a migration class. This is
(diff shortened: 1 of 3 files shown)


greg0ire:
Can you elaborate on the `yes/no` part of the BC-break? I'd like to know if this can go in 2.2…

lib/Doctrine/Migrations/Tools/Console/Command/DumpSchemaCommand.php
```diff
@@ -144,4 +141,14 @@ public function execute(
 
         return 0;
     }
+
+    private function checkNoPreviousDumpExistsForNamespace(string $namespace) : void
+    {
+        $versions = $this->getDependencyFactory()->getMigrationRepository()->getMigrations();
```
greg0ire: Are you sure you want to name that variable `$versions` rather than `$migrations`? I think it's confusing…

goetas:
> but maybe this is built on things that are only on 3.x

this feature makes sense only for 3.x as in older versions only one namespace is available

lib/Doctrine/Migrations/Tools/Console/Command/DumpSchemaCommand.php
goetas: true. fixed

Issue #722, opened 2018-06-05
Eliminate dependency on doctrine/common


jwage:
This will require doctrine/dbal 2.8 without common and the work to continue to split up doctrine/common in to separate projects has to be finished.

stof:
AFAIK, the only part of Doctrine Common used directly in doctrine/migrations is the event part. So as soon as that part is extracted to a dedicated package, we can change the requirement here

jwage:
It will remove doctrine/common from doctrine/migrations directly, but it will still be pulled in by doctrine/dbal dependency until we upgrade to doctrine/dbal 2.8.

stof:
But until we use the event system directly, we cannot remove the direct dependency.

goetas:
Do you thing that https://github.com/doctrine/migrations/pull/883 addresses this issue?

alcaeus:
This library already doesn't depend on doctrine/common anymore, so I'm closing this issue. DBAL is tracked separately.

PR #883, opened 2019-12-01
Depend explicitly on doctrine/event-manager

goetas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

#### Summary

doctrine/migrations is explicitly using the event manager, but its dependency is obtained via transitive dependency on the dbal package. This PR adds it as explicit dependency for the project.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,6 +13,7 @@
     "require": {
         "php": "^7.2",
         "doctrine/dbal": "^2.9",
+        "doctrine/event-manager": "^1.0",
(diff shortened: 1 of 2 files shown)


goetas:
I'm not sure if this is addressing implicitly https://github.com/doctrine/migrations/issues/722

alcaeus:
Was this dependency missing? If so, it might be worth using https://packagist.org/packages/maglnet/composer-require-checker to avoid such issues in the future.

PR #885, opened 2019-12-02
Allow to throw Exception in Migrations

VincentLanglet:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

When generating a migration, I getting
```
    {
        $this->abortIf($this->connection->getDatabasePlatform()->getName() !== 'mysql', 'Migration can only be executed safely on \'mysql\'.');

        ...
    }
```

This means the up function is allowed to throw AbortIf and DBAL Exception but it's not specified in the phpdoc of the parent function. The parent/children Exception should be compatible and some tools are looking for `@throws` tag to check this.

### DIFF
--- a/lib/Doctrine/Migrations/AbstractMigration.php
+++ b/lib/Doctrine/Migrations/AbstractMigration.php
@@ -5,11 +5,13 @@ declare(strict_types=1);
 namespace Doctrine\Migrations;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\DBALException;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\Migrations\Exception\AbortMigration;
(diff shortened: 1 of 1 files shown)


lib/Doctrine/Migrations/AbstractMigration.php
```diff
@@ -93,24 +95,42 @@ public function skipIf(bool $condition, string $message = '') : void
...
     }
 
+    /**
+     * @throws MigrationException|DBALException
+     */
     public function postDown(Schema $schema) : void
     {
     }
-
+
```
greg0ire:
```suggestion
```

VincentLanglet: I should stop doing PR directly on github without cloning/forking and thinking it's faster :/

Fixed !

VincentLanglet:
@alcaeus Like this ? https://github.com/doctrine/migrations/pull/887

greg0ire:
You could have edited the target of this branch and force-pushed ;)

alcaeus:
#887 has been merged; closing here.

PR #889, opened 2019-12-06
Executed at must be an integer

goetas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | -
| Fixed issues | 

#### Summary

- `executed_at` must be an integer

### DIFF
--- a/tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
+++ b/tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
@@ -80,7 +80,7 @@ class TableMetadataStorageTest extends TestCase
     public function testComplete() : void
     {
         $result = new ExecutionResult(new Version('1230'), Direction::UP, new DateTimeImmutable('2010-01-05 10:30:21'));
-        $result->setTime(31);
+        $result->setTime(31.0);
         $this->storage->complete($result);
 
         $sql  = sprintf(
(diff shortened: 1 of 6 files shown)


tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
```diff
@@ -98,6 +98,27 @@ public function testComplete() : void
         ], $rows);
     }
 
+    public function testCompleteWithFloatTime() : void
```
goetas: Unfortunately this test is useless. The issue seems to be in how PDO handles the `Type::INTEGER` typehint for postgres (for SQLite seems fine). Do not know how to test it properly TBH.

tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
```diff
@@ -98,6 +98,27 @@ public function testComplete() : void
...
+    {
+        $result = new ExecutionResult(new Version('1230'), Direction::UP, new DateTimeImmutable('2010-01-05 10:30:21'));
+        $result->setTime(31.49);
```
alcaeus: Why can we set a float when we need an `int`?

goetas: Hmm, it should be a float, since it represents microseconds too. Here the issue is how data are saved. Currently it seems that we loose some info (the sub second execution time).
Will update the PR

goetas:
I have reworked this PR. Now internally the executed time is a float (number of seconds+microtime), but when that value gets saved to the database, it gets converted in an integer (multiplied for 1000).
In this way at DB level is an integer (better compatibility with different databases). When saving the execution time, we loose the microseconds part, but I think that having milliseconds is enough.

PR #891, opened 2019-12-07
Support multiple versions in ExecuteCommand

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | -
| Fixed issues | https://github.com/doctrine/migrations/pull/796

#### Summary

Implementation of https://github.com/doctrine/migrations/pull/796 for the 3.x branch

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
@@ -10,12 +10,13 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use function array_map;
 use function getcwd;
 use function is_string;
 use function is_writable;
 
 /**
- * The ExecutCommand class is responsible for executing a single migration version up or down.
+ * The ExecutCommand class is responsible for executing migration versions up or down manually.
  */
 class ExecuteCommand extends DoctrineCommand
 {
@@ -27,12 +28,12 @@ class ExecuteCommand extends DoctrineCommand
         $this
             ->setAliases(['execute'])
             ->setDescription(
-                'Execute a single migration version up or down manually.'
+                'Execute one or more migration versions up or down manually.'
             )
             ->addArgument(
-                'version',
-                InputArgument::REQUIRED,
-                'The version to execute.',
+                'versions',
+                InputArgument::REQUIRED|InputArgument::IS_ARRAY,
+                'The versions to execute.',
                 null
             )
             ->addOption(
@@ -67,7 +68,7 @@ class ExecuteCommand extends DoctrineCommand
                 'Time all the queries individually.'
             )
             ->setHelp(<<<EOT
-The <info>%command.name%</info> command executes a single migration version up or down manually:
+The <info>%command.name%</info> command executes migration versions up or down manually:
 
     <info>%command.full_name% FQCN</info>
 
@@ -94,14 +95,17 @@ EOT
 
     public function execute(InputInterface $input, OutputInterface $output) : ?int
     {
-        $version   = $input->getArgument('version');
+        $versions  = $input->getArgument('versions');
         $path      = $input->getOption('write-sql');
         $direction = $input->getOption('down') !== false
             ? Direction::DOWN
             : Direction::UP;
 
-        $migrator = $this->getDependencyFactory()->getMigrator();
-        $plan     = $this->getDependencyFactory()->getMigrationPlanCalculator()->getPlanForExactVersion(new Version($version), $direction);
+        $migrator       = $this->getDependencyFactory()->getMigrator();
+        $planCalculator = $this->getDependencyFactory()->getMigrationPlanCalculator();
+        $plan           = $planCalculator->getPlanForVersions(array_map(static function (string $version) : Version {
+            return new Version($version);
+        }, $versions), $direction);
 
         $migratorConfigurationFactory = $this->getDependencyFactory()->getConsoleInputMigratorConfigurationFactory();
         $migratorConfiguration        = $migratorConfigurationFactory->getMigratorConfiguration($input);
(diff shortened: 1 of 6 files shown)


goetas:
@SenseException good catch. I have also added more tests to make sure migrations are properly sorted (keeping sort order and not the order provided by the user via command line)

goetas:
Any other feedback on this?

lib/Doctrine/Migrations/MigrationPlanCalculator.php
```diff
@@ -39,13 +44,30 @@ public function __construct(MigrationRepository $migrationRepository, MetadataSt
...
+    /**
+     * @param Version[] $versions
+     */
+    public function getPlanForVersions(array $versions, string $direction) : MigrationPlanList
     {
-        $migration = $this->migrationRepository->getMigration($version);
-
-        $planItem = new MigrationPlan($migration->getVersion(), $migration->getMigration(), $direction);
+        $migrationsToCheck   = $this->arrangeMigrationsForDirection($direction, $this->migrationRepository->getMigrations());
+        $availableMigrations = array_filter($migrationsToCheck, static function (AvailableMigration $availableMigration) use ($versions) {
```
greg0ire:
```suggestion
        $availableMigrations = array_filter($migrationsToCheck, static function (AvailableMigration $availableMigration) use ($versions) : bool {
```

lib/Doctrine/Migrations/MigrationPlanCalculator.php
```diff
@@ -39,13 +44,30 @@ public function __construct(MigrationRepository $migrationRepository, MetadataSt
...
+            // in_array third parameter is intentionally false to force object to string casting
+            return in_array($availableMigration->getVersion(), $versions, false);
+        });
+
+        $planItems = array_map(static function (AvailableMigration $availableMigration) use ($direction) : MigrationPlan {
+            return new MigrationPlan($availableMigration->getVersion(), $availableMigration->getMigration(), $direction);
+        }, $availableMigrations);
+
+        if (count($planItems) !== count($versions)) {
+            $plannedVersions = array_map(static function (MigrationPlan $migrationPlan) {
```
greg0ire:
```suggestion
            $plannedVersions = array_map(static function (MigrationPlan $migrationPlan) : string {
```

lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
```diff
@@ -67,7 +68,7 @@ protected function configure() : void
                 'Time all the queries individually.'
             )
             ->setHelp(<<<EOT
-The <info>%command.name%</info> command executes a single migration version up or down manually:
+The <info>%command.name%</info> command executes migration versions up or down manually:
 
     <info>%command.full_name% YYYYMMDDHHMMSS</info>
```
alcaeus: This help test should include an example how to run multiple migrations. In addition to that, we should use a FQCN for the migrations instead of the legacy date format.

phpstan.neon.dist
```diff
@@ -12,6 +12,9 @@ parameters:
         # Ignore proxy manager magic
         - '~ProxyManager\\Proxy\\VirtualProxyInterface~'
         - '~Variable method call on Doctrine\\Migrations\\AbstractMigration~'
+        -
+            message: '~^Call to function in_array\(\) requires parameter #3 to be true\.$~'
```
alcaeus: Out of curiosity, is there a reason why we can't enable strict mode on the `in_array` call?

lib/Doctrine/Migrations/MigrationPlanCalculator.php
```diff
@@ -15,17 +16,21 @@
...
+use function reset;
 
 /**
  * The MigrationPlanCalculator is responsible for calculating the plan for migrating from the current
  * version to another version.
  *
  * @internal
  */
-final class MigrationPlanCalculator
+class MigrationPlanCalculator
```
alcaeus: Not happy with opening the method for extension, but I trust that this is the most sensible way to do this. I'm not sure if extracting an interface for the purpose of mocking would be sensible.

lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
goetas: done

phpstan.neon.dist
goetas: the reason of this is explained in https://github.com/doctrine/migrations/pull/891/files#diff-11b21ce809b5364acce6a5fd2482f280R54

lib/Doctrine/Migrations/MigrationPlanCalculator.php
goetas: `MigrationPlanCalculator` will be refactored in a upcoming PR. Currently `getNewMigrations` and `getExecutedUnavailableMigrations` dont belong conceptually to this class. After extracting those methods, I will introduce an interface and mark the class as final again.

lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
```diff
@@ -10,12 +10,13 @@
...
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use function array_map;
 use function getcwd;
 use function is_string;
 use function is_writable;
 
 /**
- * The ExecutCommand class is responsible for executing a single migration version up or down.
+ * The ExecutCommand class is responsible for executing migration versions up or down manually.
```
greg0ire: Let's take this as an occasion to fix the typo?

```suggestion
 * The ExecuteCommand class is responsible for executing migration versions up or down manually.
```

goetas:
@alcaeus @SenseException the changes you asked are done. can you please check again/approve this PR?

lib/Doctrine/Migrations/MigrationPlanCalculator.php
goetas: this has been fixed in https://github.com/doctrine/migrations/pull/897, now the mock is done on the interface and the class is back to final

goetas:
Rebased on top of the latest master branch.

greg0ire:
Thanks @goetas !

PR #893, opened 2019-12-14
Expose the connection in DependencyFactory

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no-

#### Summary
 Expose the connection from the DependencyFactory to ease the testing of the bundle

### DIFF
--- a/lib/Doctrine/Migrations/DependencyFactory.php
+++ b/lib/Doctrine/Migrations/DependencyFactory.php
@@ -93,7 +93,7 @@ class DependencyFactory
         return $this->configuration;
     }
 
-    private function getConnection() : Connection
+    public function getConnection() : Connection
     {
         return $this->connection;
     }
--- a/tests/Doctrine/Migrations/Tests/DependencyFactoryTest.php
+++ b/tests/Doctrine/Migrations/Tests/DependencyFactoryTest.php
@@ -66,4 +66,13 @@ final class DependencyFactoryTest extends MigrationTestCase
 
         self::assertInstanceOf(GlobFinder::class, $finder);
     }
+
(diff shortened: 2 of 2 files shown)


greg0ire:
What's the reason not to contribute this on the stable branch?

goetas:
Mainly because I needed it to write the tests for the new bundle, that will support only migrations 3.x.

greg0ire:
I'd rather have the branches diverge as little as possible, because otherwise merges up will become more and more painful. I had a quick look, and merging up comes with only one conflict, so I think it could be done quickly after this is merged in the stable branch. I'd like to know what @alcaeus thinks about this (the last merge up into master was #867, but it was with 2.1.x, not 2.2.x or 2.3.x, I'm not sure why). If it turns out to be too complicated we should merge into master directly.

goetas:
>  I had a quick look, and merging up comes with only one conflict, so I think it could be done quickly after this is merged in the stable branch. I'd like to know what @alcaeus thinks about this (the last merge up into master was #867, but it was with 2.1.x, not 2.2.x or 2.3.x, I'm not sure why). If it turns out to be too complicated we should merge into master directly.

I was talking with @alcaeus about that and he has to show me how the merging is done with the release tool. 
On the other side, I think that sooner or later 2.x and 3.x will diverge, and backporting commits will be pointless since 2.x and 3.x are really different codebases now. I can continue the backport for some time.

goetas:
Thanks! I'm merging it as it is. Solving conflicts will be a pain anyway, but I do not think that will happen on this lines of code in the `DependencyFactory` class.

Issue #874, opened 2019-11-27
Do not allow down migration, if it up migration was never run


cb0:
### Feature Request

When running a down migration, that was never migrated up, doctrine should notice that and not try to migrate down.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

I'm currently using doctrine/migrations version 2.2.0 with Zend.

If by accident someone calls a down script, as written in the text description after migration was generated. e.g.

    To revert the migration you can use migrations:execute --down 20191127145240

This will most likely result in an error. In the worst case, the database could result in an inconsistent state.

As all migrations that were executed are stored in the `migrations` table, it should just check if it was run before.
Otherwise, it should result in an error like 'This migration was never not run'.

SenseException:
Thank you for opening this issue. As far as I understand this is an edge case where a user has to set a migration version with `--down` that wasn't executed before in a migration?

Do you have a suggestion for a fix for a pull request or even want to create one?

cb0:
Hi @SenseException,

I thought about checking the Version table and only allow Migration if it is in there.
I did create the PR #900 for that issue.

As this is my first PR here I'm not quite sure if I'm missing someting else. And if this will also work for upcoming 3.0.

Please let me know if I should add some other things like tests e.g.

cb0:
I realised that I did the fix for old 2.X release and therefore the build broke.
Changed target again to be 'master' and fixed it in there. 

Had tocreated a new class 'RolldownFailed' as only 'RollupFailed' existed for now.

PR #900, opened 2019-12-29
Disallow down migration if it was never migrated up

cb0:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #874

#### Summary
Before running a down migration script, the migration repository will be checked to see if it was ever executed. 
It the migration did not run, a Exception of type VersionDoesNotExist will be thrown.


goetas:
The whole propose of `execute`, is to run a migration independently if it was migrated or not.
I think that this PR should be closed.

goetas (changes requested):
The whole propose of execute, is to run a migration independently if it was migrated or not.
I think that this PR should be closed.

cb0:
@goetas Thanks for your review. 
I'm just beginning and might not see the whole picture.

But from my point of view, it is dangerous to "to run a migration independently if it was migrated or not". It should only be able to migrate down, if it was migrated up. Otherwise the database "could" result in an inconsistent state, depending on the content of the migration.

Or do I missunderstand something here ?

goetas:
You are perfectly right. it is dangerous, but this is the whole propose of this command, being able to run arbitrary migrations. Sometimes is useful sometimes is dangerous. 

What about changing the behavior of this command to not execute up/down the migration if it has not been executed in the opposite direction? and adding an `--force` option to override the check.

In this way we get a safe execute by default (the behavior you wanted), but it allows to override it when necessary.

goetas:
@cb0 let me know if you are interested to continue to work on this. Implementing it as suggested in https://github.com/doctrine/migrations/pull/900#issuecomment-573570567 will be possible only before tagging 3.0.

PR #908, opened 2020-01-13
Add List migrations command

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | -
| Fixed issues | https://github.com/doctrine/migrations/issues/734

#### Summary

as suggested in https://github.com/doctrine/migrations/issues/734, this PR adds a dedicated `list` command for listing migrations.

For now, this is just a draft, since is based on https://github.com/doctrine/migrations/pull/898. For a better diff, check https://github.com/doctrine/migrations/pull/908/commits/3f344d2466ddf74c52c2278d5786a5756b48cd92

PR #910, opened 2020-01-13
Introduce version comparator

goetas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | -
| Fixed issues | -

Closes https://github.com/doctrine/migrations/pull/902

#### Summary

This is an alternative implementation to https://github.com/doctrine/migrations/pull/902

The main change is that now the comparing function does not receive instances of `AvailableMigration` but `Version`. This change is needed if we want to be able to sort executed unavailable migrations. Being able to sort unavailable migrations is important to allow to implement properly https://github.com/doctrine/migrations/pull/908 by offering all the migrations in a single table instead of two separate.

### DIFF
--- a/lib/Doctrine/Migrations/DependencyFactory.php
+++ b/lib/Doctrine/Migrations/DependencyFactory.php
@@ -32,6 +32,8 @@ use Doctrine\Migrations\Tools\Console\ConsoleInputMigratorConfigurationFactory;
 use Doctrine\Migrations\Tools\Console\Helper\MigrationStatusInfosHelper;
 use Doctrine\Migrations\Tools\Console\MigratorConfigurationFactory;
 use Doctrine\Migrations\Version\AliasResolver;
+use Doctrine\Migrations\Version\AlphabeticalComparator;
+use Doctrine\Migrations\Version\Comparator;
 use Doctrine\Migrations\Version\CurrentMigrationStatusCalculator;
 use Doctrine\Migrations\Version\DbalExecutor;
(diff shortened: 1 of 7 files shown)


lib/Doctrine/Migrations/DependencyFactory.php
```diff
@@ -104,10 +104,10 @@ public function getConnection() : Connection
         return $this->connection;
     }
 
-    public function getSorter() : ?callable
+    public function getVersionComparator() : Comparator
     {
-        return $this->getDependency(self::MIGRATIONS_SORTER, static function () {
-            return null;
+        return $this->getDependency(Comparator::class, static function () {
```
greg0ire:
```suggestion
        return $this->getDependency(Comparator::class, static function () : AlphabeticalComparator {
```

lib/Doctrine/Migrations/Version/AlphabeticalComparator.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Migrations\Version;
+
+use function strcmp;
+
+class AlphabeticalComparator implements Comparator
```
greg0ire:
```suggestion
final class AlphabeticalComparator implements Comparator
```

greg0ire:
@pulsarraider please review

goetas:
@pulzarraider I'm merging this in order to be able to continue to work on https://github.com/doctrine/migrations/pull/908 .
We can continue the discussion here, on a merged PR.

PR #916, opened 2020-01-15
Fix migration timing/memory-usage display

thePanz:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

For single migrations, the timing and memory-consumption values shown in the CLI are about the overall migration process, not about the single migration that just completed.

Using the stopwatch periods, we can use the right values for the current migration.

### DIFF
--- a/tests/Doctrine/Migrations/Tests/Version/ExecutorTest.php
+++ b/tests/Doctrine/Migrations/Tests/Version/ExecutorTest.php
@@ -21,6 +21,7 @@ use Doctrine\Migrations\Version\Version;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Stopwatch\StopwatchEvent;
+use Symfony\Component\Stopwatch\StopwatchPeriod;
 
 class ExecutorTest extends TestCase
 {
@@ -212,12 +213,10 @@ class ExecutorTest extends TestCase
             $this->stopwatch
         );
 
-        $this->configuration->expects(self::any())
-            ->method('getConnection')
+        $this->configuration->method('getConnection')
             ->willReturn($this->connection);
 
-        $this->connection->expects(self::any())
-            ->method('getDatabasePlatform')
+        $this->connection->method('getDatabasePlatform')
             ->willReturn($this->createMock(AbstractPlatform::class));
 
         $this->version = new Version(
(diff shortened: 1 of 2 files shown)


dbu (approved):
looks right to me.

should we somehow output a summary at the end, with the overall duration and the peak memory usage (assuming that the stopwatch->getMemory is about peak memory)?

michellesanver (approved):
Looks good to me, exactly what we need =)

thePanz:
> looks right to me.
> 
> should we somehow output a summary at the end, with the overall duration and the peak memory usage (assuming that the stopwatch->getMemory is about peak memory)?

At the end of the whole migrations process, the command will show the total amount of memory used, plus the total amount of time spent running the migrations :)

tests/Doctrine/Migrations/Tests/Version/ExecutorTest.php
```diff
@@ -245,6 +246,18 @@ protected function setUp() : void
         $stopwatchEvent->expects(self::any())
             ->method('getMemory')
             ->willReturn(100);
+
+        $stopWatchPeriod = $this->createMock(StopwatchPeriod::class);
+        $stopWatchPeriod->expects(self::any())
```
greg0ire:
```suggestion
        $stopWatchPeriod
```

tests/Doctrine/Migrations/Tests/Version/ExecutorTest.php
```diff
@@ -245,6 +246,18 @@ protected function setUp() : void
...
+            ->method('getDuration')
+            ->willReturn(100);
+        $stopWatchPeriod->expects(self::any())
```
greg0ire:
```suggestion
        $stopWatchPeriod
```

tests/Doctrine/Migrations/Tests/Version/ExecutorTest.php
```diff
@@ -245,6 +246,18 @@ protected function setUp() : void
...
+            ->method('getMemory')
+            ->willReturn(100);
+
+        $stopwatchEvent->expects(self::any())
```
greg0ire:
```suggestion
        $stopwatchEvent
```

alcaeus:
@greg0ire would you consider this a bugfix? If so, please change the label. If not I suggest targeting 2.3.x for this and ensuring that the changes get merged to master properly, as the 3.0 changes may make the entire process significantly different.

dbu:
if the current output is the aggregated time since the first migration, i would consider that a bug.

alcaeus:
Bug it is then. Easier to change the label than to rebase :P

tests/Doctrine/Migrations/Tests/Version/ExecutorTest.php
thePanz: I quite agree with this change, but why changing it here only, and not in all other places? Like the existing `$stopwatchEvent->expects(self::any())`?

greg0ire: Feel absolutely free to :), but in a separate commit please.

greg0ire (approved):

tests/Doctrine/Migrations/Tests/Version/ExecutorTest.php
thePanz: Sorry, I amended my previous commit :(

greg0ire:
Thanks @thePanz !

PR #917, opened 2020-01-15
Extended details on up to date migrations

goetas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | -
| Fixed issues | https://github.com/doctrine/migrations/pull/886

#### Summary

alternative version to https://github.com/doctrine/migrations/pull/886

### DIFF
--- a/lib/Doctrine/Migrations/DependencyFactory.php
+++ b/lib/Doctrine/Migrations/DependencyFactory.php
@@ -401,7 +401,9 @@ class DependencyFactory
             return new MigrationStatusInfosHelper(
                 $this->getConfiguration(),
                 $this->getConnection(),
-                $this->getVersionAliasResolver()
+                $this->getVersionAliasResolver(),
+                $this->getMigrationRepository(),
+                $this->getMetadataStorage()
             );
         });
     }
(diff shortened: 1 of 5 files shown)


goetas:
@dbu can you have a look at this?

goetas:
i'm exploring the idea to use the `-vv` option (verbosity level) to instead of `--extended` to decide if print or not the extra table... what do you think?

(if accepted, this could be applied to all the other commands...)

dbu:
if using verbosity, i would use single -v, otherwise there might be too much noise from symfony debugging things. it sounds kind of natural to do that, but then again i don't know if there are other things reacting to the added verbosity. and having an explicitly named parameter would make it easier to spot the feature, though the description of the command could mention -v. another option could be to print by default but not with -q. 

really not sure which is better, but don't think its terribly important either. maybe check if -v (or -vv) prints much extra stuff in a vanilla symfony installation, and if not, go with that?

alcaeus:
If I understand it correctly, the current `extended` option shows a detailed table of all migrations instead of just saying "up-to-date" or "out-of-date"? If so, I would avoid using the verbosity switches (either `-q` or `-v`) for this, as those imply that they print more or less debug output. Maybe `--show-migration-info` or `--list-migrations` or something similar can be used to better convey what the setting does. `--extended` doesn't quite say what's going to change if I use the option.

goetas:
@alcaeus @dbu thank for the feedback! i have implemented it now using `--list-migrations` since the output can contain both unexecuted and missing migrations.

PR #918, opened 2020-01-17
do not define finders final so users may add code without copying

mercuriev:


alcaeus:
Why would people need to add code to them? What’s the use-case here?

mercuriev:
As part of integration to my existing App I wanted to load SQL from files (legacy), but I still wanted to be able to use generators.
So I wrote my own finder that complements the shipped one.

```php
class SqlFileFinder extends RecursiveRegexFinder
{
    const PATTERN = '*.sql';

    public function findMigrations(string $directory, ?string $namespace = null) : array
    {
        // here Doctrine finds classes
        $versions = parent::findMigrations($directory, $namespace);

        // here we pass doctrine object it works with for later execution
        $files = glob($directory . '/' . static::PATTERN);
        $versions += $this->loadSqlMigrations($files, $namespace);

        ksort($versions, SORT_STRING);

        return $versions;
    }

    protected function loadSqlMigrations(array $files, ?string $namespace) : array
    {
        $versions = [];
        foreach ($files as $file) {
            $version = explode('.', basename($file))[0];
            $versions[$version] = SqlFileMigration::class;
        }

        return $versions;
    }
}
```

It is a case of useful inheritance when you enhance existing parent code with your own.

greg0ire:
You can achieve the same by using composition:
```php
final class SqlFileFinder implements MigrationFinder
{
    const PATTERN = '*.sql';

    private  RecursiveRegexFinder $wrapped;

    public function __construct(RecursiveRegexFinder $finder)
    {
         $this->wrapped = $wrapped;
    }


    public function findMigrations(string $directory, ?string $namespace = null) : array
    {
        // here Doctrine finds classes
        $versions = $this->wrapped->findMigrations($directory, $namespace);

        // here we pass doctrine object it works with for later execution
        $files = glob($directory . '/' . static::PATTERN);
        $versions += $this->loadSqlMigrations($files, $namespace);

        ksort($versions, SORT_STRING);

        return $versions;
    }

    private function loadSqlMigrations(array $files, ?string $namespace) : array
    {
        $versions = [];
        foreach ($files as $file) {
            $version = explode('.', basename($file))[0];
            $versions[$version] = SqlFileMigration::class;
        }

        return $versions;
    }
}
```

mercuriev:
I could, but to me inheritance is shorter and matches intention better.
`SqlFileFinder` **is** the `RecursiveRegexFinder` finder with all its properties and code, plus the extended feature applicable to current use.

If `RecursiveRegexFinder` would have protected methods instead of private, I would be able to _reuse_ for example `createIterator` method to enable regex instead of simple `*.sql` glob oneliner I wrote. 
Either copypaste or you won't have this already configured iterator no no. Why?

greg0ire:
Here is why:

- https://ocramius.github.io/blog/when-to-declare-classes-final/
- https://fabien.potencier.org/pragmatism-over-theory-protected-vs-private.html

alcaeus:
Closing here. Why it may be shorter to use inheritance, it creates additional use-cases for us to think about when modifying the class. By closing the class for extension, we limit the use-case of the class to what we designed. As shown above, you can use composition to achieve what you want without implementing against a specific implementation.

PR #919, opened 2020-01-18
Introduce query value object to wrap sql queries

goetas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | -
| Fixed issues | -

#### Summary
I think that the `Query` value object was a great idea proposed by @pulzarraider in #906. Independently if #906 is going to be merged, I see value in the `Query` part, so here is a full refactoring of the code base to use it everywhere instead of the current 3 different arrays holding the same data.

### DIFF
--- a/lib/Doctrine/Migrations/Query/Query.php
+++ b/lib/Doctrine/Migrations/Query/Query.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace Doctrine\Migrations\Query;
 
+use Doctrine\Migrations\Query\Exception\InvalidArguments;
+use function count;
+
 /**
  * The Query wraps the sql query, parameters and types.
  */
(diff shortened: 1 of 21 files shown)


lib/Doctrine/Migrations/Query/Query.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Migrations\Query;
+
+/**
+ * The Query wraps the sql query, parameters and types.
```
greg0ire:
```suggestion
 * The Query wraps the sql statement, parameters and types.
```

lib/Doctrine/Migrations/Query/Query.php
```diff
@@ -0,0 +1,53 @@
...
+
+    /** @var mixed[] */
+    private $types;
+
+    /**
+     * @param mixed[] $parameters
+     * @param mixed[] $types
+     */
+    public function __construct(string $statement, array $parameters = [], array $types = [])
+    {
```
greg0ire: What should happen if the number of parameters and the number of types is not the same?

goetas: Good point. Will check what is doing doctrine DBAL, and do something similar

goetas: done

PR #924, opened 2020-01-28
Remove trailing spaces from generated SQL

julienfalque:
|      Q       |   A
|------------- | -----------
| Type         | improvement?
| BC Break     | no
| Fixed issues | -

When generating a new version using the diff command with `--formatted` option, the SQL formatted by the `jdorn/sql-formatter` lib contains trailing spaces. This PR makes the internal SQL generator removes them.

A better solution would probably be to do the same PR on `jdorn/sql-formatter` but the package seems abandoned.

### DIFF
--- a/tests/Doctrine/Migrations/Tests/Generator/SqlGeneratorTest.php
+++ b/tests/Doctrine/Migrations/Tests/Generator/SqlGeneratorTest.php
@@ -9,8 +9,6 @@ use Doctrine\Migrations\Configuration\Configuration;
 use Doctrine\Migrations\Generator\SqlGenerator;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
-use SqlFormatter;
-use function sprintf;
 
 final class SqlGeneratorTest extends TestCase
 {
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Migrations/Tests/Generator/SqlGeneratorTest.php
```diff
@@ -103,7 +103,28 @@ private function prepareGeneratedCode(string $expectedCode) : string
             'SELECT * FROM migrations_table_name',
         ];
 
-        $formattedUpdate = SqlFormatter::format($this->sql[2], false);
+        if (Versions::getVersion('jdorn/sql-formatter') === 'v1.1.0@e3083aafdf447af4abb47342158f6778e03c6d4b') {
```
goetas: Why target this particular version?

julienfalque: It's the version installed when using Composer `--prefer-lowest` option on Travis CI.

goetas: Shouldn't this work with all the other versions?

julienfalque: It should, but having different versions of `jdorn/sql-formatter` means different possible formattings for the SQL queries, so `$expected` must take that into account.

Though instead of checking the version of `jdorn/sql-formatter` I could make `$expected` a regular expression so that it accepts several SQL formattings while still rejecting trailing spaces, WDYT?

julienfalque: I replaced the `jdorn/sql-formatter` version check with regular expressions.

greg0ire:
Please RT: https://twitter.com/greg0ire/status/1222949210208817153

goetas:
@greg0ire @alcaeus I'm for merging this as https://github.com/doctrine/migrations/pull/924#issuecomment-580392424 did not bring any result.

What do you think?

greg0ire:
I think we should probably fork the lib, possibly under the Doctrine organization. We have a dependency on it, and that's unhealthy.

greg0ire:
FYI the forking is underway: https://github.com/doctrine/sql-formatter

greg0ire:
Closing in favor of https://github.com/doctrine/sql-formatter/issues/29

PR #925, opened 2020-01-31
Extract migration factory interface

goetas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | -
| Fixed issues | -

#### Summary

Extract migration factory interface

### DIFF
--- a/lib/Doctrine/Migrations/DependencyFactory.php
+++ b/lib/Doctrine/Migrations/DependencyFactory.php
@@ -37,6 +37,7 @@ use Doctrine\Migrations\Version\AlphabeticalComparator;
 use Doctrine\Migrations\Version\Comparator;
 use Doctrine\Migrations\Version\CurrentMigrationStatusCalculator;
 use Doctrine\Migrations\Version\DbalExecutor;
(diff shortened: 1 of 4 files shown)


lib/Doctrine/Migrations/DependencyFactory.php
```diff
@@ -281,12 +282,19 @@ public function getMigrationRepository() : MigrationRepository
...
                 $this->getConfiguration()->getMigrationDirectories(),
                 $this->getMigrationsFinder(),
-                new MigrationFactory($this->getConnection(), $this->getLogger()),
+                $this->getMigrationFactory(),
                 $this->getVersionComparator()
             );
         });
     }
 
+    public function getMigrationFactory() : MigrationFactory
```
SenseException: Is it necessary that this method is public?

goetas: good point, fixed

goetas: sorry, i have re exposed it again. users need to fetch the old migration factory and being able to decorate it with their own implementation

greg0ire (approved):

PR #926, opened 2020-01-31
Allow to decorate dependency factory services

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | -
| Fixed issues | -

#### Summary

The DI factory become a central extension point for doctrine migrations. 
Being able to decorate services is fundamental for most of DI containers and allows to apply composition over inheritance


goetas:
thinking it twice, this is already possible by doing

```php

$oldComparator = $di->getComparator();
$oldComparator->setComparator(new NewComparator($oldComparator));

```

PR #927, opened 2020-02-04
fix logger

rodion-k:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

Logging not work in master branch. According to [this line](https://github.com/doctrine/migrations/blob/f19a9f1/lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php#L89) in DoctineCommand class logger is set as service. But that service was not used.


alcaeus:
Closing in favour of #928.

PR #929, opened 2020-02-09
TableMetadataStorage uses the repository to provide migrations for ta…

goetas:
…ble schema

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes/no
| Fixed issues | 

#### Summary
without this dependency, the migration of the table structure from v2 to w3 wont work

### DIFF
--- a/lib/Doctrine/Migrations/DependencyFactory.php
+++ b/lib/Doctrine/Migrations/DependencyFactory.php
@@ -319,7 +319,8 @@ class DependencyFactory
         return $this->getDependency(MetadataStorage::class, function () : MetadataStorage {
             return new TableMetadataStorage(
                 $this->getConnection(),
-                $this->getMetadataStorageConfiguration()
+                $this->getMetadataStorageConfiguration(),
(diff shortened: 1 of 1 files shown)


greg0ire:
Not super important for this PR, but please try to make commit that follow the rules in [the contributing guide](https://www.doctrine-project.org/contribute/index.html#working-on-topic-branches).

For instance, it's good to avoid having the PR title/commit message subject split into 2 with `…`, and it's really great to include useful explanations such as "without this dependency, the migration of the table structure from v2 to w3 wont work" directly in the body of the commit message, that way the changes can be understood without browsing Github.

goetas:
> Not super important for this PR, but please try to make commit that follow the rules in the contributing guide.
For instance, it's good to avoid having the PR title/commit message subject split into 2 with …, and it's really great to include useful explanations such as "without this dependency, the migration of the table structure from v2 to w3 wont work" directly in the body of the commit message, that way the changes can be understood without browsing Github.

Thanks for the feedback, will try to make it better next time!

PR #932, opened 2020-02-21
Bugs in 3.0-dev

goetas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | -
| Fixed issues | -

#### Summary
I've tested the master branch on a couple of real projects, this are ~5~ 3 independent bugs i've found, one for each commit

~All the commits include tests except of https://github.com/doctrine/migrations/pull/932/commits/66b13665515fc0eab636ea8a486b93b71d9e1e31. What happens if we want to do dry-run on an empty database? should the metadata table be created anyway?~

### DIFF
--- a/lib/Doctrine/Migrations/Version/DbalExecutor.php
+++ b/lib/Doctrine/Migrations/Version/DbalExecutor.php
@@ -190,7 +190,7 @@ final class DbalExecutor implements Executor
         $migration->{'post' . ucfirst($direction)}($toSchema);
         $stopwatchEvent->stop();
 
-        $result->setTime((float) $stopwatchEvent->getDuration());
+        $result->setTime((float) $stopwatchEvent->getDuration()/1000);
         $result->setMemory($stopwatchEvent->getMemory());
 
         $params = [
--- a/lib/Doctrine/Migrations/Version/DefaultAliasResolver.php
+++ b/lib/Doctrine/Migrations/Version/DefaultAliasResolver.php
@@ -9,7 +9,6 @@ use Doctrine\Migrations\Exception\NoMigrationsToExecute;
 use Doctrine\Migrations\Exception\UnknownMigrationVersion;
 use Doctrine\Migrations\Metadata\Storage\MetadataStorage;
 use Doctrine\Migrations\MigrationRepository;
-use function count;
 use function substr;
 
 /**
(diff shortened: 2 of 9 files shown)


SenseException:
Can you please describe the 5 Bugs this PR is covering? Why did you decide to handle 5 bugs with one PR?

goetas:
- d235583 - ExecutionResult time is expressed in seconds (in some parts of the code the time was in seconds and sometimes in milliseconds, thus because of conversions, in the list of executed migrations, the displayed time was 1000 times longer)  
- b108ac9 - The first migration is always the migration "0" (as the description says, "first" should be the state when there are no migrations, previously was the first executed migration)
- fe628ad - Make target plan version argument mandatory (this is a minor change that allows to have less code duplication)
- fe628ad Do not display the project name twice via status command (in the migration status command, the project name is displayed twice, this could be not needed if https://github.com/doctrine/migrations/pull/933 gets merged)
 - 66b1366 - Initialize metadata when running migrations (this is controversial. What happens if we want to do dry-run on an empty database? should the metadata table be created anyway? will have to verify how 2.x behaves)

goetas:
For an easier review and a quicker merge, I have removed the last commit. The metadata initialization will be fixed in a separate PR were will be easier to review the solution and check the tests.

goetas:
Updated PR. only 3 bugs here now.

greg0ire:
> d235583 - ExecutionResult time is expressed in seconds (in some parts of the code the time was > in seconds and sometimes in milliseconds, thus because of conversions, in the list of executed 
> migrations, the displayed time was 1000 times longer)
> b108ac9 - The first migration is always the migration "0" (as the description says, "first" should be > the state when there are no migrations, previously was the first executed migration)
> fe628ad - Make target plan version argument mandatory (this is a minor change that allows to 
> have less code duplication)

Please include these explanations in the commit messages.

> ExecutionResult time is expressed in seconds

Is that the buggy behavior you want to fix, or is that the correct behavior you enforced

> The first migration is always the migration "0"

Same question, here, I think it's the correct one, but I'd like to be sure.

Tip: complete the sentence `If applied, this commit will` to make your commit message subject, it will avoid that kind of confusion.

goetas:
@greg0ire thanks for the review. I have pushed the updated commit messages

greg0ire:
Thanks @goetas !

PR #933, opened 2020-02-21
Drop application name as it has no real use case

goetas:
|      Q       |   A
|------------- | -----------
| Type         | rfc
| BC Break     | -
| Fixed issues | -

#### Summary

currently the "application name" has no real propose and people (as I know...) always is set to some default. i suggest to simply remove it


goetas:
@alcaeus @greg0ire @SenseException  what do you think about this?

SenseException:
Judging by the diff it doesn't seem to have an impact on migrations when the application name is removed. Are there any changes needed for the documentation if this PR would get an okay?

alcaeus:
> what do you think about this?

No point keeping it around just to output it when a command is run. I'm ok with dropping it as long as we deprecate it first ;)

goetas:
Can someone approve it? (@alcaeus  the forward compatibility chapter is still open for 2.x and would like to handle it once all the changes in 3.x are merged)

greg0ire:
What do you mean by "chapter"?

goetas:
I mean that 3.x introduced BC breaks, and none of them have the deprecation layer in place for now inside the 2.x branch

goetas:
Will do my best with that. I think that not too many people have used doctrine/migrations as a library, but just as dependency

Issue #930, opened 2020-02-11
Display a warning if running diff command with previous non-executed migrations


VincentLanglet:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

Related to https://github.com/doctrine/DoctrineMigrationsBundle/issues/177

Running the diff command should display a warning if not in sync with the current migrations is welcome.

goetas:
sounds a good idea.

goetas:
Are you interested in implementing this feature?

VincentLanglet:
@goetas Why not, but I don't know the project so If you know where I should look at, I take it.

goetas:
done in https://github.com/doctrine/migrations/pull/948

PR #941, opened 2020-03-08
Add warning for diffCommand

VincentLanglet:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | #930 

#### Summary

Add warning when running diff command with executed unavalaible or non executed migrations.

### DIFF
--- a/tests/Doctrine/Migrations/Tests/Tools/Console/Command/DiffCommandTest.php
+++ b/tests/Doctrine/Migrations/Tests/Tools/Console/Command/DiffCommandTest.php
@@ -4,12 +4,18 @@ declare(strict_types=1);
 
 namespace Doctrine\Migrations\Tests\Tools\Console\Command;
 
+use Doctrine\Migrations\AbstractMigration;
 use Doctrine\Migrations\Configuration\Configuration;
 use Doctrine\Migrations\DependencyFactory;
 use Doctrine\Migrations\Generator\ClassNameGenerator;
 use Doctrine\Migrations\Generator\DiffGenerator;
+use Doctrine\Migrations\Metadata\AvailableMigration;
+use Doctrine\Migrations\Metadata\AvailableMigrationsList;
+use Doctrine\Migrations\Metadata\ExecutedMigration;
+use Doctrine\Migrations\Metadata\ExecutedMigrationsSet;
 use Doctrine\Migrations\Tools\Console\Command\DiffCommand;
 use Doctrine\Migrations\Version\MigrationStatusCalculator;
+use Doctrine\Migrations\Version\Version;
 use PHPUnit\Framework\MockObject\MockObject;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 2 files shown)


goetas:
I think that you should be targeting the master branch

goetas:
That seems to be the reason why tests are failing

VincentLanglet:
@goetas I fix some stuffs, but I don't understand the static analysis failure.

tests/Doctrine/Migrations/Tests/Tools/Console/Command/DiffCommandTest.php
```diff
@@ -8,6 +8,7 @@
 use Doctrine\Migrations\DependencyFactory;
 use Doctrine\Migrations\Generator\ClassNameGenerator;
 use Doctrine\Migrations\Generator\DiffGenerator;
+use Doctrine\Migrations\MigrationPlanCalculator;
```
alcaeus: This should fix static analysis:

```suggestion
use Doctrine\Migrations\Version\MigrationPlanCalculator;
```

You can run phpstan locally by running `vendor/bin/phpstan analyse` to see if it worked without pushing.

VincentLanglet:
@alcaeus Thanks. I should be tired...

VincentLanglet:
@goetas @alcaeus Phpstan is fixed now. Is the Scrutinizer check required ?

SenseException:
I've restarted the check and it's green now.

goetas:
I'm not sure if we need that many warnings... I was just thinking about adding a a warning but keep the flow exactly as before. I'm this pr in the worst case we as to the user confirmation for two times...

Opinions?

VincentLanglet:
> I'm not sure if we need that many warnings... I was just thinking about adding a a warning but keep the flow exactly as before. I'm this pr in the worst case we as to the user confirmation for two times...

IMO, you barely never want to create a new migration with non executed ones or executed non available one because it will generate some conflict between the migrations.

The goal of adding these warning was to prevent the developper from messing up BEFORE the migrations was created.

If you look at the issue https://github.com/doctrine/DoctrineMigrationsBundle/issues/177, you'll have opinion. They wanted to not allow running diff command when not in sync, or to run migrate before diff. This is going in my way.

goetas:
I'm convinced then. Looks good to me.

Can you add tests please for this feature?

VincentLanglet:
@goetas Tests added

SenseException:
I'd like to see the `true` case of these two new methods covered too. You can have an early exit after their if-blocks are done by having the empty diff functionality.

VincentLanglet:
@SenseException Sorry, I don't understand what you mean by the `true` case ?
There was already a test for this command. I added the case with non-executed migration or executed-non-available one.

goetas:
Please see #948

VincentLanglet:
@goetas Should I close my PR in favor of yours ?

goetas:
yeah, I suggest to move the discussion to https://github.com/doctrine/migrations/pull/948

PR #944, opened 2020-03-12
Mark console commands as final classes

goetas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | -
| Fixed issues | 

#### Summary

Mark console commands as final to avoid extension. Console commands are not meant for that propuse.

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
@@ -15,7 +15,6 @@ use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\ConfirmationQuestion;
-use Symfony\Component\Process\Process;
 use function is_string;
 
 /**
@@ -102,14 +101,5 @@ abstract class DoctrineCommand extends Command
         OutputInterface $output
     ) : bool {
         return ! $input->isInteractive() || $this->askConfirmation($question, $input, $output);
-    }
-
-    protected function procOpen(OutputInterface $output, string $editorCommand, string $path) : void
-    {
-        $process = new Process([$editorCommand, $path]);
-        $process->setTty(true);
-        $this->getHelper('process')->mustRun($output, $process, null, static function ($type, $buffer) : void {
-            echo $buffer;
-        });
     }
 }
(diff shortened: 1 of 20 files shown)


goetas:
I really don't understand why scrutinizer started to complain lately

greg0ire:
Do you mean appveyor?

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
```diff
@@ -105,8 +104,12 @@ protected function canExecute(
         return ! $input->isInteractive() || $this->askConfirmation($question, $input, $output);
     }
 
-    protected function procOpen(string $editorCommand, string $path) : void
+    protected function procOpen(OutputInterface $output, string $editorCommand, string $path) : void
     {
-        proc_open($editorCommand . ' ' . escapeshellarg($path), [], $pipes);
+        $process = new Process([$editorCommand, $path]);
+        $process->setTty(true);
```
greg0ire: I guess this must be why Appveyor is failing

goetas:
> Do you mean appveyor?

did not realize was appveyor... in the last PR scritinizer complained about something... but it stopped now...

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
goetas: Do you know if the feature of opening an editor works/worked at all on windows systems?

(or was simply never tested...)

If so, can we simply skip the test?

greg0ire: No. (no). Probably, I don't know why an editor is involved here TBH.

goetas: What about dropping the possibility of opening an editor from doctrine migrations? I think that is out of scope for such library...

greg0ire: Had no idea this was a thing… what is it used for?

goetas: When you run the diff or generate command, and pass `--editor-cmd=vi`, it will open the generated migration file with vi

greg0ire: Oh ok… well I know I wouldn't use it, because I don't like having many instances of vim running for the same project. In favor of dropping that feature.

goetas: @alcaeus what do you think?

alcaeus: I've never used it either. If it's causing problems, drop it. If it's easier to skip something on Windows, do that. I don't have any preference here ;)

goetas: I will drop it then, as it will allow us to not depend on the `symfony/process` component.

goetas:
I think that this is ready. 

(still do not know what causes scrutinizer to fail)

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
goetas: done, removed

PR #949, opened 2020-03-22
Upgrade dependencies

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a
#### Summary

The main goals are to be able to use `createStub`, and PHPUnit 9.


composer.json
```diff
@@ -24,12 +25,12 @@
...
-        "phpstan/phpstan-phpunit": "^0.11",
-        "phpstan/phpstan-strict-rules": "^0.11",
-        "phpstan/phpstan-symfony": "^0.11.6",
-        "phpunit/phpunit": "^8.3",
+        "phpstan/phpstan": "^0.12",
+        "phpstan/phpstan-deprecation-rules": "^0.12",
+        "phpstan/phpstan-phpunit": "^0.12",
+        "phpstan/phpstan-strict-rules": "^0.12",
+        "phpstan/phpstan-symfony": "^0.12",
+        "phpunit/phpunit": "^9.0",
```
goetas: `9.0` requires `php 7.3`, i guess we will have to put `^8.0|^9.0`

greg0ire: Since we put the `composer.lock` under version control, we might as well use `^8.0` here…

lib/Doctrine/Migrations/Finder/Finder.php
```diff
@@ -78,7 +78,7 @@ protected function loadMigrations(array $files, ?string $namespace) : array
      * @param string[]    $files     The set of files that were `required`
      * @param string|null $namespace If not null only classes in this namespace will be returned
      *
-     * @return ReflectionClass[] the classes in `$files`
+     * @return ReflectionClass<object>[] the classes in `$files`
```
goetas: does this syntax work in phpstorm?

lib/Doctrine/Migrations/InlineParameterFormatter.php
```diff
@@ -91,7 +91,5 @@ private function parameterToString($value) : string
         if (is_bool($value)) {
             return $value === true ? 'true' : 'false';
         }
-
```
goetas: any particular reason for removing this?

lib/Doctrine/Migrations/Finder/Finder.php
greg0ire: I wouldn't know, I use vim

goetas: :)

lib/Doctrine/Migrations/InlineParameterFormatter.php
greg0ire: Yes, it's unreachable according to the docblock. If I leave it, phpstan complains with `Unreachable statement - code above always terminates.`

goetas: :+1:

PR #162, opened 2014-03-24
[WIP] 2.0 Prototype

beberlei:
Pull-Request for Discussion of Doctrine Migrations 2.0 Prototype. Text is copied from new ``README.md``.

Please view the rewrite from the branches screen, the Diff isn't very valuable: https://github.com/doctrine/migrations/tree/Rewrite

Starting points are:

- https://github.com/doctrine/migrations/blob/Rewrite/tests/Doctrine/Tests/Migrations/DBAL/IntegrationTest.php
- https://github.com/doctrine/migrations/blob/Rewrite/lib/Doctrine/Migrations/Migrations.php

Migrations helps developers migrate database schema and data in a simple and reproducible way.
It is currently in alpha version after a full rewrite of its precedessor, version 1.0 that
never left alpha status.

The goal of the rewrite is to reach a stable version as soon as possible,
hopefully leaving the long period of rather unstable migrations in Doctrine
behind us.

The current verion aims to solve various problems:

- Remove reliance on Schema API of DBAL at runtime, because it proved to be a source
  of bugs and when executed with reproduceability in mind.
- Remove ability to migrate down. Downmigrations are not possible for 100% of up cases
  and you can therefore not rely on this feature to work. Providing this feature
  only adds a false sense of safety that can bite you hard.
- Remove ability to generate migrations by diff. This is not a feature Migrations
  should provide, instead it only works when you have two schemas present like
  in the ORM and should be provided there.
- Introduce SQL file based migrations as default for the 80% use-case, leaving
  PHP based migrations for only complex scenarios.
- Improve ability to check executed SQL beforehand (dry-run).
- Improve information about migrations
- Add explicit handling for failed migrations
- Simplify code-base and reduce code complexity, decouple core from DBAL and Symfony Console.
  This should improve integration of Migrations into other libraries and frameworks.

## Todos

- Add Symfony Console and PHAR Generation
- Tests for migration failures
- Dry-Run mode
- Multi-Platform migrations (with subdirectory SqlFileLoader)
- Documentation

## Introduction

Doctrine Migrations updates a database schema from one version to another using
a concept called migration.  Each migration has a version and a description.
Versions **MUST** be unique.

From a configured directory Doctrine loads SQL- and PHP-based migrations, where
SQL is used for simple tasks and PHP can be used for more complex operations.

Doctrine sorts all migrations by version and executes them in order.

Migration files have to be in the following format or an error will be thrown:

    V[version]_[description].[ext]

Valid filenames include:

[...]

PR #952, opened 2020-03-28
Make the down() migration optional

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | -
| Fixed issues | -

#### Summary
Removing the `down()` migration is something that has been already proposed in https://github.com/doctrine/migrations/pull/162 , https://github.com/goetas/migrations/pull/2 . 
Each time the decision has been postponed as some people still use that feature while others don't.

This PR proposes to make it optional. It provides a default implementation in `AbstractMigration` but still allows its use as it was before (the default migration generation works as usual by generating the `down()` method as expected).

### DIFF
--- a/tests/Doctrine/Migrations/Tests/Stub/AbstractMigrationWithoutDownStub.php
+++ b/tests/Doctrine/Migrations/Tests/Stub/AbstractMigrationWithoutDownStub.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Migrations\Tests\Stub;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+class AbstractMigrationWithoutDownStub extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+    }
+}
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Migrations/Tests/Stub/AbstractMigrationWithoutDownStub.php
```diff
@@ -0,0 +1,15 @@
...
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+class AbstractMigrationWithoutDownStub extends AbstractMigration
+{
+    public function up(Schema $schema) : void
+    {
+    }
+}
```
greg0ire: I think that kind of stub is seldom reused. Have you considered using an anonymous class for this?

goetas: yes, that was my first attempt, but was hard to test this https://github.com/doctrine/migrations/pull/952/files#diff-e75b76ba40ceb8b96f42333c2ca4dcdcR33 as the class name was variable.

goetas: Something like `No down() migration implemented for "class@anonymous/home/goetas/projects/doctrine-migrations/tests/Doctrine/Migrations/Tests/AbstractMigrationTest.php0x7f7a5aafa468"`

PR #954, opened 2020-03-30
Make compared tables order idempotent

julienfalque:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

This PR makes sure the schema generated from mapping (to be compared with the actual database) is always provided tables data in the same order.

This helps when you want to have only one migration per deployment: instead of creating several small migration versions, you create a single one that you remove and recreate each time you change the mapping. For some reasons I couldn't figure out, it happens that the order in which tables data are loaded can vary. In my case it was different from one developper setup to another (despite all using the same Docker setup). In such case, even adding a simple column can generate a completely different migration, which makes it hard to read in VCS history and code reviews.

### DIFF
--- a/lib/Doctrine/Migrations/Provider/OrmSchemaProvider.php
+++ b/lib/Doctrine/Migrations/Provider/OrmSchemaProvider.php
@@ -7,8 +7,10 @@ namespace Doctrine\Migrations\Provider;
 use Doctrine\DBAL\Schema\Schema;
 use Doctrine\Migrations\Provider\Exception\NoMappingFound;
 use Doctrine\ORM\EntityManagerInterface;
+use Doctrine\ORM\Mapping\ClassMetadata;
 use Doctrine\ORM\Tools\SchemaTool;
 use function count;
+use function usort;
 
 /**
  * The OrmSchemaProvider class is responsible for creating a Doctrine\DBAL\Schema\Schema instance from the mapping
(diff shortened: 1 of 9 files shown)


lib/Doctrine/Migrations/Provider/OrmSchemaProvider.php
```diff
@@ -38,6 +41,14 @@ public function createSchema() : Schema
             throw NoMappingFound::new();
         }
 
+        usort($metadata, static function (ClassMetadata $a, ClassMetadata $b) {
+            if (method_exists($a, 'getTableName') && method_exists($b, 'getTableName')) {
```
julienfalque: PHPStan reports undefined method calls here, which looks like a false positive to me since there are `method_exists()` checks. Though I'm happy to remove this `if` entirely if you think it's best.

alcaeus: That's because you used the `ClassMetadata` type from doctrine/persistence which doesn't have the `getTableName` methods. Since this is always in the context of ORM, you can use `Doctrine\ORM\Mapping\ClassMetadata` instead.

julienfalque: Done.

goetas:
Travis build was successful (see https://travis-ci.org/github/doctrine/migrations/builds/669306030).

goetas:
@julienfalque can you please rebase ? (to trigger a new build... i think that travis had some issues...)

julienfalque:
Sure :)

goetas:
Thank you!

PR #959, opened 2020-04-08
Allow to change configurations in the dependency factory

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature (re introduce feature removed in previous pull  requirests) 
| BC Break     | -
| Fixed issues | https://github.com/doctrine/DoctrineMigrationsBundle/issues/311

#### Summary

Currently the Dependency Factory contains already all the configurations.
This allows to change configurations (not db connection or entity manager) even if they are already specified in the Dependency Factory


lib/Doctrine/Migrations/DependencyFactory.php
```diff
@@ -136,6 +136,15 @@ public function hasEntityManager() : bool
         return $this->emLoader !== null;
     }
 
+    public function setConfigurationLoader(ConfigurationLoader $configurationLoader) : void
+    {
+        if ($this->configuration !== null) {
+            throw FrozenDependencies::new();
```
goetas, line 142: Here a better error message will be provided

SenseException: Should there be also a check if the configuration is frozen? Being `!== null` doesn't mean it's already frozen.

goetas:
Functionality wise this has been tested in https://github.com/doctrine/DoctrineMigrationsBundle/issues/311#issuecomment-617661913

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
```diff
@@ -63,6 +63,9 @@ protected function initialize(InputInterface $input, OutputInterface $output) :
             );
             $connectionLoader        = new ConfigurationFile((string) $input->getOption('db-configuration'));
             $this->dependencyFactory = DependencyFactory::fromConnection($configurationLoader, $connectionLoader);
+        } elseif (is_string($input->getOption('configuration'))) {
```
SenseException, line 66: Can you please call `$input->getOption('configuration')` once and use a variable?

tests/Doctrine/Migrations/Tests/DependencyFactoryTest.php
```diff
@@ -91,6 +91,7 @@ public function testConnection() : void
 
         self::assertSame($this->connection, $di->getConnection());
         self::assertFalse($di->hasEntityManager());
+        self::assertTrue($di->isFrozen());
```
SenseException, line 94: Testing explicitly if the `DependencyFactory`. Does this mean that in the other tests it should be checked if it isn't frozen in case this is relevant for regressions?

goetas:
@SenseException both notes you've left have been addressed in the last two commits

PR #968, opened 2020-04-22
Use SymfonyStyle to output text in console

emodric:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

This adds a bit of style to console commands. Main visual differences are whitespaces, with occasional change in coloring.

I can add some example screenshots with before and after if needed.

Still needs tests.

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
@@ -14,7 +14,8 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Console\Question\ConfirmationQuestion;
+use Symfony\Component\Console\Style\StyleInterface;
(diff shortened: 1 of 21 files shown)


goetas:
Thanks a lot for this contribution. This is something I would like to have for 3.0! I think that screenshots are not needed.

Fixing tests for sure is necessary.  :)

emodric:
@goetas I have updated the tests. I did have to increase the terminal width on Travis for the tests to pass since SymfonyStyle obviously wraps the output if the display is too small, causing output assertions to fail.

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
```diff
@@ -25,6 +26,9 @@ abstract class DoctrineCommand extends Command
     /** @var DependencyFactory|null */
     private $dependencyFactory;
 
+    /** @var StyleInterface */
+    protected $io;
```
stof, line 30: should be private

emodric: Can't be private. It is used by other commands which extend from `DoctrineCommand`.

goetas:
Thanks

PR #971, opened 2020-04-28
Migrate to Doctrine's fork of jdorn/sql-formatter

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #924 

#### Summary

We like our dependencies maintained, and `jdorn/sql-formatter` sadly no
longer is. I am targeting master, because `doctrine/sql-formatter` requires php 7.2

### DIFF
--- a/lib/Doctrine/Migrations/Generator/SqlGenerator.php
+++ b/lib/Doctrine/Migrations/Generator/SqlGenerator.php
@@ -7,7 +7,8 @@ namespace Doctrine\Migrations\Generator;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\Migrations\Configuration\Configuration;
 use Doctrine\Migrations\Metadata\Storage\TableMetadataStorageConfiguration;
-use SqlFormatter;
+use Doctrine\SqlFormatter\NullHighlighter;
+use Doctrine\SqlFormatter\SqlFormatter;
 use function array_unshift;
 use function count;
 use function implode;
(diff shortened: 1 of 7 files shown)


lib/Doctrine/Migrations/Generator/SqlGenerator.php
```diff
@@ -30,10 +31,14 @@ class SqlGenerator
...
     private $platform;
 
+    /** @var SqlFormatter */
+    private $sqlFormatter;
+
     public function __construct(Configuration $configuration, AbstractPlatform $platform)
     {
         $this->configuration = $configuration;
         $this->platform      = $platform;
+        $this->sqlFormatter  = new SqlFormatter(new NullHighlighter());
```
greg0ire, line 41: I hesitated to have a ctor argument for this, but then that would mean do some mocking in the test, and "Don't mock what you don't own" :tm: 
When I see this, I think it could be cool to have a named constructor like `SqlFormatter::withoutHighlight()`

goetas, line 41: creating the sql formatter should be done on `generate` when `$formatted=true`, having it in the constructor makes is a hard dependency

goetas: should we move the `if (! class_exists(SqlFormatter::class)) {` checks inside this class?

greg0ire: The exception talks about a CLI option, so I don't think we should.

Issue #967, opened 2020-04-22
[3.0-alpha] Seamless 2.0 -> 3.0 upgrade, keeping migration versions data


pamil:
Is there a recommended way to upgrade an existing application to use Migrations 3.0 and not lose migration versions data?

Ideally, after updating the package versions and adjusting the configuration, Migrations 3.0 would know which migrations have already been applied by Migrations 2.0.

My ideas:

1) An extension point to transform previous migration versions to the new format. For example, I could define that `20170101235959` is now `Sylius\CoreBundle\Migrations\Version20170101235959` and it would execute before actually running the migrations.

2) More of a meta-solution. Does it make sense to make a migration that migrates that data? Migration that inserts rows into the new migrations versions table based on the old one.

goetas:
That should be already automatically handled by https://github.com/doctrine/migrations/blob/c8ccaeb98bcc118b4657b854802e5202f04f40dc/lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php#L204 

if is not the case, there might be a bug. Did you already try it?

emodric:
I had no problems with keeping migrations data. I did have to run `doctrine:migrations:sync-metadata-storage` to upgrade to the new storage format, though.

On a related note, @goetas would it be possible to run `doctrine:migrations:sync-metadata-storage` command automatically the first time `doctrine:migrations:migrate` command is ran after upgrading to v3?

goetas:
> On a related note, @goetas would it be possible to run doctrine:migrations:sync-metadata-storage command automatically the first time doctrine:migrations:migrate command is ran after upgrading to v3?

That should have been already the case.
Will double check it.

pamil:
I'll try whether I can reproduce it.

pamil:
It was mostly my fault, I deleted `table_name` configuration option instead of setting its equivalent in the new configuration. After configuring migrations to use the same table, migration works, but only after running `sync-metadata-storage`:

<img width="1680" alt="Screenshot 2020-04-22 at 12 17 15" src="https://user-images.githubusercontent.com/1897953/79970622-79fe7680-8493-11ea-9823-77d64bce3bfd.png">

goetas:
Will be fixed in https://github.com/doctrine/migrations/pull/973

PR #973, opened 2020-04-30
Auto update metadata table when migrating

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | -
| Fixed issues | https://github.com/doctrine/migrations/issues/967 (in particular https://github.com/doctrine/migrations/issues/967#issuecomment-617688917)

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
@@ -101,6 +101,18 @@ EOT
 
     protected function execute(InputInterface $input, OutputInterface $output) : int
     {
+        $migratorConfigurationFactory = $this->getDependencyFactory()->getConsoleInputMigratorConfigurationFactory();
+        $migratorConfiguration        = $migratorConfigurationFactory->getMigratorConfiguration($input);
+
+        $question = 'WARNING! You are about to execute a database migration that could result in schema changes and data loss. Are you sure you wish to continue?';
+        if (! $migratorConfiguration->isDryRun() && ! $this->canExecute($question, $input)) {
+            $this->io->error('Migration cancelled!');
+
+            return 1;
+        }
+
+        $this->getDependencyFactory()->getMetadataStorage()->ensureInitialized();
+
         $versions  = $input->getArgument('versions');
         $path      = $input->getOption('write-sql');
         $direction = $input->getOption('down') !== false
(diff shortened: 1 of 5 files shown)


tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
```diff
@@ -340,9 +381,28 @@ protected function setUp() : void
...
         $this->dependencyFactory->setService(MetadataStorage::class, $this->storage);
     }
+
+    private function alterMetadataTable() : void
+    {
+        $originalTable = $this->connection->getSchemaManager()
+            ->listTableDetails($this->metadataConfiguration->getTableName());
+
+        $modifiedTable = $this->connection->getSchemaManager()
+            ->listTableDetails($this->metadataConfiguration->getTableName());
```
stof, line 396: why listing table details twice rather than cloning the Table object ?

goetas: wasn't sure that `clone` will deep clone the required props, but will try to use that.

goetas: seems working, fixed, thanks

lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
```diff
@@ -101,6 +101,18 @@ protected function configure() : void
 
     protected function execute(InputInterface $input, OutputInterface $output) : int
     {
+        $migratorConfigurationFactory = $this->getDependencyFactory()->getConsoleInputMigratorConfigurationFactory();
+        $migratorConfiguration        = $migratorConfigurationFactory->getMigratorConfiguration($input);
+
+        $question = 'WARNING! You are about to execute a database migration that could result in schema changes and data lost. Are you sure you wish to continue?';
```
greg0ire, line 107:
```suggestion
        $question = 'WARNING! You are about to execute a database migration that could result in schema changes and data loss. Are you sure you wish to continue?';
```

tests/Doctrine/Migrations/Tests/Tools/Console/Command/ExecuteCommandTest.php
```diff
@@ -174,7 +178,7 @@ protected function setUp() : void
 
         $this->planCalculator = $this->createMock(MigrationPlanCalculator::class);
         $this->planCalculator
-            ->expects(self::once())
+            ->expects(self::any())
```
greg0ire, line 181:
```suggestion
```

lchrusciel:
Hey Asmir,

apparently it will not solve the problem raised by Kamil (or did something wrong). Current pure Sylius 1.8 installation results with the same error: 
image

Running the sync command fixes the problem.

goetas:
@lchrusciel most of the ppl having a similar issue have solved it by specifying the `serverVersion` parameter for the connection.

PR #975, opened 2020-05-03
Allow to define lazy services in the dependency factory

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature/improvement
| BC Break     | -
| Fixed issues | -

#### Summary
In the current master, overriding services was possible only via `setService()`, but if that dependency needed other services from the dependency factory, was tricky because that operation most probably will freeze the container.

With this PR is possible to define services using `setDefinition()`. This method as second argument takes a callable responsible for resolving the service id. That callable will receive as first argument the dependency factory so it can retrieve other services while resolving the current one.


goetas:
Do you have suggestions on how to fix scrutinizer? The dependency factory has almost no logic, it is simply big...

greg0ire:
Scrutinizer is not a required check, so maybe just ignore it and merge?

PR #976, opened 2020-05-03
Move status info fetching into MigrationStatusInfosHelper

goetas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | -
| Fixed issues | -

#### Summary

Move logic from the command class to the `MigrationStatusInfosHelper` to display the current migration status


greg0ire:
Please improve your commit message, the subject should be shorter than 50 chars, if you need more you can use the commit message _body_ . Explaining why you are doing this in it would be nice too, something like "commands should be about handling CLI-related things" or whatever the actual reason for this is.

goetas:
done

PR #979, opened 2020-05-23
Allow to use --write-sql to save the migration execution log to file

goetas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | -
| Fixed issues | -

#### Summary

In the 2.x version, passing `--write-sql` implied `--dry-run`. 
This PR decouples the two, so is possible to run:
- `--dry-run` without writing the result to file
- `--dry-run` and `--write-sql`
- `--write-sql` to have a "log" of executed queries

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -10,7 +10,9 @@ please refer to the [Code BC breaks](#code-bc-breaks) section.
   Console output is not covered by the BC promise, so please try not to rely on specific a output.
   Different levels of verbosity are available now (`-v`, `-vv` and `-vvv` ).
 - The `--show-versions` option from `migrations:status` command has been removed, 
-  use `migrations:list` instead. 
+  use `migrations:list` instead.
+- The `--write-sql` option for `migrations:migrate` and `migrations:execute` does not imply dry-run anymore,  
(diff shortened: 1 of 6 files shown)


greg0ire:
The fact you didn't touch `composer.*` makes the Travis failure a bit concerning :/

goetas:
I think that is more a travis issue rather than php

greg0ire:
I noticed that the phpunit output was very slow, and also, XDebug seems to be disabled in a non standard way.

greg0ire:
Link to the standard way: https://docs.travis-ci.com/user/languages/php/#disabling-preinstalled-php-extensions

greg0ire:
Anyway I think you are right, because there definitely was output before it displayed "No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself."

goetas:
Found the issue. It was waiting for user input (when running migrations it ask to confirm before proceeding). It seems that recent symfony console detects that the terminal is not interactive. I have updated the test.

greg0ire:
Ah nice! That was tricky!

Issue #820, opened 2019-05-11
migrations:dump-schema loses information


guilliamxavier:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 2.0.2

Refs #689

#### Summary

The migration generated for the dumped schema is not complete.

#### Current behavior

The `migrations:dump-schema` command does not handle existing foreign keys, as opposed to the `migrations:diff` command.

#### How to reproduce

PHP 7.2, MySQL 5.7

1. Run `composer create-project symfony/website-skeleton myapp && cd myapp`
    => relevant packages installed:
    
    <details>
    
    - doctrine/dbal (v2.9.2)
    - doctrine/migrations (v2.0.2)
    - doctrine/orm (v2.6.3)
    - doctrine/doctrine-bundle (1.10.2)
    - doctrine/doctrine-migrations-bundle (v2.0.0)
    
    </details>

2. Copy file .env to .env.local and configure the `DATABASE_URL` environment variable (`mysql://db_user:db_password@127.0.0.1:3306/db_name`)

3. Run `php bin/console doctrine:database:create`

4. Create two entity files, with an association:
    - src/Entity/Foo.php:
        
        <details>
        
        ```php
        <?php
        
        namespace App\Entity;
        
        use Doctrine\ORM\Mapping as ORM;
        
        /** @ORM\Entity */
        class Foo
        {
            /** @ORM\Column(type="integer") @ORM\Id */
            private $id;
        
            /** @ORM\Column(type="string", length=50) */
            private $name;
        }
        ```
        
        </details>
        
    - src/Entity/Bar.php:
        
        <details>
        
        ```php
        <?php
        
        namespace App\Entity;
        
        use Doctrine\ORM\Mapping as ORM;
        
        /** @ORM\Entity */
        class Bar
        {
            /** @ORM\Column(type="integer") @ORM\Id */
            private $id;
        
            /** @ORM\Column(type="string", length=100) */
            private $label;
        
            /** @ORM\ManyToOne(targetEntity=Foo::class) */
            private $foo;
        }
        ```
        
        </details>

5. Run `php bin/console cache:clear && php bin/console doctrine:migrations:diff`
    => generated file (src/Migrations/Version20190511090651.php):
    
    <details>
    
    ```php
    <?php

    declare(strict_types=1);

    namespace DoctrineMigrations;

    use Doctrine\DBAL\Schema\Schema;
    use Doctrine\Migrations\AbstractMigration;

    /**
[...]

PR #981, opened 2020-06-07
Add "from-empty-schema" option for "diff" command

guilliamxavier:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #820

#### Summary

Allow to execute `migrations:diff --from-empty-schema`
(as alternative to `migrations:dump-schema` but using the mapping instead of the database)


guilliamxavier:
Not tested yet (hence draft) but WDYT?

guilliamxavier:
Update: I have tested the patch locally and I confirm it works as expected 😃 (but writing an automated test would be another story...)

tests/Doctrine/Migrations/Tests/Tools/Console/Command/DiffCommandTest.php
```diff
@@ -55,9 +55,14 @@ public function testExecute() : void
         $input->expects(self::at(4))
             ->method('getOption')
             ->with('check-database-platform')
-            ->willReturn(true);
+            ->willReturn(false);
```
guilliamxavier, line 58: This change is just to have a value different from the default (like all other tested options) and for the call below

goetas:
I like this change! 

A test should test the `createEmptySchema` method asserting something on the expected output.
(the current changes to the tests are just making sure that args are read as expected)

guilliamxavier:
I have added tests (but I'm not sure because I couldn't mock the `new Schema(...)` in `DiffGenerator::createEmptySchema()`...)

tests/Doctrine/Migrations/Tests/Generator/DiffGeneratorTest.php
```diff
@@ -118,6 +119,60 @@ static function ($name) {
...
+            ->willReturn(null);
+
+        $toSchema->method('getName')->willReturn('schema');
+        $toSchema->method('getTables')->willReturn(['schema.table_name' => $table]);
+        $toSchema->method('getTable')->with('table_name')->willReturn($table);
+        $toSchema->method('getNamespaces')->willReturn([]);
+        $toSchema->method('getSequences')->willReturn([]);
+
+        $table->method('getName')->willReturn('schema.table_name');
+        $table->method('getShortestName')->with('schema')->willReturn('table_name');
```
guilliamxavier, line 152: After thinking, that part of the test is too fragile (tightly coupled to internal DBAL implementation). I'll rework the implementation (during the weekend)

guilliamxavier:
I don't know why the "**continuous-integration/travis-ci**" check is blocked at "_Expected — Waiting for status to be reported_"... you can see that it passed: https://travis-ci.org/github/doctrine/migrations/builds/700372177

Anyway, this is now ready for review @goetas :slightly_smiling_face:

greg0ire:

How to do that?

1. `git rebase -i origin/2.3.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

greg0ire (commented):
Can you please add some documentation?

guilliamxavier:
@greg0ire: done and done ;)

greg0ire (approved):

goetas:
thanks @guilliamxavier 

I think that I will have fun times when merging back this to `master` :)

alzee:
Thanks @guilliamxavier
Great job!

Issue #987, opened 2020-06-14
Get Back Meaningful Message When "latest" Version Already Reached


bugalot:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 3.0.0

#### Summary

When launching a migration on a fully migrated platform (i.e. on "latest" version), the message changed from 3.0.0-alpha1 to 3.0.0-beta1/stable. 

Before we had a nice and clean `"[WARNING] Already at latest version"` which was perfect.

Now we have `"[ERROR] The version "latest" couldn't be reached, you are at version xxx"` which surprised a bit our developer community.

We provide a [Pull Request ](https://github.com/doctrine/migrations/pull/988) in order to get back the more meaningful message. It would be great if you guys could have a look at it and maybe integrate it :smile:.

It might look like a detail but we feel that providing nice and meaningful is essential. It seems that the message change happened in this [commit](https://github.com/doctrine/migrations/commit/3cb3932d85caed3d0016208abe1f2440f9f650d5). It is just a "loss" we would like to see back again.

#### Current behavior

When a platform is at "latest" version, launching a new migration produces the following output:

```
[ERROR] The version "latest" couldn't be reached, you are at version xxx
```

#### How to reproduce

Launch a successful migration process, then run it again.

#### Expected behavior

When a platform is at "latest" version, launching a new migration produce the following output:

```
[WARNING] Already at latest version
```

goetas:
It is not a BC break, console messages are not covered by BC promise

bugalot:
Agreed, it was more for convenience! Up to you to decide !

Gemorroj:
@goetas but why it's now ERROR (not WARNING)? it breaks our ci.

goetas:
> UPD: use --allow-no-migration

yes

akondas:
> It is not a BC break, console messages are not covered by BC promise

Console messages not, but this is not only message text change but also exist code. In my opinion, however, this is a change worth at least adding to the changelog.

fmarchalemisys:
The message is not only misleading, it is wrong too.

The latest version was reached without error! It Just did not happen during this run. 

There is no need to alarm the user about it.

goetas:
https://github.com/doctrine/migrations/pull/988 has been merged

goetas:
- #988 has been merged (that improves the wording  in the message)
- #/993 has been merged (that explain the `--allow-no-migration` command)

So I'm closing this.

[...]

PR #988, opened 2020-06-14
Re-enable alpha1 meaningful message when "latest" already reached.

bugalot:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #987 

#### Summary

Tries to fix the issue described in issue #987. To summarize, it is a proposal to get back the nice and clean `"[WARNING] Already at latest version"` when trying to perform a migration on a platform already at "latest" version.

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
@@ -218,11 +218,16 @@ EOT
         if (in_array($versionAlias, ['first', 'next', 'latest'], true) || strpos($versionAlias, 'current') === 0) {
             $version = $this->getDependencyFactory()->getVersionAliasResolver()->resolveVersionAlias('current');
 
-            $message = sprintf(
-                'The version "%s" couldn\'t be reached, you are at version "%s"',
-                $versionAlias,
-                (string) $version
-            );
+            // Allow meaningful message when latest version already reached.
+            if ($versionAlias === 'next' || $versionAlias === 'latest') {
+                $message = 'Already at latest version';
+            } else {
+                $message = sprintf(
+                    'The version "%s" couldn\'t be reached, you are at version "%s"',
+                    $versionAlias,
+                    (string) $version
+                );
+            }
 
             if ($allowNoMigration) {
                 $this->io->warning($message);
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
```diff
@@ -218,11 +218,16 @@ private function errorForAlias(string $versionAlias, bool $allowNoMigration) : i
...
             $version = $this->getDependencyFactory()->getVersionAliasResolver()->resolveVersionAlias('current');
 
-            $message = sprintf(
-                'The version "%s" couldn\'t be reached, you are at version "%s"',
-                $versionAlias,
-                (string) $version
-            );
+            // Allow meaningful message when latest version already reached.
+            if ($versionAlias === 'next' || $versionAlias === 'latest') {
+                $message = 'Already at latest version';
```
goetas, line 223: What about adding info on the current version ?

Like `Already at latest version (%s)`

bugalot:
I implemented the suggested change. The message will be now like "Already at version "VERSION_ALIAS" ("VERSION")".

Thanks a lot for the review @goetas !

jmsche:
Hi,

I'm just wondering if it makes sense to throw a warning/error if all migrations have been executed?

Also throwing a warning while passing the `--allow-no-migration` option seems weird to me as we explicitly allow it (eg. during deploy).

And thank you @bugalot for the message improvement :)

elkuku:
Not sure why this isn't considered a BC break since it breaks CI without passing the (new) `--allow-no-migration` option :(

Please merge soon ;)

greg0ire:
@elkuku what makes you say it is not considered a BC-break? This is version 3.0.0…

greg0ire:
> I'm just wondering if it makes sense to throw a warning/error if all migrations have been executed?

If all migrations have been executed, running the migrations command is a bit weird, and notifying the user about it makes sense IMO.

> Also throwing a warning while passing the --allow-no-migration option seems weird to me as we explicitly allow it (eg. during deploy).

Good point, maybe it fits more the description for notice: https://github.com/php-fig/log/blob/master/Psr/Log/LoggerInterface.php#L82

greg0ire:

How to do that?

1. `git rebase -i origin/master`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

goetas:
Thanks!

PR #994, opened 2020-06-16
Site meta and branch alias

goetas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

update branch info

### DIFF
--- a/.doctrine-project.json
+++ b/.doctrine-project.json
@@ -5,10 +5,17 @@
     "docsSlug": "doctrine-migrations",
     "versions": [
         {
-            "name": "3.0",
+            "name": "3.1",
             "branchName": "master",
-            "slug": "latest",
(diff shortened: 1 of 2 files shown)


goetas:
You are right, there is still the topic of forward compatibility layer between 2.x and 3.x.. that honestly to me seems very very hard and not worth.

What do you think?

greg0ire:
Well you certainly know far more than I do on that topic. If you can think of obvious things that could help with the upgrade path, consider doing them. Maybe have a look at issues and see what people struggle with? For instance they could be encouraged to migrate their custom templates to a format that uses the new variable: https://github.com/doctrine/migrations/issues/1000

goetas:
Will wait some time and see what it make sense to include in #1000

PR #997, opened 2020-06-16
Fix --write-sql command option description

bobvandevijver:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

I believe this fixes a small oversight from f3c7942c91272a8ce3b774182f8c882b42387294, which didn't update the command help descriptions with the new `--write-sql` behaviour.

Note: I'm not sure on the use of `prepared`, as I do not know at what stage the sql file is actually written.


goetas:
thanks!

PR #999, opened 2020-06-17
Allow using on PHP 7.1 with Composer 2

nicolas-grekas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1066

Similar to https://github.com/doctrine/dbal/pull/4084

In order to support Composer 2, `ocramius/package-versions` requires PHP 7.4.
This effectively means that using it in "require" forces ppl on PHP 7.1 to use Composer 1.

@seldaek created the `composer/package-versions-deprecated` to unlock this situation.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,8 +12,8 @@
     ],
     "require": {
         "php": "^7.1",
+        "composer/package-versions-deprecated": "^1.8",
         "doctrine/dbal": "^2.9",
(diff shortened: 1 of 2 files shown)


greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @nicolas-grekas !

PR #1006, opened 2020-06-19
Document table metadata configs are optional

goetas:
|      Q       |   A
|------------- | -----------
| Type         | documentation
| BC Break     | -
| Fixed issues | -

#### Summary
Add note that table metadata configurations are optional

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -52,7 +52,7 @@ return [
     'table_storage' => [
         'table_name' => 'doctrine_migration_versions',
         'version_column_name' => 'version',
-        'version_column_length' => 1024,
+        'version_column_length' => 191,
         'executed_at_column_name' => 'executed_at',
         'execution_time_column_name' => 'execution_time',
     ],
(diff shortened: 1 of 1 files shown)


UPGRADE.md
```diff
@@ -3,6 +3,7 @@
 - The "version" is the FQCN of the migration class (existing entries in the migrations table will be automatically updated).
 - `MigrationsEventArgs` and `MigrationsVersionEventArgs` expose different API, 
 please refer to the [Code BC breaks](#code-bc-breaks) section.
+- Using select statements in `$this->addSql()` with the `PDO::MSYQL_ATTR_USE_BUFFERED_QUERY` will not work anymore.
```
bcremer, line 6: This will not work regardless of the `PDO::MSYQL_ATTR_USE_BUFFERED_QUERY` setting. The generated MySQL Exception message is missleading here.

UPGRADE.md
```diff
@@ -69,6 +70,25 @@ return [
 
 Files in XML, YAML or JSON also changed in a similar way. Please refer to the official documentation for more details.
 
+Note: the option in `table_storage` needs to be updated only if you have changed the metadata table settings
+by using v2 options as `table_name`, `column_name`, `column_length` or `executed_at_column_name`. If you did not change
```
greg0ire, line 74:
```suggestion
by using v2 options such as `table_name`, `column_name`, `column_length` or `executed_at_column_name`. If you did not change
```

UPGRADE.md
```diff
@@ -69,6 +70,25 @@ return [
...
+those settings is recommended to not provide the options and let doctrine figure out the best settings.
```
greg0ire, line 75:
```suggestion
those settings, it is recommended to not provide the options and let doctrine figure out the best settings.
```

UPGRADE.md
goetas: @bcremer will move this to a separate pull request where we can discuss it better

PR #1008, opened 2020-06-19
Fixed out-of-date lock file

bobvandevijver:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

The lock was not updated in b9469d354d784d02ac288a07e30490efa3ee1537. Gave me an error when building some other improvements.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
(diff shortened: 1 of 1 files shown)


greg0ire:
Is `master` the lowest branch where this change is relevant?

bobvandevijver:
> Is `master` the lowest branch where this change is relevant?

Looking at the commit that caused this, I do believe so. I also checked 3.0.x, which doesn't have an outdated lock file.

Issue #996, opened 2020-06-16
Missing SQL output during migration


bobvandevijver:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | not really, it was available in 2.x
| RFC         | no
| BC Break    | no

#### Summary

I'm missing the SQL output when running the migrate command, and I do not see any option to enable it (except for writing to file). This is something that was available earlier, and I would like to see it return if possible.

goetas:
as explained in the upgrade notes (https://github.com/doctrine/migrations/blob/master/UPGRADE.md#console), you can use the `-vvv` option

bobvandevijver:
Okay, I missed that, thanks. 

Would it be an idea to add this to the help of the command itself? Because that was actually the place I was looking for the option, and the upgrade notes will probably not be read by people using this in new projects. In my opinion the default description for `-v|vv|vvv, --verbose` does not reflect that it is actually used for this functionality.

goetas:
that can be a nice thing to do. 
Do you want to send a pull request for it?

PR #1009, opened 2020-06-19
#996: Add verbosity notice to command help

bobvandevijver:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #996

#### Summary

As discussed in #996, add some help text for the verbosity levels in the `execute` and `migrate` commands.

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
@@ -74,6 +74,11 @@ The <info>%command.name%</info> command executes migration versions up or down m
 
     <info>%command.full_name% FQCN</info>
 
+You can show more information about the process by increasing the verbosity level. To see the
+executed queries, set the level to debug with <comment>-vv</comment>:
+
+    <info>%command.full_name% FQCN -vv</info>
+
 If no <comment>--up</comment> or <comment>--down</comment> option is specified it defaults to up:
 
     <info>%command.full_name% FQCN --down</info>
@@ -92,6 +97,7 @@ Or you can also execute the migration without a warning message which you need t
 
 All the previous commands accept multiple migration versions, allowing you run execute more than
 one migration at once:
+
     <info>%command.full_name% FQCN-1 FQCN-2 ...FQCN-n </info>
 
 EOT
--- a/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
@@ -80,6 +80,11 @@ The <info>%command.name%</info> command executes a migration to a specified vers
 
     <info>%command.full_name%</info>
 
+You can show more information about the process by increasing the verbosity level. To see the
+executed queries, set the level to debug with <comment>-vv</comment>:
+
+    <info>%command.full_name% -vv</info>
+
 You can optionally manually specify the version you wish to migrate to:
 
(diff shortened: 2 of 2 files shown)


greg0ire:
Aaaand Travis is not responding, again… try closing and reopening from time to time I guess…

lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
```diff
@@ -74,6 +74,11 @@ protected function configure() : void
 
     <info>%command.full_name% FQCN</info>
 
+You can show more information about the process by increasing the verbosity level. To see the
+executed queries, set the level to debug with <comment>-vv</comment>:
```
greg0ire, line 78: [The docs say debug level is obtained with `-vvv`](https://symfony.com/doc/current/console/verbosity.html)

bobvandevijver: I decided to use the terminology used in the printed message, instead of the one used by Symfony. 

```
bobvandevijver@webdev:/var/www/dev$ bin/console doctrine:migrations:migrate -n -vv
[notice] Migrating up to DoctrineMigrations\Version20200618130453
[info] ++ migrating DoctrineMigrations\Version20200618130453
[debug] ALTER TABLE <snip>
[info] Migration DoctrineMigrations\Version20200618130453 migrated (took 180.7ms, used 22M memory)
[notice] finished in 199.9ms, used 22M memory, 1 migrations executed, 1 sql queries
```

bobvandevijver: The actual call in the code is also `$this->logger->debug()`, so there might be some place where it is configured wrongly, or there is just simply a mismatch between the used terminology.

greg0ire: > some place

That kind of configuration would be in your application though, there does not seem to be any code related to monolog in the bundle. You can use `bin/console debug:config monolog` to see what is up. [These docs](https://symfony.com/doc/current/logging/monolog_console.html) seem to talk about a mapping between log levels and verbosity, maybe that would be the root cause?

bobvandevijver: I totally agreed with you concerning the application configuration. But as this is a clean project with only the default flex installed configuration in place for the loggin, I was sceptical. So I went digging.

It turns out not Monolog is used for the console output, but the [`ConsoleLogger`](https://github.com/doctrine/migrations/blob/f327c40fe06e30d68f4b3745ac88d0eded5e0777/lib/Doctrine/Migrations/Tools/Console/ConsoleLogger.php) from this library. And that clearly maps the debug loglevel to the very verbose output:

https://github.com/doctrine/migrations/blob/f327c40fe06e30d68f4b3745ac88d0eded5e0777/lib/Doctrine/Migrations/Tools/Console/ConsoleLogger.php#L47

greg0ire: Excellent sleuth job! :detective:  So that's why the style is a bit different!

goetas:
@bobvandevijver i think that you will have to rebase and push again your changes just to trigger again the travis ci build. (something must gone wrong with webhook delivery...)

greg0ire:
Closing and reopening is usually enough in that kind of case

goetas:
Ah, did not know that

Issue #998, opened 2020-06-16
--query-time option requires -vv as well


bobvandevijver:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.0.0

#### Summary

The `--query-time` command option requires `-vv` to be also supplied in order to produce output.

#### Current behavior

See summary.

#### How to reproduce

```
bobvandevijver@webdev:/var/www/dev$ bin/console doctrine:migrations:migrate --no-interaction --query-time
[notice] Migrating up to DoctrineMigrations\Version20200616155007
[notice] finished in 60.1ms, used 22M memory, 1 migrations executed, 1 sql queries

bobvandevijver@webdev:/var/www/dev$ bin/console doctrine:migrations:migrate --no-interaction prev
[notice] Migrating down to 0
[notice] finished in 51.4ms, used 22M memory, 1 migrations executed, 1 sql queries

bobvandevijver@webdev:/var/www/dev$ bin/console doctrine:migrations:migrate --no-interaction --query-time -vv
[notice] Migrating up to DoctrineMigrations\Version20200616155007
[info] ++ migrating DoctrineMigrations\Version20200616155007 (Add entity)
[debug] CREATE TABLE <snip>
[debug] 26.2ms
[info] Migration DoctrineMigrations\Version20200616155007 migrated (took 55.2ms, used 22M memory)
[notice] finished in 66.2ms, used 22M memory, 1 migrations executed, 1 sql queries
```
#### Expected behavior

The `--query-time` option should output the query time even when `-vv` is not supplied. 

I would propose to add some descriptive text with the query time, instead of only showing the milliseconds (cause, what is the time for?). For example: `[debug] Previous query took 26.2ms`.

goetas:
Solved in https://github.com/doctrine/migrations/pull/1010

PR #1010, opened 2020-06-19
#998: Show query and time information when using --query-time

bobvandevijver:
|      Q       |   A
|------------- | -----------
| Type         | bug/improvement
| BC Break     | no
| Fixed issues |  #998

#### Summary

Always show the queries and their timing when supplying the `--query-time` option

### DIFF
--- a/lib/Doctrine/Migrations/Version/DbalExecutor.php
+++ b/lib/Doctrine/Migrations/Version/DbalExecutor.php
@@ -19,6 +19,7 @@ use Doctrine\Migrations\Provider\SchemaDiffProvider;
 use Doctrine\Migrations\Query\Query;
 use Doctrine\Migrations\Tools\BytesFormatter;
 use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
 use Symfony\Component\Stopwatch\Stopwatch;
 use Throwable;
 use function count;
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Migrations/Version/DbalExecutor.php
```diff
@@ -301,23 +302,24 @@ private function executeResult(MigratorConfiguration $configuration) : void
...
 
-        $this->logger->debug('{query} {params}', [
-            'query' => $query->getStatement(),
-            'params' => $params,
-        ]);
+        $this->logger->log($configuration->getTimeAllQueries() ? LogLevel::NOTICE : LogLevel::DEBUG,
+            '{query} {params}', [
+                'query'  => $query->getStatement(),
+                'params' => $params,
+            ]);
```
greg0ire, line 322: This is a PSR-2 violation, the call has to be fully multiline.

bobvandevijver: Guess I fixed it now 😃

bobvandevijver:
> Sorry for the late review... why not just trigger the `-vvv` option when `--query-time` is passed rather than changing logging levels at runtime?

If I dot that, passing the `--query-time` would imply debug verbosity, which is not something I would expect. I'm only expecting the lines relevant to my supplied options to be printed, which is what it now does. Otherwise it would be possible that it prints a full debug log (which isn't that extensive at the moment, but still), resulting in the desired output to be "flooded" with information I didn't ask for.

This way, it just makes sure that, without depending on the user set verbosity level (except for quiet obviously), it always print the information requested.

goetas:
@bobvandevijver  That makes sense.
Thank you!

PR #1015, opened 2020-06-20
Migrate returns 0 as exit codes for empty plans as in 2.x

goetas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | -
| Fixed issues | 

#### Summary

In doctrine 2.x the migration command throws an exception if there
are not migrations to execute (unless the --allow-no-migrations
parameter is passed). In 3.0.0 this was misunderstood and the exception
was thrown even if there are registered migrations but there are no migrations to
execute for the current migration plan.
With this commit we are back to the 2.x strategy.

This commit also improves the error messages when the various aliases
can not be reached or are already reached.

### DIFF
--- a/tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
+++ b/tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
@@ -75,16 +75,61 @@ class MigrateCommandTest extends MigrationTestCase
 
     public function testTargetUnknownVersion() : void
     {
-        $result = new ExecutionResult(new Version('A'));
-        $this->storage->complete($result);
-
         $this->migrateCommandTester->execute(
-            ['version' => 'A'],
+            ['version' => 'B'],
             ['interactive' => false]
         );
 
-        self::assertStringContainsString('[ERROR] Unknown version: A', $this->migrateCommandTester->getDisplay(true));
+        self::assertStringContainsString('[ERROR] Unknown version: B', $this->migrateCommandTester->getDisplay(true));
         self::assertSame(1, $this->migrateCommandTester->getStatusCode());
+    }
+
+    /**
+     * @return bool[][]
+     */
+    public function getMigrateWithMigrationsOrWithout() : array
+    {
+        return [
+            // migrations available, allow-no-migrations
+            [false, false],
+            [true, false],
+            [false, true],
+            [true, true],
+        ];
+    }
+
+    /**
+     * @dataProvider getMigrateWithMigrationsOrWithout
+     */
+    public function testMigrateWhenNoMigrationsAvailable(bool $hasMigrations, bool $allowNoMigration) : void
+    {
+        $finder                    = $this->createMock(Finder::class);
+        $factory                   = $this->createMock(MigrationFactory::class);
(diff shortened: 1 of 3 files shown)


tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
```diff
@@ -74,63 +75,121 @@ class MigrateCommandTest extends MigrationTestCase
...
+     * @return bool[][]
+     */
+    public function getMigrateWithMigrationsOrWithout() : array
+    {
+        return [
+            // migrations available, allow-no-migrations
+            [false, false],
+            [true, false],
+            [false, true],
+            [true, true],
```
greg0ire, line 97: Tests should not be complex IMO, the expected return code should be here
```suggestion
            [false, false, 1],
            [true, false, 0],
            [false, true, 0],
            [true, true, 0],
```

tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
```diff
@@ -74,63 +75,121 @@ class MigrateCommandTest extends MigrationTestCase
...
+        ];
+    }
+
+    /**
+     * @dataProvider getMigrateWithMigrationsOrWithout
+     */
+    public function testMigrateWhenNoMigrationsAvailable(bool $hasMigrations, bool $allowNoMigration) : void
```
greg0ire, line 104:
```suggestion
    public function testMigrateWhenNoMigrationsAvailable(bool $hasMigrations, bool $allowNoMigration, int $expectedExitCode) : void
```

tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
```diff
@@ -74,63 +75,121 @@ class MigrateCommandTest extends MigrationTestCase
...
+            self::assertStringContainsString(
+                sprintf(
+                    '[%s] The version "latest" couldn\'t be reached, there are no registered migrations.',
+                    $allowNoMigration ? 'WARNING' : 'ERROR'
+                ),
+                $display
+            );
+        }
+
+        self::assertSame($hasMigrations || $allowNoMigration ? 0 : 1, $this->migrateCommandTester->getStatusCode());
```
greg0ire, line 132:
```suggestion
        self::assertSame($expectedExitCode, $this->migrateCommandTester->getStatusCode());
```

tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
```diff
@@ -74,63 +75,121 @@ class MigrateCommandTest extends MigrationTestCase
...
         self::assertSame(1, $this->migrateCommandTester->getStatusCode());
     }
 
+    /**
+     * @return bool[][]
+     */
+    public function getMigrateWithMigrationsOrWithout() : array
+    {
+        return [
+            // migrations available, allow-no-migrations
```
greg0ire, line 93:
```suggestion
            // migrations available, allow-no-migrations, expected exit code
```

Issue #1021, opened 2020-06-23
--db option  not in Doctrine migrations 3 ?


darkcattz:
Hello, 

New version of Doctrine migrations no longer has the --db for the command `doctrine:migrations:migrate` option

`[error] Error thrown while running command "crm:createNewDbClient 1". Message: "The "db" argument does not exist."`

How to do now with version 3 of Doctrine Migrations ?

Thanks.

goetas:
to make this feature work with 3.0 we should create a `ConnectionManagerAwareConnectionLoader` (that implements `ConnectionLoader`). and pass that as connection loader to the dependency factory. (Similar to what has been done the the `--config` param in https://github.com/doctrine/migrations/blob/df6138ffd06af43639f73b5b6447bcbe45d289d8/lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php#L74)

darkcattz:
Thanks. 

the --db option is indeed essential for applications with several databases (in my case more than a hundred, one per customer -> Saas model).
This option allows you to launch Symfony commands with connection on the fly

Regards

goetas:
does your symfony app has defined the connections for all of those customers?

goetas:
@darkcattz Can you please explain a bit more in detail how are you using this feature? We had a team discussion on how to re-implement this in 3.0 and we wanted to see if there are ways to improve it (as example allowing to run migrations across all defined connections...?)

darkcattz:
No, only 2 connections :
- One for the factory (customer list : each line describes the client + the (random) name of his database
- Another default connection which serves as the basis for the connection but which I must complete with the correct client base name to connect to its database.

E.g. : Symfony crm:createNewCustomerDatabase command :
1°) I check that the user has been created in the factory and I get the name of its random database name (which does not yet exist)
2°) I create a new Mysql user for this customer with only the rights on his database (Doctrine Dbal).
3°) I create the client's database (empty for now)
4°) I launch the Doctrine Migrations command with the --db option which will directly create all the tables in the new client database and hydrate important tables (creation of default users for example).
5°) The client base is now operational and the client can connect to the application (checking the client's name in the factory and changing the Doctrine connection on the fly at each login with the correct credentials).

E.g. for doctrine migrations : 
`
                    $reflectionConn = new \ReflectionObject($this->doctrine->getConnection('factory'));
                    $reflectionParams = $reflectionConn->getProperty('params');
                    $reflectionParams->setAccessible(true);

[...]

PR #1023, opened 2020-06-25
Allow alternative connection/entity-manager via --em or --conn option

goetas:
|      Q       |   A
|------------- | -----------
| BC Break     | no
| Fixed issues | https://github.com/doctrine/migrations/issues/1021 https://github.com/doctrine/DoctrineMigrationsBundle/issues/341

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -41,6 +41,8 @@ Now, in the root of your project place a file named ``migrations.php``, ``migrat
 
             'all_or_nothing' => true,
             'check_database_platform' => true,
+            'connection' => null,
+            'em' => null,
         ];
 
     .. code-block:: yaml
@@ -61,6 +63,9 @@ Now, in the root of your project place a file named ``migrations.php``, ``migrat
         all_or_nothing: true
         check_database_platform: true
 
+        connection: null
+        em: null
+
     .. code-block:: xml
 
         <?xml version="1.0" encoding="UTF-8"?>
@@ -69,6 +74,10 @@ Now, in the root of your project place a file named ``migrations.php``, ``migrat
               xsi:schemaLocation="http://doctrine-project.org/schemas/migrations/configuration/3.0
                             http://doctrine-project.org/schemas/migrations/configuration-3.0.xsd">
 
+            <name>My Project Migrations</name>
+
+            <connection>default</connection>
+            <em>default</em>
             <name>My Project Migrations</name>
 
             <storage>
@@ -110,7 +119,10 @@ Now, in the root of your project place a file named ``migrations.php``, ``migrat
             },
 
             "all_or_nothing": true,
-            "check_database_platform": true
+            "check_database_platform": true,
+
+            "connection": null,
+            "em": null
         }
 
 Please note that if you want to use the YAML configuration option, you will need to install the ``symfony/yaml`` package with composer:
@@ -135,6 +147,10 @@ Here are details about what each configuration option does:
 | migrations                 | no         | []                           | Manually specify the array of migration versions instead of finding migrations.  |
 +----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
 | check_database_platform    | no         | true                         | Whether to add a database platform check at the beginning of the generated code. |
++----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
+| connection                 | no         | null                         | The named connection to use (available only when ConnectionRegistryConnection is used). |
++----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
+| em                         | no         | null                         | The named entity manager to use (available only when ManagerRegistryEntityManager is used). |
 +----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
 
 
--- a/docs/en/reference/custom-configuration.rst
+++ b/docs/en/reference/custom-configuration.rst
@@ -68,3 +68,62 @@ Once you have your custom integration setup, you can modify it to look like the 
     $cli->run();
 
 :ref:`Next Chapter: Migrations Events <events>`
+
+
+It is possible to use multiple entity managers or connections, this is a way to configure your application:
+
+
+.. code-block:: php
+
+    #!/usr/bin/env php
+    <?php
+
+    require_once __DIR__.'/vendor/autoload.php';
+
+    use Doctrine\DBAL\DriverManager;
+    use Doctrine\Migrations\Configuration\Configuration;
+    use Doctrine\Migrations\Configuration\Configuration\ExistingConfiguration;
+    use Doctrine\Migrations\Configuration\Connection\ConnectionRegistryConnection;
+    use Doctrine\Migrations\DependencyFactory;
+    use Doctrine\Migrations\Tools\Console\Command;
+    use Symfony\Component\Console\Application;
+
+    $connection1 = DriverManager::getConnection([...]);
+    $connection2 = DriverManager::getConnection([...]);
+
+    $connectionRegistry = new class(
+        'some_registry',
+        ['foo' => $connection1, 'bar' => $connection2],
+        [], // entity managers
+        'foo', // default connection
+        null, // default entity manager
+        'Doctrine\Persistence\Proxy' // proxy class
+    ) extends AbstractManagerRegistry {
+        // implement abstract methods here
+    };
+
+    $configuration = new Configuration($connection);
+    $configuration->addMigrationsDirectory('MyProject\Migrations', 'some path');
+    $configurationLoader = new ExistingConfiguration($configuration);
+
+    $connectionLoader = ConnectionRegistryConnection::withSimpleDefault($connectionRegistry);
+
+    $dependencyFactory = DependencyFactory::fromConnection(
+        $configurationLoader,
+        $connectionLoader
+    );
+
+    $cli = new Application('Doctrine Migrations');
+    $cli->setCatchExceptions(true);
+
+    $cli->addCommands(array(
+        new Command\MigrateCommand($dependencyFactory),
+        // more commands here
+    ));
+
+    $cli->run();
+
+With this configuration you can use the ``--conn`` parameter to specify a connection that will be used for running
+migrations. If the parameter is not passed, it will fallback to the one passed in the configuration,
+and if that is also not provided it will fallback to the default connection name specified when creating
+the connection registry.
--- a/lib/Doctrine/Migrations/Configuration/Connection/Exception/InvalidConfiguration.php
+++ b/lib/Doctrine/Migrations/Configuration/Connection/Exception/InvalidConfiguration.php
@@ -4,12 +4,20 @@ declare(strict_types=1);
 
 namespace Doctrine\Migrations\Configuration\Connection\Exception;
 
+use Doctrine\DBAL\Connection;
 use InvalidArgumentException;
+use function get_class;
+use function sprintf;
 
 final class InvalidConfiguration extends InvalidArgumentException implements LoaderException
 {
     public static function invalidArrayConfiguration() : self
     {
         return new self('The connection file has to return an array with database configuration parameters.');
+    }
+
+    public static function invalidConnectionType(object $connection) : self
+    {
+        return new self(sprintf('The returned connection must be a %s instance, %s returned.', Connection::class, get_class($connection)));
     }
 }
--- a/lib/Doctrine/Migrations/Configuration/EntityManager/Exception/InvalidConfiguration.php
+++ b/lib/Doctrine/Migrations/Configuration/EntityManager/Exception/InvalidConfiguration.php
@@ -4,12 +4,20 @@ declare(strict_types=1);
 
 namespace Doctrine\Migrations\Configuration\EntityManager\Exception;
 
+use Doctrine\ORM\EntityManagerInterface;
 use InvalidArgumentException;
+use function get_class;
+use function sprintf;
 
 final class InvalidConfiguration extends InvalidArgumentException implements LoaderException
 {
     public static function invalidArrayConfiguration() : self
     {
         return new self('The EntityManager file has to return an array with database configuration parameters.');
+    }
+
+    public static function invalidManagerType(object $em) : self
+    {
+        return new self(sprintf('The returned manager must implement %s, %s returned.', EntityManagerInterface::class, get_class($em)));
     }
 }
--- a/lib/Doctrine/Migrations/Configuration/Exception/InvalidLoader.php
+++ b/lib/Doctrine/Migrations/Configuration/Exception/InvalidLoader.php
@@ -1,0 +1,24 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Migrations\Configuration\Exception;
+
+use Doctrine\Migrations\Configuration\Connection\ConnectionLoader;
+use Doctrine\Migrations\Configuration\EntityManager\EntityManagerLoader;
+use InvalidArgumentException;
+use function get_class;
+use function sprintf;
+
+final class InvalidLoader extends InvalidArgumentException implements ConfigurationException
+{
+    public static function noMultipleConnections(ConnectionLoader $loader) : self
+    {
+        return new self(sprintf('Only one connection is supported by %s', get_class($loader)));
+    }
+
+    public static function noMultipleEntityManagers(EntityManagerLoader $loader) : self
+    {
+        return new self(sprintf('Only one entity manager is supported by %s', get_class($loader)));
+    }
+}
--- a/lib/Doctrine/Migrations/Configuration/Migration/ConfigurationArray.php
+++ b/lib/Doctrine/Migrations/Configuration/Migration/ConfigurationArray.php
@@ -41,6 +41,10 @@ final class ConfigurationArray implements ConfigurationLoader
                     $configuration->addMigrationClass($className);
                 }
             },
+
+            'connection' => 'setConnectionName',
+            'em' => 'setEntityManagerName',
+
             'table_storage' => [
                 'table_name' => 'setTableName',
                 'version_column_name' => 'setVersionColumnName',
--- a/lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
@@ -9,6 +9,7 @@ use Doctrine\Migrations\Configuration\Migration\ConfigurationFileWithFallback;
 use Doctrine\Migrations\DependencyFactory;
 use Doctrine\Migrations\Tools\Console\ConsoleLogger;
 use Doctrine\Migrations\Tools\Console\Exception\DependenciesNotSatisfied;
+use Doctrine\Migrations\Tools\Console\Exception\InvalidOptionUsage;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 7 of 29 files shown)


lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
```diff
@@ -75,6 +92,26 @@ protected function initialize(InputInterface $input, OutputInterface $output) :
...
 
+        if ($input->getOption('em')!== null && $input->getOption('conn')!==null) {
+            throw new InvalidOptionUsage('You can specify only one of the --em and --conn options.');
+        }
+
+        $connectionLoader = $this->dependencyFactory->getConnectionLoader();
+        if ($connectionLoader instanceof ConnectionRegistryConnection && $input->getOption('conn')!==null) {
+            $connectionName = $input->getOption('conn');
+            $connectionLoader->setConnectionName((string) $connectionName);
+        } elseif ($input->getOption('conn')!==null) {
```
SenseException, line 103:
```suggestion
        } elseif ($input->getOption('conn') !== null) {
```

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
```diff
@@ -75,6 +92,26 @@ protected function initialize(InputInterface $input, OutputInterface $output) :
             $this->dependencyFactory->setConfigurationLoader($configurationLoader);
         }
 
+        if ($input->getOption('em')!== null && $input->getOption('conn')!==null) {
+            throw new InvalidOptionUsage('You can specify only one of the --em and --conn options.');
+        }
+
+        $connectionLoader = $this->dependencyFactory->getConnectionLoader();
+        if ($connectionLoader instanceof ConnectionRegistryConnection && $input->getOption('conn')!==null) {
```
SenseException, line 100:
```suggestion
        if ($connectionLoader instanceof ConnectionRegistryConnection && $input->getOption('conn') !== null) {
```

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
```diff
@@ -75,6 +92,26 @@ protected function initialize(InputInterface $input, OutputInterface $output) :
...
+            $connectionName = $input->getOption('conn');
+            $connectionLoader->setConnectionName((string) $connectionName);
+        } elseif ($input->getOption('conn')!==null) {
+            throw new InvalidOptionUsage('You can use the --conn option only if the connection loader is of type ConnectionRegistryConnection.');
+        }
+
+        $connectionLoader = $this->dependencyFactory->getEntityManagerLoader();
+        if ($connectionLoader instanceof ManagerRegistryEntityManager && $input->getOption('em')!==null) {
```
SenseException, line 108:
```suggestion
        if ($connectionLoader instanceof ManagerRegistryEntityManager && $input->getOption('em') !== null) {
```

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
```diff
@@ -75,6 +92,26 @@ protected function initialize(InputInterface $input, OutputInterface $output) :
...
+            $emName = $input->getOption('em');
+            $connectionLoader->setManagerName((string) $emName);
+        } elseif ($input->getOption('em')!==null) {
```
SenseException, line 111:
```suggestion
        } elseif ($input->getOption('em') !== null) {
```

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
```diff
@@ -75,6 +92,26 @@ protected function initialize(InputInterface $input, OutputInterface $output) :
...
+            $connectionName = $input->getOption('conn');
+            $connectionLoader->setConnectionName((string) $connectionName);
+        } elseif ($input->getOption('conn')!==null) {
+            throw new InvalidOptionUsage('You can use the --conn option only if the connection loader is of type ConnectionRegistryConnection.');
+        }
+
+        $connectionLoader = $this->dependencyFactory->getEntityManagerLoader();
+        if ($connectionLoader instanceof ManagerRegistryEntityManager && $input->getOption('em')!==null) {
+            $emName = $input->getOption('em');
+            $connectionLoader->setManagerName((string) $emName);
```
SenseException, line 110: At this point `$input->getOption('em')` can't return a null value.

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
```diff
@@ -75,6 +92,26 @@ protected function initialize(InputInterface $input, OutputInterface $output) :
...
         }
 
+        if ($input->getOption('em')!== null && $input->getOption('conn')!==null) {
+            throw new InvalidOptionUsage('You can specify only one of the --em and --conn options.');
+        }
+
+        $connectionLoader = $this->dependencyFactory->getConnectionLoader();
+        if ($connectionLoader instanceof ConnectionRegistryConnection && $input->getOption('conn')!==null) {
+            $connectionName = $input->getOption('conn');
+            $connectionLoader->setConnectionName((string) $connectionName);
```
SenseException, line 102: At this point `$input->getOption('conn')` can't return a null value.

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
goetas: That is true, but phpstan detects that can be `int|string`, so a casting to string is needed (i guess that happens if you name your entity manager something as `792`, passing `--em=792`, it seems that `792` will be an integer.... (i did not test it  but phpstan says so...)

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
goetas: same as https://github.com/doctrine/migrations/pull/1023#discussion_r446008824

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
```diff
@@ -75,6 +92,26 @@ protected function initialize(InputInterface $input, OutputInterface $output) :
             $this->dependencyFactory->setConfigurationLoader($configurationLoader);
         }
 
+        if ($input->getOption('em') !== null && $input->getOption('conn') !== null) {
+            throw new InvalidOptionUsage('You can specify only one of the --em and --conn options.');
+        }
+
+        $connectionLoader = $this->dependencyFactory->getConnectionLoader();
+        if ($connectionLoader instanceof ConnectionRegistryConnection && $input->getOption('conn') !== null) {
```
goetas, line 100: I'm not really happy about this conditionals based on concrete classes... any ideas on how to avoid this?

@greg0ire @SenseException ?

goetas:
Note for my self: there should be also a config parameter in the configurations allowing you to choose the entity manager or the connection

docs/en/reference/configuration.rst
```diff
@@ -136,6 +148,10 @@ Here are details about what each configuration option does:
 +----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
 | check_database_platform    | no         | true                         | Whether to add a database platform check at the beginning of the generated code. |
 +----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
+| connection                 | no         | null                         | The named connection to use (available only when ConnectionRegistryConnection is used). |
++----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
+| em                         | no         | nul                          | The named entity manager to use (available only when ManagerRegistryEntityManager is used). |
```
greg0ire, line 153:
```suggestion
| em                         | no         | null                         | The named entity manager to use (available only when ManagerRegistryEntityManager is used). |
```

docs/en/reference/configuration.rst
```diff
@@ -71,6 +76,10 @@ Now, in the root of your project place a file named ``migrations.php``, ``migrat
 
             <name>My Project Migrations</name>
 
+            <connection>default</connection>
+            <em>default</em>
```
greg0ire, line 80: Why is it not null in xml (or whatever the notation for `null` is)?

docs/en/reference/custom-configuration.rst
```diff
@@ -68,3 +68,62 @@ Once you have your custom integration setup, you can modify it to look like the
...
+    $cli->setCatchExceptions(true);
+
+    $cli->addCommands(array(
+        new Command\MigrateCommand($dependencyFactory),
+        // more commands here
+    ));
+
+    $cli->run();
+
+With this configurations you can use the ``--conn`` parameter to specify a connection that will be used for running
```
greg0ire, line 126:
```suggestion
With this configuration you can use the ``--conn`` parameter to specify a connection that will be used for running
```

docs/en/reference/custom-configuration.rst
```diff
@@ -68,3 +68,62 @@ Once you have your custom integration setup, you can modify it to look like the
...
+migrations. If the parameter is not passed, it will fallback to the one passed in the configurations,
```
greg0ire, line 127:
```suggestion
migrations. If the parameter is not passed, it will fallback to the one passed in the configuration,
```

docs/en/reference/custom-configuration.rst
```diff
@@ -68,3 +68,62 @@ Once you have your custom integration setup, you can modify it to look like the
...
+and if that is also not provided it will fallback to the default connection name specified when creting
```
greg0ire, line 128:
```suggestion
and if that is also not provided it will fallback to the default connection name specified when creating
```

docs/en/reference/custom-configuration.rst
```diff
@@ -68,3 +68,62 @@ Once you have your custom integration setup, you can modify it to look like the
...
+    use Doctrine\Migrations\Configuration\Configuration\ExistingConfiguration;
+    use Doctrine\Migrations\Configuration\Connection\ConnectionRegistryConnection;
+    use Doctrine\Migrations\DependencyFactory;
+    use Doctrine\Migrations\Tools\Console\Command;
+    use Symfony\Component\Console\Application;
+
+    $connection1 = DriverManager::getConnection([...]);
+    $connection2 = DriverManager::getConnection([...]);
+
+    $connectionRegistry = new class (
```
greg0ire, line 94: What does `class` represent here?

docs/en/reference/configuration.rst
goetas: that is the right question, in XML to have it `null` the tag should not be there... but for documentation proposes i had to put something.... and `default` is the default name for a connection or entity manager

docs/en/reference/custom-configuration.rst
goetas: good catch, i forgot the "extends" part, now I have added it

docs/en/reference/custom-configuration.rst
```diff
@@ -68,3 +68,62 @@ Once you have your custom integration setup, you can modify it to look like the
...
+    $connection2 = DriverManager::getConnection([...]);
+
+    $connectionRegistry = new class(
+        'some_registry',
+        ['foo' => $connection1, 'bar' => $connection2],
+        [], // entity managers
+        'foo', // default connection
+        null, // default entity manager
+        'Doctrine\Persistence\Proxy' // proxy class
+        ) extends AbstractManagerRegistry {
```
greg0ire, line 101: I think lines 100 and 101 shouldn't be indented the same way
```suggestion
    ) extends AbstractManagerRegistry {
```

AndyDune:
It is very important future for me.

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
SenseException: It's like `ConnectionRegistryConnection` isn't supposed to be a `ConnectionLoader`. The `DependencyFactory` is already too crowded for implementing new methods there.

goetas:
@SenseException 

> It's like ConnectionRegistryConnection isn't supposed to be a ConnectionLoader. And the DependencyFactory is already too crowded for implementing new methods there. I have currently no good ideas for this.

Yeah, ideally the connection `ConnectionLoader` should have a "name", passed as parameter.... I'm not sure if it is backward compatible (yes?) it feels as if the connection should not be part to the dependency factory...

SenseException:
> I'm not sure if it is backward compatible (yes?)

I guess it's not backward compatible, but `ConnectionLoader` is marked as internal (and probably out of BC promise?) and the classes that are implementing it are declared `final`. Is there a guide line for such a case?

goetas:
Hmm... that is a good point... (in reality that makes it confusing, `ConnectionLoader` probably should not be internal.... since that is how ppl can specify custom connections...)

goetas:
With https://github.com/doctrine/migrations/pull/1023/commits/9f75d67f3c34ba95755c314c1197a95b27a56019 i have refactored how connections are loaded. 

The biggest change is the following:

before
```php
/**
 * @internal
 */
interface ConnectionLoader
{
    public function getConnection() : Connection;
}
```
after
```php
interface ConnectionLoader
{
    public function getConnection(?string $name = null) : Connection;
}
```

The change is not backward compatible, but given:
-  the particularity of the feature
- the fact that was marked as `@internal` 
- the [not too high number of downloads](https://packagist.org/packages/doctrine/migrations/stats#major/3) 

I think that we could discuss the possibility  to release it as 3.1 instead of making a new major release just for this change.

What do you think?

cc @greg0ire @SenseException  @alcaeus

greg0ire:
If the changed API was marked as `@internal`, why are you saying the change isn't backward compatible?

SenseException:
I guess this is a fitting change for the interface.

> `ConnectionLoader` probably should not be internal.... since that is how ppl can specify custom connections...)

Should the interface stay `@internal`?

goetas:
> If the changed API was marked as `@internal`, why are you saying the change isn't backward compatible?

to be honest i got lost and i did not notice that was marked as `@internal`. IMO it should not have been marked so, but now is convenient.

> Should the interface stay `@internal`?

I think not, if ppl have weird connection setup, that interface is the way how they can provide the connection to the library.

It seems that both of you agree with proceeding with this, I will update tests and documentation by using the currently proposed approach.

SenseException:
It looks like it.

goetas:
After a lot of work, this is ready for review!

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
```diff
@@ -44,6 +48,20 @@ protected function configure() : void
             'The path to a migrations configuration file. <comment>[default: any of migrations.{php,xml,json,yml,yaml}]</comment>'
         );
 
+        $this->addOption(
+            'em',
+            null,
+            InputOption::VALUE_REQUIRED,
+            'The name of the connection to use (available only with ManagerRegistryEntityManager loader).'
```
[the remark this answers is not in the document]
greg0ire: This still looks switched to me.

lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
```diff
@@ -44,6 +45,20 @@ protected function configure() : void
...
+        );
+
+        $this->addOption(
+            'conn',
+            null,
+            InputOption::VALUE_REQUIRED,
+            'The name of the entity manager to use (available only with ConnectionRegistryConnection loader).'
```
andrew-demb, line 59: Looks like this message should be for --em option

goetas:
@andrew-demb @greg0ire You are right, i somehow missed that. Thanks!

lib/Doctrine/Migrations/Configuration/Connection/Exception/InvalidConfiguration.php
```diff
@@ -4,12 +4,20 @@
...
 final class InvalidConfiguration extends InvalidArgumentException implements LoaderException
 {
     public static function invalidArrayConfiguration() : self
     {
         return new self('The connection file has to return an array with database configuration parameters.');
     }
+
+    public static function invalidConnectionType(object $connection) : self
+    {
+        return new self(sprintf('The returned connection must be a %s connection, %s returned.', Connection::class, get_class($connection)));
```
greg0ire, line 21:
```suggestion
        return new self(sprintf('The returned connection must be a %s instance, %s returned.', Connection::class, get_class($connection)));
```

lib/Doctrine/Migrations/Configuration/EntityManager/Exception/InvalidConfiguration.php
```diff
@@ -4,12 +4,20 @@
...
 final class InvalidConfiguration extends InvalidArgumentException implements LoaderException
 {
     public static function invalidArrayConfiguration() : self
     {
         return new self('The EntityManager file has to return an array with database configuration parameters.');
     }
+
+    public static function invalidManagerType(object $em) : self
+    {
+        return new self(sprintf('The returned manager must be a %s entity manager, %s returned.', EntityManagerInterface::class, get_class($em)));
```
greg0ire, line 21:
```suggestion
        return new self(sprintf('The returned manager must implement %s, %s returned.', EntityManagerInterface::class, get_class($em)));
```

lib/Doctrine/Migrations/Configuration/Exception/InvalidLoader.php
```diff
@@ -0,0 +1,24 @@
...
+use function sprintf;
+
+final class InvalidLoader extends InvalidArgumentException implements ConfigurationException
+{
+    public static function noMultipleConnections(ConnectionLoader $loader) : self
+    {
+        return new self(sprintf('Only one connection is supported by %s', get_class($loader)));
+    }
+
+    public static function noMultipleEntitMangers(EntityManagerLoader $loader) : self
```
greg0ire, line 20:
```suggestion
    public static function noMultipleEntityManagers(EntityManagerLoader $loader) : self
```

lib/Doctrine/Migrations/Configuration/Migration/ConfigurationArray.php
```diff
@@ -41,6 +41,10 @@ public function getConfiguration() : Configuration
                     $configuration->addMigrationClass($className);
                 }
             },
+
+            'connection' => 'setConnectionName',
+            'em' => 'setEntityMangerName',
```
greg0ire, line 46:
```suggestion
            'em' => 'setEntityManagerName',
```

goetas:
All suggestions have been applied.

tests/Doctrine/Migrations/Tests/DependencyFactoryWithEntityManagerRegistryTest.php
```diff
@@ -0,0 +1,89 @@
...
+        $this->configuration = new Configuration();
+
+        $this->connection1 = $this->createMock(Connection::class);
+        $this->connection2 = $this->createMock(Connection::class);
+
+        $this->em1 = $this->createMock(EntityManager::class);
+        $this->em2 = $this->createMock(EntityManager::class);
+
+        $this->em1
+            ->expects(self::any())
```
greg0ire, line 51:
```suggestion
```

tests/Doctrine/Migrations/Tests/DependencyFactoryWithEntityManagerRegistryTest.php
```diff
@@ -0,0 +1,89 @@
...
+            ->method('getConnection')
+            ->willReturn($this->connection1);
+
+        $this->em2
+            ->expects(self::any())
```
greg0ire, line 56:
```suggestion
```

SenseException (approved):

greg0ire (approved):

PR #1028, opened 2020-06-25
Update MigrateCommand.php

ThomasLandauer:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

Adding the name of the database to CLI's yes/no question, analogous to https://github.com/doctrine/DoctrineFixturesBundle/blob/master/Command/LoadDataFixturesDoctrineCommand.php#L103

TODO: Do the same in `ExecuteCommand` and `RollupCommand`

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
@@ -105,7 +105,10 @@ EOT
         $migratorConfigurationFactory = $this->getDependencyFactory()->getConsoleInputMigratorConfigurationFactory();
         $migratorConfiguration        = $migratorConfigurationFactory->getMigratorConfiguration($input);
 
-        $question = sprintf('WARNING! You are about to execute a migration in database "%s" that could result in schema changes and data loss. Are you sure you wish to continue?', $this->getDependencyFactory()->getConnection()->getDatabase());
+        $question = sprintf(
+            'WARNING! You are about to execute a migration in database "%s" that could result in schema changes and data loss. Are you sure you wish to continue?',
+            $this->getDependencyFactory()->getConnection()->getDatabase()
+        );
         if (! $migratorConfiguration->isDryRun() && ! $this->canExecute($question, $input)) {
             $this->io->error('Migration cancelled!');
 
--- a/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
@@ -105,7 +105,7 @@ EOT
         $migratorConfigurationFactory = $this->getDependencyFactory()->getConsoleInputMigratorConfigurationFactory();
         $migratorConfiguration        = $migratorConfigurationFactory->getMigratorConfiguration($input);
 
-        $question = 'WARNING! You are about to execute a database migration that could result in schema changes and data loss. Are you sure you wish to continue?';
+        $question = sprintf('WARNING! You are about to execute a migration in database "%s" that could result in schema changes and data loss. Are you sure you wish to continue?', $this->getDependencyFactory()->getConnection()->getDatabase());
         if (! $migratorConfiguration->isDryRun() && ! $this->canExecute($question, $input)) {
             $this->io->error('Migration cancelled!');
 
--- a/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
@@ -107,7 +107,7 @@ EOT
 
         $question = sprintf(
             'WARNING! You are about to execute a migration in database "%s" that could result in schema changes and data loss. Are you sure you wish to continue?',
-            $this->getDependencyFactory()->getConnection()->getDatabase()
+            $this->getDependencyFactory()->getConnection()->getDatabase() ?? '<unnamed>'
         );
         if (! $migratorConfiguration->isDryRun() && ! $this->canExecute($question, $input)) {
             $this->io->error('Migration cancelled!');
--- a/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
@@ -121,7 +121,10 @@ EOT
         $migratorConfigurationFactory = $this->getDependencyFactory()->getConsoleInputMigratorConfigurationFactory();
         $migratorConfiguration        = $migratorConfigurationFactory->getMigratorConfiguration($input);
 
-        $question = sprintf('WARNING! You are about to execute a migration in database "%s" that could result in schema changes and data loss. Are you sure you wish to continue?', $this->getDependencyFactory()->getConnection()->getDatabase());
+        $question = sprintf(
+            'WARNING! You are about to execute a migration in database "%s" that could result in schema changes and data loss. Are you sure you wish to continue?',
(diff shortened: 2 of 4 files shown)


lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
```diff
@@ -121,7 +121,7 @@ protected function execute(InputInterface $input, OutputInterface $output) : int
         $migratorConfigurationFactory = $this->getDependencyFactory()->getConsoleInputMigratorConfigurationFactory();
         $migratorConfiguration        = $migratorConfigurationFactory->getMigratorConfiguration($input);
 
-        $question = 'WARNING! You are about to execute a database migration that could result in schema changes and data loss. Are you sure you wish to continue?';
+        $question = sprintf('WARNING! You are about to execute a migration in database "%s" that could result in schema changes and data loss. Are you sure you wish to continue?', $this->getDependencyFactory()->getConnection()->getDatabase());
```
goetas, line 124: `getDatabase` is not part of the `\Doctrine\DBAL\Driver\Connection` interface, thus i'm not sure if we can do this...

@greg0ire ?

greg0ire: :thinking: but `getConnection` returns a `\Doctrine\DBAL\Connection` instead of that interface, right?

goetas: hmm. right. being so, we know the database... (i remember that there was a reason for choosing the concrete implementation, but can't recall why :-( )

ThomasLandauer: In doctrine/migrations **2.2.1** I'm getting:
> In MigrateCommand.php line 151:
> Attempted to call an undefined method named "getDependencyFactory" of class "Doctrine\Bundle\MigrationsBundle\Command\MigrationsMigrateDoctrineCommand". Did you mean to call "setDependencyFactory"?

But in **3.0.1** `get_class($this->getDependencyFactory()->getConnection())` gives me:
> Doctrine\DBAL\Connection

goetas: the command `getDependencyFactory` has been introduced in 3.0

goetas: @greg0ire I'm not sure if is worth this, it would be nice to change the return type from `\Doctrine\DBAL\Driver\Connection` to `\Doctrine\DBAL\Connection`..

greg0ire: That wouldn't be a BC break, except for extending classes, so… go ahead?

goetas: May be I'm worrying too much. Probably the improved dx is worth this change.

@ThomasLandauer  can you do the same change on the other commands?

ThomasLandauer: Sure, done :-) (I just copied it over - since they all extend `DoctrineCommand`)

@goetas: What means "DX"? Developer Experience - as opposed to User Experience (UX)?

goetas: @ThomasLandauer yes, DX means Developer Experience

goetas:
@ThomasLandauer tests are failing 

Sqlite database name is empty.... what should we put in that case? the DB path? (is it possible with the current DBAL api?

ThomasLandauer:
Hm, could it be that the failure is an artifact of your testing environment? Cause I am getting the full path to my `.sqlite` file! (same as in fixtures bundle)...

lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
```diff
@@ -105,7 +105,7 @@ protected function execute(InputInterface $input, OutputInterface $output) : int
         $migratorConfigurationFactory = $this->getDependencyFactory()->getConsoleInputMigratorConfigurationFactory();
         $migratorConfiguration        = $migratorConfigurationFactory->getMigratorConfiguration($input);
 
-        $question = 'WARNING! You are about to execute a database migration that could result in schema changes and data loss. Are you sure you wish to continue?';
+        $question = sprintf('WARNING! You are about to execute a migration in database "%s" that could result in schema changes and data loss. Are you sure you wish to continue?', $this->getDependencyFactory()->getConnection()->getDatabase());
```
greg0ire, line 108:
```suggestion
        $question = sprintf(
            'WARNING! You are about to execute a migration in database "%s" that could result in schema changes and data loss. Are you sure you wish to continue?',
            $this->getDependencyFactory()->getConnection()->getDatabase()
        );
```

lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
greg0ire, line 124:
```suggestion
        $question = sprintf(
            'WARNING! You are about to execute a migration in database "%s" that could result in schema changes and data loss. Are you sure you wish to continue?',
            $this->getDependencyFactory()->getConnection()->getDatabase()
        );
```

lib/Doctrine/Migrations/Tools/Console/Command/RollupCommand.php
```diff
@@ -37,7 +37,7 @@ protected function configure() : void
 
     protected function execute(InputInterface $input, OutputInterface $output) : int
     {
-        $question = 'WARNING! You are about to execute a database migration that could result in schema changes and data loss. Are you sure you wish to continue?';
+        $question = sprintf('WARNING! You are about to execute a migration in database "%s" that could result in schema changes and data loss. Are you sure you wish to continue?', $this->getDependencyFactory()->getConnection()->getDatabase());
```
greg0ire, line 40:
```suggestion
        $question = sprintf(
            'WARNING! You are about to execute a migration in database "%s" that could result in schema changes and data loss. Are you sure you wish to continue?',
            $this->getDependencyFactory()->getConnection()->getDatabase()
        );
```

goetas:
Tests are failing with a message as 
`You are about to execute a migration in database "" that could result in schema changes and data loss` 
As I was saying in one of my previous comments, for SQLite (with in-memory storage option) the database name seems to be empty)

ThomasLandauer:
OK, I see. So what should we do in that case? Display some constant string, e.g. "SQLite :memory:"? It looks like FixturesBundle isn't doing anything special, so it's probably also just showing `""`: https://github.com/doctrine/DoctrineFixturesBundle/blob/master/Command/LoadDataFixturesDoctrineCommand.php#L103

goetas:
@ThomasLandauer what about `$this->getDependencyFactory()->getConnection()->getDatabase() ?? 'unknown'` ?

ThomasLandauer:
I changed it to ` ?? '<unnamed>'`, to emphasize that this is not an error but the database just doesn't have a "name". The tests will probably still fail, since they expect another string. Should I go ahead and adjust them?

goetas:
@ThomasLandauer i'm fine with `<unnamed>`, can you please update the tests then?

ThomasLandauer:
Done :-) BTW: I only updated `MigrateCommandTest.php`; there seem to be no tests for the other two commands.

goetas (approved):

goetas:
@greg0ire I guess that for this we can do squash merge

greg0ire:
Sure! Done.

Issue #1037, opened 2020-07-12
False Negative Test Failure on OSX: RecursiveRegexFinderTest::testFindMigrationsReturnsTheExpectedFilesFromDirectory


chrisguitarguy:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | b65eb7cc2cc0d569d21222ba343ee7298bb180d7

#### Summary & Current Behavior

`RecursiveRegexFinderTest::testFindMigrationsReturnsTheExpectedFilesFromDirectory` does this:

https://github.com/doctrine/migrations/blob/b65eb7cc2cc0d569d21222ba343ee7298bb180d7/tests/Doctrine/Migrations/Tests/Finder/RecursiveRegexFinderTest.php#L45-L47

Which evaluates to true on on OSX because `PHP_OS` comes back as `Darwin` (on OSX 10.13.6 anyway).

Don't think this is a huge deal, but I had to make sure I wasn't going crazy trying to a fix another bug and thinking I somehow broke code I didn't touch.

#### How to reproduce

 Run the tests on OSX

#### Expected behavior

Test passes

goetas:
Hi thanks for posting this issue, I'm really not familiar with OSX, what would be the solution here?

chrisguitarguy:
@goetas I think checking `PHP_OS_FAMILY` instead of `PHP_OS` would work. Sent a PR for that.

Possible alternative might be to use PHPUnits `@requires OSFAMILY ...` annotation, but it's not clear if something like `@requires OSFAMILY !Windows`

chrisguitarguy:
This is in https://github.com/doctrine/migrations/pull/1039

PR #1039, opened 2020-07-13
Check PHP_OS_FAMILY for Windows in Tests

chrisguitarguy:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1037 

#### Summary

Instead of checking `PHP_OS` for `Win`. The old version was matching
`Darwin` on OSX and causing a false negative.


greg0ire:
Thanks @chrisguitarguy !

PR #1043, opened 2020-08-09
Set dependencyFactory before configure is called in DoctrineCommand

jkrzefski:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

While reading through the code I found [this piece here](https://github.com/doctrine/migrations/blob/b65eb7cc2cc0d569d21222ba343ee7298bb180d7/lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php#L47-L49) and I figured, it would never actually do anything with the way things are set up. The constructor always sets the value for `$dependencyFactory` only [after calling its parent](https://github.com/doctrine/migrations/blob/b65eb7cc2cc0d569d21222ba343ee7298bb180d7/lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php#L34-L35), so until then `$dependencyFactory` is considered `null`. The parent's constructor in turn calls the `configure` method which checks if `$dependencyFactory` is `null`.

I have switched the assignment of `$dependencyFactory` with the call to the parent's constructor, so that the condition in the `configure` method actually does something depending on the presence of a `$dependencyFactory`.

I consider this a breaking change, since this can potentially lead to one less input option for a doctrine command, thus changing the command definition and breaking scripts that provide this option.

### DIFF
--- a/tests/Doctrine/Migrations/Tests/Tools/Console/Command/DoctrineCommandTest.php
+++ b/tests/Doctrine/Migrations/Tests/Tools/Console/Command/DoctrineCommandTest.php
@@ -73,4 +73,18 @@ class DoctrineCommandTest extends MigrationTestCase
             ['interactive' => false]
         );
     }
+
+    public function testDependencyFactoryIsSetFirst(): void
+    {
+        $dependencyFactory = $this->createMock(DependencyFactory::class);
+        $command           = new class ($dependencyFactory) extends DoctrineCommand
+        {
+            protected function configure(): void
+            {
+                $this->getDependencyFactory();
+            }
+        };
+
+        self::assertFalse($command->getDefinition()->hasOption('db-configuration'));
+    }
 }
--- a/lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php
@@ -32,8 +32,8 @@ abstract class DoctrineCommand extends Command
 
     public function __construct(?DependencyFactory $dependencyFactory = null, ?string $name = null)
     {
-        parent::__construct($name);
         $this->dependencyFactory = $dependencyFactory;
+        parent::__construct($name);
     }
 
     protected function configure(): void


greg0ire:
Nice find! Can you please add a test here: https://github.com/doctrine/migrations/blob/3.0.x/tests/Doctrine/Migrations/Tests/Tools/Console/Command/DoctrineCommandTest.php ?

jkrzefski:
@greg0ire I added a test case and made sure it fails without my change. Please take a look at it. And thank you for your patience.

tests/Doctrine/Migrations/Tests/Tools/Console/Command/DoctrineCommandTest.php
```diff
@@ -72,4 +73,27 @@ protected function execute(InputInterface $input, OutputInterface $output) : int
...
         );
     }
+
+    public function testDependencyFactory() : void
+    {
+        $exceptionThrown = false;
+
+        $dependencyFactory = $this->getMockBuilder(DependencyFactory::class)
+            ->disableOriginalConstructor()
+            ->getMock();
```
greg0ire:
```suggestion
        $dependencyFactory = $this->createMock(DependencyFactory::class);
```

jkrzefski: Done.

greg0ire:

How to do that?

1. `git rebase -i origin/3.0.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

greg0ire (approved):

jkrzefski:
Rebased and squashed.

tests/Doctrine/Migrations/Tests/Tools/Console/Command/DoctrineCommandTest.php
```diff
@@ -72,4 +73,25 @@ protected function execute(InputInterface $input, OutputInterface $output) : int
...
+        try {
+            new class ($dependencyFactory) extends DoctrineCommand
+            {
+                protected function configure() : void
+                {
+                    $this->getDependencyFactory();
+                }
+            };
+        } catch (DependenciesNotSatisfied $exception) {
+            $exceptionThrown = true;
```
SenseException: If not having an Exception is the expected result, I think you don't need the whole try-catch construct. It's true that the fail-message is expressive, but you can put the expected behaviour in the name of the test method like `testDependencyFactoryIsInitialized()` for an also expressive message on a failing test.

SenseException (changes requested):

tests/Doctrine/Migrations/Tests/Tools/Console/Command/DoctrineCommandTest.php
jkrzefski: So do you suggest to drop the assertion completely? My original attempt was to have no assertion and just let the exception trigger the error. But that would be a test without any assertion (PHPUnit seems to not like those).

Then I tried an assertion at the end of it with `assertTrue(true)`, but static analysis was not happy about it. It said, this assertion would always be true.

So that's what lead to the decision to use a try-catch-contraption. If you have a better way, I will gladly implement it.

greg0ire: You could assert that the `db-configuration` option is available in the command when a dependency factory is provided.

> I consider this a breaking change, since this can potentially lead to one less input option for a doctrine command, thus changing the command definition and breaking scripts that provide this option.

:thinking: your change can only result in that option being available, so, one more input option, not one fewer… I don't understand. This is the block that is enabled with your PR: https://github.com/doctrine/migrations/blob/b65eb7cc2cc0d569d21222ba343ee7298bb180d7/lib/Doctrine/Migrations/Tools/Console/Command/DoctrineCommand.php#L51-L57

SenseException: > You could assert that the db-configuration option is available in the command when a dependency factory is provided.

This should also test the availability of the `db-configuration` option depending on the dependency factory and covers the logic when the option is set.

goetas:
@SenseException I have rebased it and added the `db-configuration` check.

goetas (approved):

greg0ire (approved):

tests/Doctrine/Migrations/Tests/Tools/Console/Command/DoctrineCommandTest.php
```diff
@@ -73,4 +74,27 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+                protected function configure(): void
+                {
+                    $this->getDependencyFactory();
+                }
+            };
+
+            self::assertFalse($command->getDefinition()->hasOption('db-configuration'));
+        } catch (DependenciesNotSatisfied $exception) {
+            $exceptionThrown = true;
+        }
```
SenseException: What I meant in https://github.com/doctrine/migrations/pull/1043/files#r484106894

```suggestion
        $command = new class ($dependencyFactory) extends DoctrineCommand
        {
            protected function configure(): void
            {
                $this->getDependencyFactory();
            }
        };

        self::assertFalse($command->getDefinition()->hasOption('db-configuration'));
```

SenseException (changes requested):

goetas:
@SenseException  makes sense! code updated

SenseException (approved):

goetas:
thanks everyone!

Issue #1047, opened 2020-08-17
Diff command without ORM


veewee:
### Diff command without ORM


|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

Currently the diff command is only added when an entity manager is available:

https://github.com/doctrine/migrations/blob/a04b141ee90d630d7b86e7334ebbb77b51a3275a/lib/Doctrine/Migrations/Tools/Console/ConsoleRunner.php#L125-L129

However, the only reason why the diff command could not work is because the `DependencyFactory` does not know how to build the desired schema:

https://github.com/doctrine/migrations/blob/a04b141ee90d630d7b86e7334ebbb77b51a3275a/lib/Doctrine/Migrations/DependencyFactory.php#L238-L243


If you allow a way to insert a custom schema provider in there, you could also make the diff command work for regular DBAL implementations.


How you can implement this in your codebase:

1.  Application logic outide of this package : Add e.g. a table builder like this:

```php
use Doctrine\DBAL\Schema\Table;
use Doctrine\DBAL\Types\Types;

final class SomeTable
{
    public const NAME = 'some_table';

    public static function createTable(): Table
    {
        $table = new Table(self::NAME);

        $table->addOption('charset', 'utf8mb4');
        $table->addOption('collate', 'utf8mb4_unicode_ci');
        $table->addOption('engine', 'InnoDB');

        $table->addColumn('id', Types::GUID, [
            'notnull' => true,
        ]);

        // ....

        $table->setPrimaryKey(['id']);

        return $table;
    }
}
```

(Extra benefit for implementing these table schema's inside your codebase : you can use these table objects as source for e.g. database integration tests, making sure they are always up to date)

2. Inside this package : Create a custom schema provider, which basically looks like this (and could be loaded from the application with e.g. tagged services or a directory loader):

```php
class DependencyFactory
{
// ....

    private function getSchemaProvider() : SchemaProvider
    {
        return $this->getDependency(SchemaProvider::class, function (): SchemaProvider {
            return new class implements SchemaProvider
            {
                public function createSchema(): Schema
                {
                    $schema = new Schema(
                        [
                            SomeTable::createTable(),
                        ]
                    );

                    return $schema;
                }
            };
        });
    }

// ....
}
```

The changes above result in:

```sh
./bin/console migrations:diff                                                                                                                                                                                                      

// ....

 Generated new migration class to "/.../migrations/Version20200817092354.php"

// ....
[...]

PR #1050, opened 2020-08-28
Add a schema_provider option that enables the diff command for DBAL

veewee:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #1047

#### Summary

This change makes it possible to configure a `schema_provider` that enables the diff command for doctrine DBAL as well.

Currently the PR allows you to do:

```php
$dependencyFactory = DependencyFactory::fromConnection(
    new ConfigurationArray([...]),
    new ExistingConnection($connection)
);
$dependencyFactory->setService(SchemaProvider::class, new MySchemaProvider());
MigrationsConsoleRunner::addCommands($application, $dependencyFactory);
```

In which, `MySchemaProvider` is a `Doctrine\Migrations\Provider\SchemaProvider`.

### DIFF
--- a/lib/Doctrine/Migrations/DependencyFactory.php
+++ b/lib/Doctrine/Migrations/DependencyFactory.php
@@ -246,10 +246,25 @@ class DependencyFactory
         });
     }
 
-    private function getSchemaProvider(): SchemaProvider
+    public function hasSchemaProvider(): bool
+    {
+        try {
+            $this->getSchemaProvider();
+        } catch (MissingDependency $exception) {
+            return false;
+        }
+
+        return true;
+    }
+
+    public function getSchemaProvider(): SchemaProvider
     {
         return $this->getDependency(SchemaProvider::class, function (): SchemaProvider {
-            return new OrmSchemaProvider($this->getEntityManager());
+            if ($this->hasEntityManager()) {
+                return new OrmSchemaProvider($this->getEntityManager());
+            }
+
+            throw MissingDependency::noSchemaProvider();
         });
     }
 
--- a/tests/Doctrine/Migrations/Tests/DependencyFactoryTest.php
+++ b/tests/Doctrine/Migrations/Tests/DependencyFactoryTest.php
@@ -16,6 +16,8 @@ use Doctrine\Migrations\Exception\MissingDependency;
 use Doctrine\Migrations\Finder\GlobFinder;
 use Doctrine\Migrations\Finder\RecursiveRegexFinder;
 use Doctrine\Migrations\Metadata\Storage\TableMetadataStorageConfiguration;
+use Doctrine\Migrations\Provider\OrmSchemaProvider;
+use Doctrine\Migrations\Provider\SchemaProvider;
 use Doctrine\ORM\EntityManager;
 use PHPUnit\Framework\MockObject\MockObject;
 use Psr\Log\LoggerInterface;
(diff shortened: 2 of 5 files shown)


lib/Doctrine/Migrations/DependencyFactory.php
```diff
@@ -235,10 +235,31 @@ public function getSchemaDumper() : SchemaDumper
...
+        }
+
+        return false;
+    }
+
+    public function getSchemaProvider() : SchemaProvider
     {
         return $this->getDependency(SchemaProvider::class, function () : SchemaProvider {
-            return new OrmSchemaProvider($this->getEntityManager());
+            if ($this->hasEntityManager()) {
```
andrew-demb, line 263: Maybe use custom provided schema provider, when it is explicitly specified and fallback to check ORM otherwise?

WDYT?

veewee: Implementations will most likely choose either ORM or a custom schema for DBAL. But I agree that the explicit configured one should have higher priority.

tests/Doctrine/Migrations/Tests/Configuration/Migration/ArrayLoaderTest.php
```diff
@@ -17,4 +20,42 @@ public function testInvalidKey() : void
...
+        $loader         = new ConfigurationArray(['schema_provider' => get_class($schemaProvider)]);
+        $config         = $loader->getConfiguration();
+
+        self::assertInstanceOf(SchemaProvider::class, $config->getSchemaProvider());
+    }
+
+    public function testUnknownSchemaProvider() : void
+    {
+        $this->expectException(UnknownConfigurationValue::class);
+        $loader = new ConfigurationArray(['schema_provider' => 42]);
```
SenseException: For a complete coverage every check in the if block that leads to the exception would need to be tested.

SenseException:
The code looks good for me so far. For the feature itself @doctrine/team-migrations should take a look first.

tests/Doctrine/Migrations/Tests/Configuration/Migration/ArrayLoaderTest.php
veewee: I am not sure what you are asking from me. The tests above cover all the allowed types. This test covers a non-supported type. Can you elaborate?

docs/en/reference/configuration.rst
```diff
@@ -125,7 +125,10 @@ Here are details about what each configuration option does:
 +----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
 | check_database_platform    | no         | true                         | Whether to add a database platform check at the beginning of the generated code. |
 +----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
+| schema_provider            | no         | null                         | If you want to use DBAL instead of ORM, you can specify a target schema to which you want to generate diffs.
++----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
 
+For DBAL, you can specify a ```schema_provider`` by providing a ``SchemaProvider`` instance, a ``callable`` that results into a schema provider or a ``string`` className that can be used to instantiate a SchemaProvider instance.
```
fabpot: typo: Triple ` before schema_provider

tests/Doctrine/Migrations/Tests/DependencyFactoryTest.php
```diff
@@ -125,6 +127,46 @@ public function testEntityManager() : void
...
+        $schemaProvider = $this->createMock(SchemaProvider::class);
+        $this->configuration->setSchemaProvider($schemaProvider);
+        $di = DependencyFactory::fromConnection(new ExistingConfiguration($this->configuration), new ExistingConnection($this->connection));
+
+        self::assertTrue($di->hasSchemaProvider());
+        self::assertSame($schemaProvider, $di->getSchemaProvider());
+        self::assertTrue($di->isFrozen());
+    }
+
+    public function testSchemaProviderWithEntityManagerConfigShouldFallBackToTheExplicitlyProvidedProvider() : void
```
greg0ire: That's not really a fallback mechanism if the schema provider from the entity manager exists.

```suggestion
    public function testProvidedProviderHasPrecedenceOverProviderFromEntityManagerConfig() : void
```

docs/en/reference/configuration.rst
veewee: A yes, thanks!

goetas:
Hello. Thanks for working on this. 

IMO a schema provider should not be a configuration option but something passed to the dependency container.

Having a callback as configuration value makes configuration non cache-able and impossible to express in XML, JSON or YAML.

What about marking as public `getSchemaProvider` in the dep-factory and make the needed adjustments to make it work?
I'm also in favor of deprecating/marking-private the `getEntityManager`, since that method should not be used anywhere outside of the dependency factory.

goetas:
Actually looking better, things are "almost" working. 


```php
$di->setService(SchemaProvider::class, $someSchemaProvider);
``` 
will already do the job.
the only thing that needs to be done is to expose the `diff` command even if the entity manager is not available

veewee:
If you don't add it through configuration - it has this double way of configuring the tool. It needs proper documentation then.

Shall I give the configuration of the dependency directly another try in this PR or do you want to discuss it first?

goetas:
> If you don't add it through configuration - it has this double way of configuring the tool. It needs proper documentation then.

Indeed it should be documented, I've added in the past how to provide custom entity managers but did not think on the possibility of providing only the schema provider (and I think it is a really good idea for doing so).

I want to distinguish things that are configurations and things that are dependencies. Anything that can not be expressed in all the configuration formats (yaml, xml, json and php) can not go in the configurations  pool.
that is the reason why you can not provide a connection via configuration (creating a connection might need access to php dependencies that can not be expressed in the configuration itself).
Later framework integrations (that are able to manage dependencies) can provide a custom dependencies in the form that each frameworks uses to express them (symfony uses its own dependency injection configured with xml or yaml,but laravel or laminas use just plain php)

goetas:
In 2.x, the configuration object was huge.... mainly because was managing both dependencies and configurations

veewee:
Ok, thanks for the side information! Will give it a try soon :)

veewee:
@goetas  : I removed the configuration part. Where do you want me to document how to configure a custom schema provider?

veewee:
Is there anything I can do in order to get this merged? Thanks!

goetas:
@veewee i'm sorry for the delay, I got some time back to look into this. Could you please rebase it? 
It looks good to me, and i hope that this PR makes its way already in 3.1

veewee:
@goetas : rebased ;)

goetas (approved):

goetas:
I've changed the base branch for this to `3.1.x` since it adds a new feature. 

Can @SenseException  or @greg0ire review this please?

SenseException (approved):

goetas:
Thanks for your work

PR #1052, opened 2020-08-31
Add PHPDoc blocks to AbstractMigration

koninka:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Add PHPDoc blocks to AbstractMigration methods. This could be useful for those who use `@inheritdoc`annotation (like me:)).


greg0ire:
Why would it be useful?

koninka:
@greg0ire 

Hey,
1. It's good practice to document your code.
2. Having a description for the methods makes it easy to understand what they do. For example, the `write` and `warnIf` methods are not immediately obvious without additional code reading (not only in the `AbstractMigration` class).
3. My case. I always document my code and use `@inheritdoc` for overridden vendor methods (e.g. `getDescription`). And if I use `@inheritdoc` and parent method has no phpdocs, then this is at least strange (also PHPStorm reports warning).

You have a description of the methods on the [site](https://www.doctrine-project.org/projects/doctrine-migrations/en/3.0/reference/migration-classes.html), why not place the documentation in the code to save programmers from the need to google or research the code in depth to understand its work?

greg0ire:
1. It is, but documentation has to bring value, otherwise you risk people not reading useful comment because most comments in your project are just noise. I'd argue that's the case of most comments in this PR.
2. That's the part of the PR that might bring some value. If I were to merge anything here, that would be it.
3. That's really personal, sorry but we are not going to pollute the code with tautological comments. If you want to add comments, they should bring value.

> You have a description of the methods on the site, why not place the documentation in the code to save programmers from the need to google or research the code in depth to understand its work?

That's a good point, maybe there is valuable information here that could be fit in the comments. `Description of this migration` is not valuable information IMO. So is information already contained in type declarations.

koninka:
@greg0ire 

I got it. So, do I understand correctly, if I add a more useful description for the `getDescription`, then you will merge pr? Do I need to add a more detailed description to other methods?

P.S. I took a description of all methods from the official site (for method [`getDescription`](https://www.doctrine-project.org/projects/doctrine-migrations/en/3.0/reference/migration-classes.html#getdescription) too).

lib/Doctrine/Migrations/AbstractMigration.php
```diff
@@ -60,11 +60,22 @@ public function isTransactional() : bool
         return true;
     }
 
+    /**
+     * Description of this migration.
```
greg0ire, line 64: Tautological, either improve or remove

lib/Doctrine/Migrations/AbstractMigration.php
```diff
@@ -60,11 +60,22 @@ public function isTransactional() : bool
...
+     *
+     * @return string
```
greg0ire, line 66: Nothing new here, let's remove it
```suggestion
```

lib/Doctrine/Migrations/AbstractMigration.php
```diff
@@ -60,11 +60,22 @@ public function isTransactional() : bool
...
     public function getDescription() : string
     {
         return '';
     }
 
+    /**
+     * Warn with a message if some condition is met.
+     *
+     * @param bool   $condition
+     * @param string $message
```
greg0ire, line 77:
```suggestion
```

lib/Doctrine/Migrations/AbstractMigration.php
```diff
@@ -60,11 +60,22 @@ public function isTransactional() : bool
...
+     *
+     * @return string
+     */
     public function getDescription() : string
     {
         return '';
     }
 
+    /**
+     * Warn with a message if some condition is met.
```
greg0ire, line 74: That's made fairly clear just by reading the signature (you don't have to read the code inside the method)

lib/Doctrine/Migrations/AbstractMigration.php
```diff
@@ -155,6 +166,11 @@ public function getSql() : array
         return $this->plannedSql;
     }
 
+    /**
+     * Write some debug information to the console.
```
greg0ire, line 170: debug information is written with `debug()`, this is a notice.

lib/Doctrine/Migrations/AbstractMigration.php
```diff
@@ -155,6 +166,11 @@ public function getSql() : array
...
+     *
+     * @param string $message
```
greg0ire, line 172:
```suggestion
```

phpcs.xml.dist
```diff
@@ -42,4 +42,12 @@
...
     </rule>
 
+    <rule ref="SlevomatCodingStandard.TypeHints.ParameterTypeHint.UselessAnnotation">
+        <exclude-pattern>lib/Doctrine/Migrations/AbstractMigration.php</exclude-pattern>
+    </rule>
+
+    <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint.UselessAnnotation">
+        <exclude-pattern>lib/Doctrine/Migrations/AbstractMigration.php</exclude-pattern>
+    </rule>
+
```
greg0ire, line 52: Please revert this.

greg0ire (changes requested):

greg0ire:
> So, do I understand correctly, if I add a more useful description for the getDescription, then you will merge pr?

It's the opposite actually, if we merge the PR, then it means you will have added a more useful description for that method, but that will probably not be enough, since other methods need a better description too.

lib/Doctrine/Migrations/AbstractMigration.php
koninka: Fixed

lib/Doctrine/Migrations/AbstractMigration.php
koninka: Fixed

lib/Doctrine/Migrations/AbstractMigration.php
koninka: Fixed

lib/Doctrine/Migrations/AbstractMigration.php
koninka: Fixed

lib/Doctrine/Migrations/AbstractMigration.php
koninka: Fixed

lib/Doctrine/Migrations/AbstractMigration.php
koninka: Fixed

phpcs.xml.dist
koninka: Fixed

lib/Doctrine/Migrations/AbstractMigration.php
```diff
@@ -60,11 +60,19 @@ public function isTransactional() : bool
         return true;
     }
 
+    /**
+     * Description of this migration.
+     *
+     * Describe what this migration does in simple terms. This information is displayed when you view the list of migrations.
+     */
```
greg0ire, line 67:
```suggestion
    /**
     * Describe what this migration does in simple terms. This information is displayed when you view the list of migrations.
     */
```

lib/Doctrine/Migrations/AbstractMigration.php
```diff
@@ -155,6 +163,11 @@ public function getSql() : array
         return $this->plannedSql;
     }
 
+    /**
+     * Write some debug information to the console.
+     *
+     * Debug information is written with debug().
```
greg0ire, line 169: :thinking: I mean it should, but it's actually written with `notice()`, so that comment is just misleading

PR #1054, opened 2020-09-16
Fix circular ref in dependency resolution

goetas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

#### Summary

Ensure dependency resolution works even when using as dependency the default service definition

### DIFF
--- a/lib/Doctrine/Migrations/DependencyFactory.php
+++ b/lib/Doctrine/Migrations/DependencyFactory.php
@@ -59,6 +59,9 @@ use function sprintf;
  */
 class DependencyFactory
 {
+    /** @psalm-var array<string, bool> */
+    private $inResolution = [];
+
     /** @var Configuration */
     private $configuration;
 
@@ -476,8 +479,10 @@ class DependencyFactory
      */
     private function getDependency(string $id, callable $callback)
     {
-        if (array_key_exists($id, $this->factories) && ! array_key_exists($id, $this->dependencies)) {
+        if (! isset($this->inResolution[$id]) && array_key_exists($id, $this->factories) && ! array_key_exists($id, $this->dependencies)) {
+            $this->inResolution[$id] = true;
             $this->dependencies[$id] = call_user_func($this->factories[$id], $this);
+            unset($this->inResolution);
         }
 
         if (! array_key_exists($id, $this->dependencies)) {
--- a/tests/Doctrine/Migrations/Tests/DependencyFactoryTest.php
+++ b/tests/Doctrine/Migrations/Tests/DependencyFactoryTest.php
@@ -16,6 +16,9 @@ use Doctrine\Migrations\Exception\MissingDependency;
 use Doctrine\Migrations\Finder\GlobFinder;
 use Doctrine\Migrations\Finder\RecursiveRegexFinder;
 use Doctrine\Migrations\Metadata\Storage\TableMetadataStorageConfiguration;
+use Doctrine\Migrations\Tests\MigrationRepository\Migrations\A\A;
+use Doctrine\Migrations\Tests\Stub\CustomClassNameMigrationFactory;
+use Doctrine\Migrations\Version\MigrationFactory;
 use Doctrine\ORM\EntityManager;
 use PHPUnit\Framework\MockObject\MockObject;
 use Psr\Log\LoggerInterface;
@@ -154,6 +157,23 @@ final class DependencyFactoryTest extends MigrationTestCase
             return $anotherLogger;
         });
         self::assertSame($anotherLogger, $di->getLogger());
+    }
+
+    public function testServiceDecoratesDefaultImplementation(): void
+    {
+        $configuration = new Configuration();
+        $configuration->addMigrationClass(A::class);
+
+        $di = DependencyFactory::fromConnection(new ExistingConfiguration($configuration), new ExistingConnection($this->connection));
+
+        $di->setDefinition(MigrationFactory::class, static function (DependencyFactory $innerDi): CustomClassNameMigrationFactory {
+            $serviceToDecorate = $innerDi->getMigrationFactory();
+
+            return new CustomClassNameMigrationFactory($serviceToDecorate, A::class);
+        });
+
+        $factory = $di->getMigrationFactory();
+        self::assertInstanceOf(A::class, $factory->createVersion('SomeVersion'));
     }
 
     public function testServiceHasPriorityOverDefinition(): void
--- a/tests/Doctrine/Migrations/Tests/Stub/CustomClassNameMigrationFactory.php
+++ b/tests/Doctrine/Migrations/Tests/Stub/CustomClassNameMigrationFactory.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Migrations\Tests\Stub;
+
+use Doctrine\Migrations\AbstractMigration;
+use Doctrine\Migrations\Version\MigrationFactory;
+
+class CustomClassNameMigrationFactory implements MigrationFactory
+{
+    /** @var MigrationFactory */
+    private $parentMigrationFactory;
+
+    /** @psalm-var class-string<AbstractMigration> */
+    private $migrationClassName;
+
+    /**
+     * @param class-string<AbstractMigration> $migrationClassName
+     */
+    public function __construct(MigrationFactory $parentMigrationFactory, string $migrationClassName)
+    {
+        $this->parentMigrationFactory = $parentMigrationFactory;
+        $this->migrationClassName     = $migrationClassName;
+    }
+
+    public function createVersion(string $migrationClassName): AbstractMigration
+    {
+        return $this->parentMigrationFactory->createVersion($this->migrationClassName);
+    }
+}


tests/Doctrine/Migrations/Tests/DependencyFactoryTest.php
```diff
@@ -156,6 +158,36 @@ public function testServiceDefinition() : void
         self::assertSame($anotherLogger, $di->getLogger());
     }
 
+    public function testServiceDefinitionWithReference() : void
+    {
+        $migration = $this->createMock(AbstractMigration::class);
+
+        $di = DependencyFactory::fromConnection(new ExistingConfiguration($this->configuration), new ExistingConnection($this->connection));
+
+        $di->setDefinition(MigrationFactory::class, static function (DependencyFactory $innerDi) use ($migration) {
```
greg0ire:
```suggestion
        $di->setDefinition(MigrationFactory::class, static function (DependencyFactory $innerDi) use ($migration): MigrationFactory {
```

tests/Doctrine/Migrations/Tests/DependencyFactoryTest.php
```diff
@@ -156,6 +158,36 @@ public function testServiceDefinition() : void
         self::assertSame($anotherLogger, $di->getLogger());
     }
 
+    public function testServiceDefinitionWithReference() : void
+    {
+        $migration = $this->createMock(AbstractMigration::class);
+
+        $di = DependencyFactory::fromConnection(new ExistingConfiguration($this->configuration), new ExistingConnection($this->connection));
```
greg0ire: What does `di` mean here? Dependency injection? Shouldn't it be `$dependencyFactory`?

tests/Doctrine/Migrations/Tests/DependencyFactoryTest.php
```diff
@@ -156,6 +158,36 @@ public function testServiceDefinition() : void
         self::assertSame($anotherLogger, $di->getLogger());
     }
 
+    public function testServiceDefinitionWithReference() : void
```
greg0ire: Where is the reference in that test?

greg0ire (commented):
Sorry, I don't really understand what this PR does. I see that we avoid calling a factory for something that we already are "resolving", but I don't understand what "the default service definition" is, or where the circular dependency is in the test.

goetas:
@greg0ire thanks for the review. I think that I should improve the naming of the tests and some better description.

What is happening is the following.

You might want to provide a custom migration factory, but to avoid re implementing some of the things that doctrine already does well (providing the connection to the migration as example), what you wand to do is to **decorate** the current migration factory (otherwise you are forced to implement also the default migration factory rather than decorating it).


The code should look to something as follows:

```php

$dependencyFactory->setDefinition(MigrationFactory::class, function() use ($dependencyFactory) {
    $originalMigrationFactory = $dependencyFactory->getMigrationFactory();

    return new MyMigrationFactory($originalMigrationFactory);
 });
```

With the `3.0` version the provided code goes in a recursive loop. Mainly becase the ` $dependencyFactory->getMigrationFactory();` will try to invoke again the closure passed to `setDefinition`, going in recursion...

My change consists in avoiding to call the closure passed to `setDefinition` and to use the "default" implementation provided by the dependency factory.

In this way is possible to do proper service decoration.

stof:
An alternative API that could be simpler would be to make the callable building the custom dependency receive a `$getOriginal` callable as argument that would return the original one.

tests/Doctrine/Migrations/Tests/DependencyFactoryTest.php
```diff
@@ -156,6 +158,36 @@ public function testServiceDefinition() : void
...
+    {
+        $migration = $this->createMock(AbstractMigration::class);
+
+        $di = DependencyFactory::fromConnection(new ExistingConfiguration($this->configuration), new ExistingConnection($this->connection));
+
+        $di->setDefinition(MigrationFactory::class, static function (DependencyFactory $innerDi) use ($migration) {
+            // this call needs to stay here just to trigger again the dependency resolution
```
stof: why do we need this exactly ? This creates a confusing API IMO

goetas:
@stof 

> An alternative API that could be simpler would be to make the callable building the custom dependency receive a $getOriginal callable as argument that would return the original one.

how would that play with the symfony dependency injection container?

The proposed solution allows you to do something as:


```yaml
doctrine_migrations:
    services:
        Doctrine\Migrations\Version\MigrationFactory: 'doctrine.migrations.migrations_factory'
       
services:
  doctrine.migrations.migrations_factory:
      class: Doctrine\Migrations\Version\MigrationFactory
      factory: [@doctrine.migrations.dependency_factory'', 'getMigrationFactory']


  my_migrations_factory:
    class: Something
    decorates: doctrine.migrations.migrations_factory
   arguments:
      - '@my_migrations_factory.inner'

```

goetas:
IMO, in the long term the whole dependency factory class could be replaced by some third party dependency injection container.... (the hard choice is to choose one as it should be compatible with all the other DICs).

I'm not aware on the status of https://github.com/Roave/psr-container-doctrine (cc @asgrim), but could be an option.

tests/Doctrine/Migrations/Tests/DependencyFactoryTest.php
goetas: This is just because I was lazy to create some service that effectively decorates the one returned by `getMigrationFactory()`. Will create one and use the return value.

asgrim:
@goetas `roave/psr-container-doctrine` is active (time permitting) :+1:

goetas:
Thanks @asgrim for the fast feedback. I'm not familiar with the repo, but would it be able to "cooperate" with the symfony container and allows something similar to https://github.com/doctrine/migrations/pull/1054#issuecomment-694094516 (service decoration using services available in the symfony's container)?

asgrim:
@goetas if you can point a service at an "invokable" factory class in Symfony's container, then yep, it should work; `psr-container-doctrine` is really just a set of factory classes

stof:
@goetas AFAICT, your PR does not allow your snippet to work

goetas:
@stof https://github.com/doctrine/DoctrineMigrationsBundle/pull/368 is using exactly the same strategy to provide container aware migrations by default in the bundle... and it seemed working...

stof:
@goetas no. You are missing the extra code where it re-injects its own service in the dependency factory.

goetas:
> @goetas no. You are missing the extra code where it re-injects its own service in the dependency factory.
This part of code does that:

```yaml
doctrine_migrations:
    services:
        Doctrine\Migrations\Version\MigrationFactory: 'doctrine.migrations.migrations_factory'
``` 

And is handled in https://github.com/doctrine/DoctrineMigrationsBundle/blob/880527161a1c88c8ccde558105441a9016f49650/DependencyInjection/DoctrineMigrationsExtension.php#L72

massimilianobraglia:
Hey there, any news on this one?

goetas:
I have updated the tests for this pull request with a (hopefully) more self explanatory test case.

lib/Doctrine/Migrations/DependencyFactory.php
```diff
@@ -59,6 +59,9 @@
  */
 class DependencyFactory
 {
+    /** @var bool[] */
```
greg0ire:
```suggestion
    /** @psalm-var array<string, bool> */
```

tests/Doctrine/Migrations/Tests/Stub/CustomClassNameMigrationFactory.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Doctrine\Migrations\AbstractMigration;
+use Doctrine\Migrations\Version\MigrationFactory;
+
+class CustomClassNameMigrationFactory implements MigrationFactory
+{
+    /** @var MigrationFactory */
+    private $parentMigrationFactory;
+
+    /** @var string */
```
greg0ire:
```suggestion
    /** @psalm-var class-string<CanItBeAnyClassOrShouldItBeASpecificType> */
```

tests/Doctrine/Migrations/Tests/Stub/CustomClassNameMigrationFactory.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Doctrine\Migrations\AbstractMigration;
+use Doctrine\Migrations\Version\MigrationFactory;
+
+class CustomClassNameMigrationFactory implements MigrationFactory
+{
+    /** @var MigrationFactory */
+    private $parentMigrationFactory;
+
+    /** @psalm-var class-string<CanItBeAnyClassOrShouldItBeASpecificType> */
```
greg0ire: You will need to adapt my suggestion :laughing:

goetas: :exploding_head:

goetas: done

tests/Doctrine/Migrations/Tests/Stub/CustomClassNameMigrationFactory.php
```diff
@@ -0,0 +1,28 @@
...
+    /** @psalm-var class-string<AbstractMigration> */
+    private $migrationClassName;
+
+    public function __construct(MigrationFactory $parentMigrationFactory, string $migrationClassName)
+    {
+        $this->parentMigrationFactory = $parentMigrationFactory;
+        $this->migrationClassName     = $migrationClassName;
+    }
+
+    public function createVersion(string $migrationClassName): AbstractMigration
```
greg0ire, line 27:
```suggestion
    /**
     * @psalm-param class-string<AbstractMigration> $migrationClassName
     */
    public function createVersion(string $migrationClassName): AbstractMigration
```

goetas:
yay, green

greg0ire:
@stof can you please give this another review :pray: ?

greg0ire (approved):

greg0ire:
Stof seems busy so I will approve, it looks good to me but I really would have liked another review from him :grimacing:

goetas:
I'm merging this as it will unblock https://github.com/doctrine/DoctrineMigrationsBundle/pull/368 https://github.com/doctrine/DoctrineMigrationsBundle/pull/364 (and allow to close https://github.com/doctrine/DoctrineMigrationsBundle/issues/352)

PR #1060, opened 2020-10-02
Cast to integer to avoid Spanner GCP incompatibility

gabrielfs7:
#### Summary

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | Avoid delegate to DB responsibility to cast `execution_time` and fix GCP Spanner issue

I am currently receiving error from GCP when trying to persist data using Spanner:

```
An exception occurred while executing 'INSERT INTO doctrine_migration_versions (version, executed_at, execution_time) VALUES (?, ?, ?)' with params ["oat\\generis\\migrations\\Version202009301828472348_generis", "2020-10-02T11:54:13.671062Z", 130]:  
                                                                                                                                                                                                                                                            
  {                                                                                                                                                                                                                                                         
      "message": "Invalid value for bind parameter param3: Expected INT64.",                                                                                                                                                                                
      "code": 3,                                                                                                                                                                                                                                            
      "status": "INVALID_ARGUMENT",                                                                                                                                                                                                                         
      "details": [                                                                                                                                                                                                                                          
          {                                                                                                                                                                                                                                                 
              "@type": "grpc-server-stats-bin",                                                                                                                                                                                                             
              "data": "<Unknown Binary Data>"                                                                                                                                                                                                               
          }                                                                                                                                                                                                                                                 
      ]                                                                                                                                                                                                                                                     
  }        
```
Cause the 130 is actually `130.0` (a float).

Since migrations always expects an integer here I would like to cast this value to avoid driver incompatibilities

The issue was reported here: https://github.com/doctrine/migrations/issues/1061

### DIFF
--- a/tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
+++ b/tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
@@ -186,6 +186,67 @@ class TableMetadataStorageTest extends TestCase
         ], $rows);
     }
 
+    public function testCompleteWillAlwaysCastTimeToInteger() : void
+    {
+        $schemaManager = $this->createMock(AbstractSchemaManager::class);
+        $connection    = $this->createMock(Connection::class);
+
+        $expectedTable = new Table($this->config->getTableName());
+        $expectedTable->addColumn(
+            $this->config->getVersionColumnName(),
+            'string',
+            [
+                'notnull' => true,
+                'length' => $this->config->getVersionColumnLength(),
+            ]
+        );
+        $expectedTable->addColumn(
+            $this->config->getExecutedAtColumnName(),
+            'datetime',
+            ['notnull' => false]
+        );
+        $expectedTable->addColumn(
+            $this->config->getExecutionTimeColumnName(),
+            'integer',
+            ['notnull' => false]
+        );
+        $expectedTable->setPrimaryKey([$this->config->getVersionColumnName()]);
+
+        $connection->expects(self::once())
+            ->method('getSchemaManager')
+            ->willReturn($schemaManager);
+
+        $connection->expects(self::once())
+            ->method('insert')
+            ->willReturnCallback(
+                function (string $tableName, array $params) : void {
+                    self::assertSame($this->config->getTableName(), $tableName);
+                    self::assertSame('1230', $params['version']);
+                    self::assertSame(31490, $params['execution_time']);
+                    self::assertEquals(new DateTimeImmutable('2010-01-05 10:30:21'), $params['executed_at']);
+                }
+            );
+
+        $schemaManager->expects(self::once())
+            ->method('tablesExist')
+            ->willReturn(true);
+
+        $schemaManager->expects(self::once())
+            ->method('listTableDetails')
+            ->willReturn($expectedTable);
+
+        $result = new ExecutionResult(
+            new Version('1230'),
+            Direction::UP,
+            new DateTimeImmutable('2010-01-05 10:30:21')
+        );
+        $result->setTime(31.49);
+
+        $storage = new TableMetadataStorage($connection, new AlphabeticalComparator(), $this->config);
+
+        $storage->complete($result);
+    }
+
     public function testRead() : void
     {
         $this->storage->ensureInitialized();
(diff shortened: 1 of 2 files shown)


mike-ionut-mihai-sandu-tao (approved):
looks good from my side, 
and i would go for this because the intent is declared below on the ` Types::INTEGER`
at tline https://github.com/doctrine/migrations/pull/1060/files#diff-fd5b2ea887e9abe3b606201b76d10bb9R144

xvzf (commented):
Nice!

goetas:
To me looks good and reasonable. Is there a way to add a test for such change?

gabrielfs7:
> To me looks good and reasonable. Is there a way to add a test for such change?

@goetas Seems this test was already done before, but with SQLite driver it always passed :) 

https://github.com/doctrine/migrations/blob/3.0.x/tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php#L161

goetas:
@greg0ire do you have suggestions here? (merge it as it is?)

siwane (approved):
- [ ] New code is covered by tests (if applicable)
- [ ] Tests are running successfully (old and new ones) on my local machine (if applicable)
- [ ] New code is respecting code style rules
- [ ] New code is respecting best practices
- [ ] New code is not subject to concurrency issues (if applicable)
- [ ] Feature is working correctly on my local machine (if applicable)
- [ ] Acceptance criteria are respected
- [ ] Pull request title and description are meaningful

SenseException:
If an execution time appears that is bigger than `int` can handle, then there is a much bigger problem that is not related to doctrine/migrations. So I guess this change should be fine.

I recommend to add a unit test because this change has to be handled to prevent breaks in the future in case this would be a major problem for users.

BTW: Should this PR target `3.0.x`? Is this a bugfix?

stof:
this is indeed a bugfix

gabrielfs7:
> If an execution time appears that is bigger than `int` can handle, then there is a much bigger problem that is not related to doctrine/migrations. So I guess this change should be fine.
> 
> I recommend to add a unit test because this change has to be handled to prevent breaks in the future in case this would be a major problem for users.
> 
> BTW: Should this PR target `3.0.x`? Is this a bugfix?

I will check better way to test this and come back soon. Thanks all for the input.

gabrielfs7:
Tests added @goetas @SenseException @stof Can you please take a look and merge if is ok. Thank you in advance

tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
```diff
@@ -186,6 +187,67 @@ public function testCompleteWithFloatTime() : void
         ], $rows);
     }
 
+    public function testCompleteWillAlwaysCastTimeToInteger() : void
+    {
+        $schemaManager = $this->createMock(SqliteSchemaManager::class);
```
greg0ire, line 192: Why choose to mock this concrete class rather than `AbstractSchemaManager`?

gabrielfs7: Because I was debugging the test to create this mock and used the same used in the test, but just moved to the abstract one. I agree it is better. Thanks!

greg0ire:
Please associate your email address with your Github account, or change the
email in your commits to an address already associated with it. If you do not
want to expose your personal email address, you may use
`$userName@users.noreply.github.com`, that way we can still reach you through Github.
Additionally, you will get credit for this contribution on your Github profile.

Also, please kindly squash your commits together. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble.

How to do that?

1. `git rebase -i origin/master`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.

And finally, since this is a bugfix, I think you should target 3.0.x

greg0ire:
Let me help you with the commits :P

greg0ire (approved):

gabrielfs7:
> Let me help you with the commits :P

I appreciate, thanks! :)

goetas:
In my experience mocks have been always a problem. The less the better. I've created https://github.com/doctrine/migrations/pull/1063 that solves the same problem just with a different testing strategy.

Can you please review that version ? (@greg0ire @gabrielfs7 )

gabrielfs7:
> #1063

@goetas I have dropped a comment there: https://github.com/doctrine/migrations/pull/1063/files#r507585488

Unfortunately the test still passes if you remove the cast. The way I did the test breaks if the cast is removed.

goetas (approved):

SenseException (approved):

greg0ire:
Thanks @gabrielfs7 !

Issue #1061, opened 2020-10-02
Migrations execution_time should be integer in order to be compatible with GCP Spanner


gabrielfs7:
### Bug Report

I am currently receiving error from GCP when trying to persist data using Spanner:

```
An exception occurred while executing 'INSERT INTO doctrine_migration_versions (version, executed_at, execution_time) VALUES (?, ?, ?)' with params ["oat\\generis\\migrations\\Version202009301828472348_generis", "2020-10-02T11:54:13.671062Z", 130]:  
                                                                                                                                                                                                                                                            
  {                                                                                                                                                                                                                                                         
      "message": "Invalid value for bind parameter param3: Expected INT64.",                                                                                                                                                                                
      "code": 3,                                                                                                                                                                                                                                            
      "status": "INVALID_ARGUMENT",                                                                                                                                                                                                                         
      "details": [                                                                                                                                                                                                                                          
          {                                                                                                                                                                                                                                                 
              "@type": "grpc-server-stats-bin",                                                                                                                                                                                                             
              "data": "<Unknown Binary Data>"                                                                                                                                                                                                               
          }                                                                                                                                                                                                                                                 
      ]                                                                                                                                                                                                                                                     
  }        
```
Cause the 130 is actually `130.0` (a float).

Since migrations always expects an integer here I would like to cast this value to avoid driver incompatibilities

Just created a PR for this https://github.com/doctrine/migrations/pull/1060

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | x.y.z / master

#### Summary

Library delegates to DB to cast float to integer while inserting migrations. This causes issues when integrating with GCP Spanner.

#### Current behavior

The migrations record cannot be persisted.

#### How to reproduce

1. Have a migrations table on GCP Spanner
2. Run migrations

#### Expected behavior

The data should be persisted in the migrations table.

mike-ionut-mihai-sandu-tao:
it is happening to me also , i saw the fix and i think it is quite small and will be a great progress

gabrielfs7:
Is there any estimation how can we merge this bugfix? https://github.com/doctrine/migrations/pull/1060

PR #1063, opened 2020-10-19
Cast to int the execution time

goetas:
Cast to int the execution time (to avoid incompatibilities with RDBMS as Google Spanner)

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/migrations/issues/1061
| Alternative to  | https://github.com/doctrine/migrations/pull/1060

### DIFF
--- a/tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
+++ b/tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
@@ -7,12 +7,14 @@ namespace Doctrine\Migrations\Tests\Metadata\Storage;
 use DateTime;
 use DateTimeImmutable;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\PDOSqlite\Driver as SQLiteDriver;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Types\DateTimeType;
 use Doctrine\DBAL\Types\IntegerType;
 use Doctrine\DBAL\Types\StringType;
+use Doctrine\DBAL\Types\Types;
 use Doctrine\Migrations\Exception\MetadataStorageError;
 use Doctrine\Migrations\Metadata\Storage\TableMetadataStorage;
 use Doctrine\Migrations\Metadata\Storage\TableMetadataStorageConfiguration;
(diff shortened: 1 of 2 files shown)


tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
```diff
@@ -53,6 +54,37 @@ public function setUp() : void
...
+            ->onlyMethods(['insert'])
+            ->getMock();
+
+        $connection
+            ->expects(self::once())
+            ->method('insert')
+            ->with($config->getTableName(), [
+                $config->getVersionColumnName() => '1230',
+                $config->getExecutedAtColumnName() => $executedAt,
+                $config->getExecutionTimeColumnName() => 31000,
```
gabrielfs7, line 76: @goetas I understand your approach. I though about the same, but, if you remove the `(int)` cast, the test still passes. That is why I used `mock` and `assertSame`. Wdyt?

tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
```diff
@@ -7,6 +7,7 @@
 use DateTime;
 use DateTimeImmutable;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\Driver\PDOSqlite\Driver as SqlLiteDirver;
```
greg0ire, line 10:
```suggestion
use Doctrine\DBAL\Driver\PDOSqlite\Driver as SqlLiteDriver;
```

tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
greg0ire: If the issue only happens on Google Spanner, there is no way we are going to get a failing test, since we don't have Google Spanner tested in our CI, right?

gabrielfs7: In my PR the test fails if we remove the casting, cause I am using `assertSame`. But SQLite does internal cast, different from Spanner.

goetas:
I have changed the tests to use `assertSame` and `willReturnCallback` that should be much more safe

gabrielfs7 (approved):
Not sure you can point this to my branch, but I applied your changes in the test there if you do not mind. I agree the test is cleaner. Thanks

https://github.com/doctrine/migrations/pull/1060

goetas:
Done in https://github.com/doctrine/migrations/pull/1060

PR #1069, opened 2020-10-28
Upgrade cs lib

greg0ire:


greg0ire:
The Travis build fails because of an issue I reported here: https://github.com/box-project/box/issues/489

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,98 @@
...
+          path: "coverage.xml"
+
+  upload_coverage:
+    name: "Upload coverage to Codecov"
+    runs-on: "ubuntu-20.04"
+    needs:
+      - "phpunit"
+
+    steps:
+      - name: "Checkout"
```
smoench, line 85: Have you tried with the job without the checkout step? I think there is no need for and will saves build time.

.github/workflows/continuous-integration.yml
```diff
@@ -0,0 +1,98 @@
...
+        if: "${{ matrix.deps == 'normal' }}"
+
+      - name: "Install lowest possible dependencies with composer"
+        run: "composer update --no-interaction --prefer-dist --prefer-lowest"
+        if: "${{ matrix.deps == 'low' }}"
+
+      - name: "Run PHPUnit"
+        run: "vendor/bin/phpunit --coverage-clover=coverage.xml"
+
+      - name: "Upload coverage file"
```
smoench, line 72: I would recommend uploading the coverage file only once and enable the coverage run only on a specific php version. It will saves build time. Here an example how it could be done https://github.com/humbug/php-scoper/pull/422

.github/workflows/continuous-integration.yml
greg0ire: Oh you're right, that should work indeed.

.github/workflows/continuous-integration.yml
greg0ire: What if I have code that only execute on a specific version of PHP?

tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrationVersionTest.php
```diff
@@ -33,7 +36,11 @@ protected function setUp() : void
 
         $this->configuration = new Configuration($this->getSqliteConnection());
         $this->configuration->setMigrationsNamespace('DoctrineMigrations');
-        $this->configuration->setMigrationsDirectory(sys_get_temp_dir());
+        $migrationsDirectory = sys_get_temp_dir() . DIRECTORY_SEPARATOR . 'doctrine-migrations';
```
goetas, line 39: good catch!

.github/workflows/continuous-integration.yml
goetas: is Codecov able to distinguish them? (i remember that scrutinizer was considering only the first received file...)

tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrationVersionTest.php
greg0ire: Yup, that one gave a weird error that was hard to understand :smile:

.github/workflows/continuous-integration.yml
greg0ire: Yes, it can tell you how many jobs cover a single line

greg0ire:
Blocked by https://github.com/box-project/box/issues/490

.github/workflows/continuous-integration.yml
smoench: Nevermind, now unterstand what the coverage steps are doing. Looks fine!

greg0ire: It's a bit more complex that uploading directly to Codecov, I had to do it on the DBAL where those upload would fail because of a limit in the number of jobs, and ending up doing this: one upload of all artifacts in one step. It's a bit more complicated but I think it might make the build more stable by reducing the possibility of a failure when uploading to Codecov.

goetas (approved):
I am good with this changes. They improve the current state of the codebase.

download-box.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 
 if [ ! -f box.phar ]; then
-    wget https://github.com/humbug/box/releases/download/3.1.0/box.phar -O box.phar
+    wget https://github.com/box-project/box/releases/download/3.9.0/box.phar -O box.phar
```
goetas, line 4: this is giving 404

goetas (dismissed):
there is a 404 to download the box

download-box.sh
greg0ire: Yeah, I'm not sure why I committed that, anyway we probably won't use 3.9.0 either since it still suffers from the version detection bug. [No need to report the issue though, the team is aware of it](https://github.com/box-project/box/issues/490)

greg0ire:
Converting to draft until the numerous issues with box are resolved.

greg0ire:
This is too complex, I will move to GA in a separate PR.

greg0ire:
I just noticed that the PHAR compilation was only ever tested with low dependencies. With 96ce53a, it's not tested at all, and I don't understand if testing it only once was deliberate. @jwage , maybe you remember? I believe you changed this in #700

greg0ire:
And of course now it fails with the same errors I had when attempting to migrate to GA… that had nothing to do with GA.

greg0ire:
Good news: I'm reproducing the failure locally. Also, I reproduce the failure without PHPUnit if I run the compilation from the `tests` directory. I think the current working directory might be important.

greg0ire:
Ah apparently this is supposed to be fixed by 3.9.0: https://github.com/box-project/box/commit/e65178c5edaf5012b8b6ae9137274f76a4b0a572 … which cannot be downloaded as a PHAR : https://github.com/box-project/box/issues/490

goetas:
> Ah apparently this is supposed to be fixed by 3.9.0: box-project/box@e65178c … which cannot be downloaded as a PHAR : box-project/box#490
:cry:

greg0ire:
A version of the PHAR was just manually uploaded! It fixes the build locally, I just pushed :tada:

goetas (approved):
:+1:  looks good!

PR #1071, opened 2020-11-05
Revert "Merge pull request #888 from goetas/update-query"

lyrixx:
This reverts commit b13df49bd2170a6646469081a7e9713650e1bec5, reversing
changes made to b5d7611093e57216eced3d0e73911cf433d5039e.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | fixes https://github.com/doctrine/migrations/pull/888#issuecomment-722318249

### DIFF
--- a/lib/Doctrine/Migrations/Version/Executor.php
+++ b/lib/Doctrine/Migrations/Version/Executor.php
@@ -345,7 +345,7 @@ final class Executor implements ExecutorInterface
 
             $this->outputSqlQuery($key, $query);
 
-            // Do not use method that returns results here, to not block next queries
+            // executeQuery() must be used here because $query might return a result set, for instance REPAIR does
             if (! isset($this->params[$key])) {
                 $this->connection->executeQuery($query);
             } else {
--- a/lib/Doctrine/Migrations/Version/Executor.php
+++ b/lib/Doctrine/Migrations/Version/Executor.php
@@ -345,10 +345,11 @@ final class Executor implements ExecutorInterface
 
             $this->outputSqlQuery($key, $query);
 
+            // Do not use method that returns results here, to not block next queries
             if (! isset($this->params[$key])) {
-                $this->connection->executeUpdate($query);
+                $this->connection->executeQuery($query);
             } else {
-                $this->connection->executeUpdate($query, $this->params[$key], $this->types[$key]);
+                $this->connection->executeQuery($query, $this->params[$key], $this->types[$key]);
             }
 
             $stopwatchEvent->stop();
--- a/tests/Doctrine/Migrations/Tests/Version/ExecutorTest.php
+++ b/tests/Doctrine/Migrations/Tests/Version/ExecutorTest.php
@@ -108,26 +108,6 @@ class ExecutorTest extends TestCase
         self::assertFalse($this->migration->postDownExecuted);
     }
 
-    public function testExecuteUsedExecuteUpdate() : void
-    {
-        $this->connection
-            ->expects(self::never())
-            ->method('executeQuery');
-        $this->connection
-            ->expects(self::exactly(2))
-            ->method('executeUpdate');
-
-        $migratorConfiguration = (new MigratorConfiguration())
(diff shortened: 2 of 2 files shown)


greg0ire:
> reversing changes made to b5d7611.

I think you meant 595489b4c456a6926365e270cdaba8c401350506, and if that's the case, why not just `git revert 595489b4c456a6926365e270cdaba8c401350506`?

lyrixx:
@greg0ire this is commit message generated by git itself 😬. I understand the message is confusing, but it talks about the parent

greg0ire:
Oh right, I misread, didn't notice it was changes applied _to_ b5d7611 . Anyway, an automated message cannot be as good as the one you could write as suggested in https://github.com/doctrine/migrations/pull/888#issuecomment-722945777 , please edit the message to include the information to understand why that revert is necessary.

goetas:
I'm sorry that this created so many issue, but i'm ok for reverting it :+1:

lyrixx:
@greg0ire is it OK now ?

lib/Doctrine/Migrations/Version/Executor.php
```diff
@@ -345,10 +345,11 @@ private function executeVersionExecutionResult(
 
             $this->outputSqlQuery($key, $query);
 
+            // Do not use method that returns results here, to not block next queries
```
greg0ire, line 348: You can use methods that returns results, that's exactly what you did. For that reason, `executeQuery` should not be changed to `executeUpdate`
```suggestion
            // executeQuery() must be used here because $query might return a result set, for instance REPAIR does
```

greg0ire (commented):
I think you got it right with the commit message, but wrong with the code comment

greg0ire (approved):

greg0ire:
There is a new failure I didn't see before in Appveyor, but since this is a revert, it's fine to merge, let's proceed, hopefully we will be able to fix the build soon :grimacing:

greg0ire:
Thanks @lyrixx !

lyrixx:
Hello, 

May I ask for a release of [2.3.1](https://github.com/doctrine/migrations/milestone/36)?

Thanks

lib/Doctrine/Migrations/Version/Executor.php
```diff
@@ -345,10 +345,11 @@ private function executeVersionExecutionResult(
 
             $this->outputSqlQuery($key, $query);
 
+            // executeQuery() must be used here because $query might return a result set, for instance REPAIR does
```
stof, line 348: then, to be sure to be compatible with `PrimaryReadReplicaConnection` (which allows `executeQuery` to run on the read replicas), we should add a check to call `ensureConnectedToPrimary` when the connection is a `PrimaryReadReplicaConnection` (for transactional migrations, there is a good chance to be connected on the primary due to starting the transaction, but it is possible to disable all usages transactions in doctrine/migrations AFAIK)

greg0ire: That wouldn't qualify as a bugfix, so in 3.1.x, right? That being said, your remark makes me notice that we have `instanceof MasterSlaveConnection` that should be replaced with `instanceof PrimaryReadReplicaConnection`

greg0ire: Attempt at a fix: https://github.com/doctrine/migrations/pull/1096

greg0ire:
@lyrixx I hope to get #1096 merged soon and then let's do a release 👍

greg0ire:
@lyrixx released as https://github.com/doctrine/migrations/releases/tag/2.3.1

lyrixx:
hello @greg0ire ; Now I'm trying to update to 3.x, and I'm facing the very same issue :) 
Could you release 3.0.2 ? thx

greg0ire:
I was on holidays, but it looks like you have been heard: https://github.com/doctrine/migrations/releases/tag/3.0.2 :)

PR #1073, opened 2020-11-08
Migrate to Github Actions

greg0ire:


greg0ire:
Argh, will have to fix PHPStan first

PR #1075, opened 2020-11-20
Allow organize_migrations to be set to false which is the default val…

TomHAnderson:
…ue.  This allows organize_migrations to be reset to default in case a previous configuration value set it to year or year_and_month

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

Allow organize_migrations to be set to false which is the default value.  This allows organize_migrations to be reset to default in case a previous configuration value set it to year or year_and_month

### DIFF
--- a/lib/Doctrine/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/Migrations/Configuration/Configuration.php
@@ -201,9 +201,19 @@ final class Configuration
         return $this->checkDbPlatform;
     }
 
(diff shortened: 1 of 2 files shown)


TomHAnderson:
This PR is broken because composer2 is not supported yet

greg0ire:
We need to merge up 2.x for it to work, I attempted it but failed. My hope is that @goetas can do it.

PR #1076, opened 2020-11-20
Update configuration.rst to include the `organize_migrations` setting.

bfoosness:
|      Q       |   A
|------------- | -----------
| Type         | documentation

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -39,6 +39,7 @@ Now, in the root of your project place a file named ``migrations.php``, ``migrat
 
             'all_or_nothing' => true,
             'check_database_platform' => true,
+            'organize_migrations' => 'none',
         ];
 
     .. code-block:: yaml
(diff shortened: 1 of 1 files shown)


docs/en/reference/configuration.rst
```diff
@@ -39,6 +39,7 @@ Now, in the root of your project place a file named ``migrations.php``, ``migrat
 
             'all_or_nothing' => true,
             'check_database_platform' => true,
+            'organize_migrations' => 'year',
```
goetas: by putting here `year` it seems that is the default....  does it make sense to say that the default is null and to list the possible values?

bfoosness: That seems reasonable, but at least until https://github.com/doctrine/migrations/pull/1075 is merged, I think passing null will trigger an error?

For what it's worth, the table below that describes each setting does specify that the default value is false, and the description lists the two acceptable values.

bfoosness: I can confirm that a value of null will trigger an error in 3.0.1. Once #1075 is merged, `false` would be the proper default value.

bfoosness: `none` is now the default value per #1079.

greg0ire (approved):

goetas (approved):

PR #1078, opened 2020-11-23
Manual merge of 2.3.x into 3.0.x

greg0ire:


lib/Doctrine/Migrations/FileQueryWriter.php
```diff
@@ -42,8 +43,8 @@ public function write(
         string $direction,
         array $queriesByVersion,
         ?DateTimeInterface $now = null
-    ) : bool {
-        $now = $now ?? new DateTimeImmutable();
+    ): bool {
+        $now ??= new DateTimeImmutable();
```
goetas, line 47: is `??=` available for php 7.2?

phpcs.xml.dist
```diff
@@ -15,6 +15,13 @@
 
     <rule ref="Doctrine"/>
 
+    <!-- Disable the rules that will require PHP 7.4 -->
```
goetas, line 18: i was disabling this (and other rules) by invoking `vendor/bin/phpcs  --config-set php_version 7020'`.. is there an equivalent command to do so in the config file?

lib/Doctrine/Migrations/FileQueryWriter.php
stof: no, it is not. that's a PHP 7.4 feature

greg0ire: I should add this to the list of disabled rules

phpcs.xml.dist
greg0ire: I didn't know you could do that, that would be great! Will look into it!

greg0ire: I found the right way to express this, see the "changes" tab :)

greg0ire:
Roave's BC-check si complaining about changes in `symfony/console`, so code that is not ours, that's weird…

stof:
This might be caused by the fact that our own class extends `Symfony\Component\Console\Command\Command` (and so inherits all these methods)

greg0ire:
That makes sense, since I removed `composer.lock` from version control, that component was probably locked to 4 before I did so.

phpcs.xml.dist
greg0ire: @morozov pointed out to me that `7020` was a bit weird, so I just pushed a new commit with `70200`, which looks more like a valid value of `PHP_VERSION`

TomHAnderson (approved):
This looks fine.  Do we really need to check the PHP version even though we're limiting PHP in the composer.json file?  https://github.com/doctrine/migrations/pull/1078/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R14

I also disagree with having this line in composer.json: https://github.com/doctrine/migrations/pull/1078/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34R55 but that can be fixed in a different PR once this is merged.

greg0ire:
We agreed to discuss and fix the details afterwards because I have to go to work while others around the globe such as @TomHAnderson might want to do some open-source.

PR #1079, opened 2020-11-23
A whole slew of changes for 3.0.x!  See the summary.

TomHAnderson:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

* Updated box to 3.9.1
* Allow for PDO namespace changes after DBAL 2.11
* Reformatted composer.json and removed config platform of 7.2
* Fixed at() calls in tests
* Allow organize_migrations to be set to false with a new 'none' value
* Add composer-normalize

Please note the branch name is now out of sync with this PR.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -1,4 +1,3 @@
-
 name: "Continuous Integration"
 
 on:
@@ -21,11 +20,14 @@ jobs:
   composer_normalize:
     name: "Composer Normalize"
     runs-on: "ubuntu-latest"
-    needs:
-      - "composer"
     steps:
-      - name: "Composer Normalize"
-        run: "composer normalize --dry-run"
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+
+      - uses: "docker://ergebnis/composer-normalize-action:latest"
+        with:
+          args: "--dry-run"
 
   phpunit:
     name: "PHPUnit"
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -19,7 +19,7 @@ jobs:
 
   composer_normalize:
     name: "Composer Normalize"
-    runs-on: "ubuntu-latest"
+    runs-on: "ubuntu-20.04"
     steps:
       - uses: actions/checkout@v2
         with:
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -18,6 +18,15 @@ jobs:
         with:
           args: --from=${{ github.event.pull_request.base.sha }}
 
+  composer_normalize:
+    name: "Composer Normalize"
+    runs-on: "ubuntu-latest"
+    needs:
+      - "composer"
+    steps:
+      - name: "Composer Normalize"
+        run: "composer normalize --dry-run"
+
   phpunit:
     name: "PHPUnit"
     runs-on: "ubuntu-20.04"
--- a/composer.json
+++ b/composer.json
@@ -1,81 +1,81 @@
 {
-  "name": "doctrine/migrations",
-  "type": "library",
-  "description": "PHP Doctrine Migrations project offer additional functionality on top of the database abstraction layer (DBAL) for versioning your database schema and easily deploying changes to it. It is a very easy to use and a powerful tool.",
-  "keywords": [
-    "database",
-    "migrations",
-    "dbal"
-  ],
-  "homepage": "https://www.doctrine-project.org/projects/migrations.html",
-  "license": "MIT",
-  "authors": [
-    {
-      "name": "Benjamin Eberlei",
-      "email": "kontakt@beberlei.de"
+    "name": "doctrine/migrations",
+    "type": "library",
+    "description": "PHP Doctrine Migrations project offer additional functionality on top of the database abstraction layer (DBAL) for versioning your database schema and easily deploying changes to it. It is a very easy to use and a powerful tool.",
+    "keywords": [
+        "database",
+        "migrations",
+        "dbal"
+    ],
+    "homepage": "https://www.doctrine-project.org/projects/migrations.html",
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "Benjamin Eberlei",
+            "email": "kontakt@beberlei.de"
+        },
+        {
+            "name": "Jonathan Wage",
+            "email": "jonwage@gmail.com"
+        },
+        {
+            "name": "Michael Simonson",
+            "email": "contact@mikesimonson.com"
+        }
+    ],
+    "require": {
+        "php": "^7.2",
+        "composer/package-versions-deprecated": "^1.8",
+        "doctrine/dbal": "^2.10",
+        "doctrine/event-manager": "^1.0",
+        "ocramius/proxy-manager": "^2.0.2",
+        "psr/log": "^1.1.3",
+        "symfony/console": "^3.4 || ^4.4.16 || ^5.0",
+        "symfony/stopwatch": "^3.4 || ^4.0 || ^5.0"
     },
-    {
-      "name": "Jonathan Wage",
-      "email": "jonwage@gmail.com"
+    "require-dev": {
+        "ext-pdo_sqlite": "*",
+        "doctrine/coding-standard": "^8.0",
+        "doctrine/orm": "^2.6",
+        "doctrine/persistence": "^1.3 || ^2.0",
+        "doctrine/sql-formatter": "^1.0",
+        "ergebnis/composer-normalize": "^2.11",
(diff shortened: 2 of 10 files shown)


stof:
is it expected to change the indentation of composer.json ?

TomHAnderson:
I ran composer.json through the npm package `json` because it was formatted with multiple properties per line.  That was intentional

lib/Doctrine/Migrations/Configuration/Configuration.php
```diff
@@ -174,9 +175,23 @@ public function isDatabasePlatformChecked(): bool
         return $this->checkDbPlatform;
     }
 
-    public function setMigrationOrganization(string $migrationOrganization): void
+    /**
+     * @param string | bool $migrationOrganization
```
goetas: instead of allowing both `string` and `bool`, what about having a sting constant  `none` that would reset the migration organization?

tests/Doctrine/Migrations/Tests/Generator/ClassNameGeneratorTest.php
```diff
@@ -7,12 +7,19 @@
...
 class ClassNameGeneratorTest extends TestCase
 {
     public function testName(): void
     {
         $generator = new ClassNameGenerator();
         $fqcn      = $generator->generateClassName('Foo');
-        self::assertRegExp('/^Foo\\\\Version[0-9]{14}$/', $fqcn);
+
+        if (method_exists($this, 'assertMatchesRegularExpression')) {
+            $this->assertMatchesRegularExpression('/^Foo\\\\Version[0-9]{14}$/', $fqcn);
```
goetas, line 20: I guess we should use the static implementation, so `self`

TomHAnderson: That doesn't play well with phpstan

lib/Doctrine/Migrations/Configuration/Configuration.php
TomHAnderson: I would be OK with that.

tests/Doctrine/Migrations/Tests/Tools/Console/Command/StatusCommandTest.php
```diff
@@ -73,6 +74,12 @@ public function testExecute(): void
 
         $lines = array_map('trim', explode("\n", trim($this->commandTester->getDisplay(true))));
 
+        if (DBALVersion::compare('2.11.0') > 0) {
+            $databaseDriver = 'Doctrine\DBAL\Driver\PDOSqlite\Driver '; // Trailing space bufferes outout assertion
```
greg0ire: > Trailing space buffers outout assertion

:thinking: this sentence seems broken and I don't know how to fix it :sweat_smile:

composer.json
```diff
@@ -1,66 +1,76 @@
...
-    "name": "doctrine/migrations",
-    "type": "library",
-    "description": "PHP Doctrine Migrations project offer additional functionality on top of the database abstraction layer (DBAL) for versioning your database schema and easily deploying changes to it. It is a very easy to use and a powerful tool.",
-    "keywords": ["php", "database", "migrations", "dbal"],
-    "homepage": "https://www.doctrine-project.org/projects/migrations.html",
-    "license": "MIT",
-    "authors": [
-        {"name": "Benjamin Eberlei", "email": "kontakt@beberlei.de"},
-        {"name": "Jonathan Wage", "email": "jonwage@gmail.com"},
-        {"name": "Michael Simonson", "email": "contact@mikesimonson.com" }
```
greg0ire, line 11: What's wrong with having multiple properties per line here? it still fits under a reasonable name. Besides, others might not have that `npm` package you used. Maybe we should adopt https://github.com/ergebnis/composer-normalize and put it in the CI? That would take care of that kind of argument over indent size, etc.

tests/Doctrine/Migrations/Tests/Generator/ClassNameGeneratorTest.php
TomHAnderson: More info:  `method_exists` cannot take `self` as the first parameter.  So if you use `$this` for `method_exists` then use `self` for `assertMatchesRegularExpression` `phpstan` has a problem with that.

TomHAnderson:
This repo needs to be set to a minimum PHP version of 7.3 in a future version so we can remove the check for the DBAL version which is deprecated and remove the logic for using using a regex assertion.

.github/workflows/continuous-integration.yml
```diff
@@ -18,6 +17,13 @@ jobs:
         with:
           args: --from=${{ github.event.pull_request.base.sha }}
 
+  composer_normalize:
+    name: "Composer Normalize"
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: "Composer Normalize"
+        run: "composer normalize --dry-run"
```
greg0ire: Here is some help regarding this: https://github.com/marketplace/actions/composer-normalize-action

TomHAnderson: Good help; thanks

tests/Doctrine/Migrations/Tests/Generator/ClassNameGeneratorTest.php
greg0ire: What about using the class name (can't use `$this::class` for now since it's PHP 8 but `TestCase::class` should do the trick)?

.github/workflows/continuous-integration.yml
```diff
@@ -18,6 +17,18 @@ jobs:
         with:
           args: --from=${{ github.event.pull_request.base.sha }}
 
+  composer_normalize:
+    name: "Composer Normalize"
+    runs-on: "ubuntu-latest"
```
greg0ire: We usually pin those:

```suggestion
    runs-on: "ubuntu-20.04"
```

composer.json
```diff
@@ -1,66 +1,81 @@
...
-    "require": {
-        "php": "^7.2",
-        "doctrine/dbal": "^2.10",
-        "doctrine/event-manager": "^1.0",
-        "composer/package-versions-deprecated": "^1.8",
-        "ocramius/proxy-manager": "^2.0.2",
-        "psr/log": "^1.1.3",
-        "symfony/console": "^3.4||^4.4.16||^5.0",
-        "symfony/stopwatch": "^3.4||^4.0||^5.0"
+  "name": "doctrine/migrations",
```
greg0ire: Let's revert to 4 spaces to make the diff minimal here? `composer normalize` is not supposed to complain as far as I understood, because it detects it:

> format the normalized content (either as sniffed, or as specified using the --indent-size and --indent-style options)

tests/Doctrine/Migrations/Tests/Generator/ClassNameGeneratorTest.php
TomHAnderson: Nope.  You see, phpstan is **clearing** (for lack of a better term) the use of $this after the method_exists function and seeing that it won't be called passes.  

```
        if (method_exists(TestCase::class, 'assertMatchesRegularExpression')) {
            self::assertMatchesRegularExpression('/^Foo\\\\Version[0-9]{14}$/', $fqcn);
```

Results in

```
 ------ ----------------------------------------------------------------------------------------------------------------------------------
  Line   tests/Doctrine/Migrations/Tests/Generator/ClassNameGeneratorTest.php
 ------ ----------------------------------------------------------------------------------------------------------------------------------
  20     Call to an undefined static method Doctrine\Migrations\Tests\Generator\ClassNameGeneratorTest::assertMatchesRegularExpression().
 ------ ----------------------------------------------------------------------------------------------------------------------------------
```

composer.json
TomHAnderson: 4 spaces was the reason I reformatted the .json file in the first place.  Note that I added composer-normalize configs into composer.json 

```json
    "composer-normalize": {
      "indent-size": 2,
      "indent-style": "space"
    }
```

greg0ire: But why? I think we use 4 spaces throughout Doctrine packages, and that's also
[what Composer documents](https://getcomposer.org/doc/01-basic-usage.md), and
what I've seen in the ecosystem, but maybe you have a different experience? 
I think the usage of composer-normalize could be generalized, and that we could
do that bikeshedding in a more centralized way to ensure consistency. I'd
recommend you stick with 4 spaces here, and then open a PR against
https://github.com/doctrine/.github with 2 spaces (and arguments for that),
that way we can discuss this once and do that change in a consistent way rather
than just on this package.

TomHAnderson: compser.json is now 4 spaces

greg0ire (approved):

goetas (approved):
looks good! thanks a lot @TomHAnderson

PR #1084, opened 2020-11-28
Use composer v1 when running the low deps CI jobs

goetas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

In https://github.com/doctrine/migrations/pull/1079, some dev deps were compatible only with composer v1, with this we run the builds using both composer versions

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -69,8 +69,12 @@ jobs:
         uses: "actions/cache@v2"
         with:
           path: "~/.composer/cache"
-          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          key: "php-${{ matrix.php-version }}-composer-${{ matrix.deps }}-locked-${{ hashFiles('composer.lock') }}"
           restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
+
+      - name: "Downgrade Composer"
+        run: "composer self-update --1"
+        if: "${{ matrix.deps == 'low' }}"
 
       - name: "Install dependencies with composer"
         run: "composer update --no-interaction --prefer-dist --no-progress"
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,7 @@
         "doctrine/orm": "^2.6",
         "doctrine/persistence": "^1.3 || ^2.0",
         "doctrine/sql-formatter": "^1.0",
-        "ergebnis/composer-normalize": "^2.11",
+        "ergebnis/composer-normalize": "^2.9",
         "phpstan/phpstan": "^0.12",
         "phpstan/phpstan-deprecation-rules": "^0.12",
         "phpstan/phpstan-phpunit": "^0.12",


greg0ire:
:thinking: maybe we should do that only for the `--prefer-lowest` job?

.github/workflows/continuous-integration.yml
```diff
@@ -69,9 +70,13 @@ jobs:
         uses: "actions/cache@v2"
         with:
           path: "~/.composer/cache"
-          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          key: "php-${{ matrix.php-version }}-composer-${{ matrix.composer }}-locked-${{ hashFiles('composer.lock') }}"
           restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
 
+      - name: "Get the right composer"
+        run: "composer self-update --${{ matrix.composer }}"
+        if: "${{ matrix.composer }}"
```
greg0ire:
```suggestion
        if: "${{ matrix.deps == 'low' }}"
```

.github/workflows/continuous-integration.yml
```diff
@@ -69,9 +70,13 @@ jobs:
         uses: "actions/cache@v2"
         with:
           path: "~/.composer/cache"
-          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          key: "php-${{ matrix.php-version }}-composer-${{ matrix.composer }}-locked-${{ hashFiles('composer.lock') }}"
```
greg0ire:
```suggestion
          key: "php-${{ matrix.php-version }}-composer-${{ matrix.deps }}-locked-${{ hashFiles('composer.lock') }}"
```

.github/workflows/continuous-integration.yml
```diff
@@ -38,6 +38,7 @@ jobs:
 
     strategy:
       matrix:
+        composer: [1, 2]
```
greg0ire:
```suggestion
```

.github/workflows/continuous-integration.yml
```diff
@@ -69,9 +70,13 @@ jobs:
         uses: "actions/cache@v2"
         with:
           path: "~/.composer/cache"
-          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          key: "php-${{ matrix.php-version }}-composer-${{ matrix.composer }}-locked-${{ hashFiles('composer.lock') }}"
           restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
 
+      - name: "Get the right composer"
+        run: "composer self-update --${{ matrix.composer }}"
```
greg0ire:
```suggestion
        run: "composer self-update --1"
```

.github/workflows/continuous-integration.yml
```diff
@@ -69,9 +70,13 @@ jobs:
         uses: "actions/cache@v2"
         with:
           path: "~/.composer/cache"
-          key: "php-${{ matrix.php-version }}-composer-locked-${{ hashFiles('composer.lock') }}"
+          key: "php-${{ matrix.php-version }}-composer-${{ matrix.composer }}-locked-${{ hashFiles('composer.lock') }}"
           restore-keys: "php-${{ matrix.php-version }}-composer-locked-"
 
+      - name: "Get the right composer"
```
greg0ire:
```suggestion
      - name: "Downgrade Composer"
```

goetas:
rebased and reworded the commit

greg0ire (approved):

greg0ire:
Thanks! Not labelling as "bug" since it's a dev dependency, the only people impacted are contributors.  Labelling as CI because it's a good thing to check for I suppose.

PR #1094, opened 2020-11-30
Always update migrated versions

kick-the-bucket:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

When migrating from v2 to v3 in a big project (with different migration classes in different branches) metadata sync becomes very cumbersome when you already have executed migrations from other branches than you are currently on, since the version names are only migrates once along with the table structure. At the moment the workaround we use is dropping the `execution_time` column before running `doctrine:migrations:migrate`, to ensure the names are always updated when you're in a branch with already executed, but not yet synced version names.

This change will make the names always update, no matter if there are structural changes to the table needed or not.

If this is not preferred - I could add a new flag to enable this behavior only when needed.

### DIFF
--- a/lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
+++ b/lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
@@ -158,11 +158,10 @@ final class TableMetadataStorage implements MetadataStorage
 
         $expectedSchemaChangelog = $this->getExpectedTable();
         $diff                    = $this->needsUpdate($expectedSchemaChangelog);
-        if ($diff === null) {
-            return;
+        if ($diff !== null) {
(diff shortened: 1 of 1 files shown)


goetas:
Hmm... I find this very risky (even if probably it will help local development...). 

I do not understand in which situations version names are not in sync? from my understanding, this happens only for a short period while migrating from 2.x to 3.x, isnt?

kick-the-bucket:
Situation:
1) you write a migration version 1 in a branch A and execute it
2) you update from migrations v2 to v3 in branch B.
3) you merge branch B into main/master, execute migrations, your metadata table structure is updated on first run, but since the already executed version 1 is not in your code at this moment (it's in a different branch) - the name does not get updated and it never will be, unless you do it manually.

PR #1095, opened 2020-12-02
Merge 3.0.x into 3.1.x

goetas:

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -39,6 +39,7 @@ Now, in the root of your project place a file named ``migrations.php``, ``migrat
 
             'all_or_nothing' => true,
             'check_database_platform' => true,
+            'organize_migrations' => 'none',
         ];
 
     .. code-block:: yaml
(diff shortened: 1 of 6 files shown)


greg0ire (approved):

docs/en/reference/configuration.rst
```diff
@@ -136,6 +140,8 @@ Here are details about what each configuration option does:
 +----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
 | check_database_platform    | no         | true                         | Whether to add a database platform check at the beginning of the generated code. |
 +----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
+| organize_migrations        | no         | ``none``                     | Whether to organize migration classes under year (``year``) or year and month (``year_and_month``) subdirectories. |
```
TomHAnderson, line 143: The text is reflective of the original no-'none' option.  I recommend changing the description to

```
Whether to not organize migration classes (``none``), organize them by year (``year``), or organize them by year and month (``year_and_month``) using subdirectories.
```

TomHAnderson (changes requested):

docs/en/reference/configuration.rst
goetas: I agree, but that should be done in the 3.0.x branch. The aim of this PR it just to port into 3.1.x whatever is there, is not to introduce any change.

TomHAnderson (approved):

PR #1096, opened 2020-12-02
Address MasterSlaveConnection rename

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | n/a

#### Summary

MasterSlaveConnection has been deprecated in favor of
PrimaryReadReplicaConnection, which it extends before being removed in
DBAL 3. This ensure both cases are handled when it comes to making sure
we connect to the primary, while not bumping the requirement to
doctrine/dbal 2.11 (which requires PHP 7.3, although subsequent version
might restore compatibility with 7.1).

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -17,6 +17,8 @@ parameters:
         - '~^Parameter #1 \$files of method Doctrine\\Migrations\\Finder\\Finder::loadMigrationClasses\(\) expects array<string>, array<int, string\|false> given\.\z~'
         - '~^Class Doctrine\\Migrations\\Tests\\DoesNotExistAtAll not found\.\z~'
         - '~^Call to method getVersion\(\) of deprecated class PackageVersions\\Versions\.$~'
+        # Drop when bumping to doctrine/dbal 2.11+
+        - '~^.*PrimaryReadReplicaConnection.*$~'
 
 includes:
     - vendor/phpstan/phpstan-deprecation-rules/rules.neon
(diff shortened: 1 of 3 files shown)


tests/Doctrine/Migrations/Tests/Configuration/ConfigurationTest.php
```diff
@@ -176,9 +178,40 @@ public function testMasterSlaveConnectionAlwaysConnectsToMaster() : void
...
+
+        self::assertTrue($configuration->connect());
+    }
+
+    /**
+     * Connection is tested via the `getMigratedVersions` method which is the
+     * simplest to set up for.
+     *
+     * @see https://github.com/doctrine/migrations/issues/336
+     */
```
greg0ire, line 204: I copy/pasted the comment of the method above.

phpstan.neon.dist
```diff
@@ -17,6 +17,8 @@ parameters:
         - '~^Parameter #1 \$files of method Doctrine\\Migrations\\Finder\\Finder::loadMigrationClasses\(\) expects array<string>, array<int, string\|false> given\.\z~'
         - '~^Class Doctrine\\Migrations\\Tests\\DoesNotExistAtAll not found\.\z~'
         - '~^Call to method getVersion\(\) of deprecated class PackageVersions\\Versions\.$~'
+        # Drop when bumping to doctrine/dbal 2.11+
+        - '~^.*PrimaryReadReplicaConnection.*$~'
```
greg0ire, line 21: phpstan is ran on 7.1, which prevents installing a recent enough version of `doctrine/dbal`, and generates quite a few errors because the class does not exist. The errors can be seen (for now) at https://github.com/doctrine/migrations/runs/1489165823?check_suite_focus=true

goetas (approved):

SenseException (approved):

PR #1099, opened 2020-12-17
Fixed metadata comparison on MariaDB 10.3.25

brusch:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

#### Summary

Currently if you run `./bin/console doctrine:migrations:sync-metadata-storage`  on MariaDB 10.3.25 it will fail, because `null !== 'NULL'`. 

How to reproduce: 
1. ensure the metadata table doesn't exist
2. run `./bin/console doctrine:migrations:sync-metadata-storage` -> table will be created
3. run the command again, you'll get the following error: `The metadata storage is not up to date, please run the sync-metadata-storage command to fix this issue.`


brusch:
Ok, it seems that we need to change the way how the comparison is made for default values ... 😐

PR #980, opened 2020-05-26
Allow PHP 8 in 2.x

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | n/a

#### Summary

We want to let downstream project experiment while we fix our issues if we have some. Failures are allowed for now, and should no longer be once feature freeze starts.

PR #1102, opened 2020-12-23
Allow PHP 8

nicolas-grekas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -

This PR includes and replaces https://github.com/doctrine/migrations/pull/980
Because so many deps forbid PHP8 unless you bump them also, this PR also bumps them.

Since bumping phpstan also adds new rules that the codebase is not compliant with, I generated a baseline to still be green on the static analysis job. This should be fixed in a later PR.

`Roave BC Check` also reports false positives BC breaks related to eg symfony/console bumping from v4 to v5. I commented the job: fixing this is also for another PR.

This is a canonical example of why dependencies should have broad version constraints: this would have reduced the size of the PR, thus would have reduced the risk. Instead, narrow constraints force doing extra unrelated work that puts the stability at risk. Let's decorrelate these other tasks in separate PRs, as just advised.


alcaeus (approved):
Changes LGTM, but I want to take some time before merging to make sure this integrates neatly into 3.0 when I have to merge this there.

PR #1105, opened 2020-12-31
Revert "Add phpstan baseline"

greg0ire:
This reverts commit e87a47ddb23abeeb0ac92a6c09b5d20ec0aea846.

#### Summary

I'm creating a draft PR because I want to ensure I get the same results locally or in CI while I'm working on this.


greg0ire:
I reported the build issue at https://github.com/shivammathur/setup-php/issues/387

SenseException:
Do you need to update the latest fix and change from draft to pull request?

greg0ire:
No, in this case I have PHPStan working on PHP 7.4 locally, but I think it would not be working if I was using 7.1 too. We have several options, I don't know which to pick:
1. Ignore the errors that happen on 7.1, which will mean breaking my local run of PHPStan.
2. Use 7.4 in the CI pipeline
3. Find a way to run PHPStan on both 7.4 and 7.1 (so have one config file per version of PHP).

Any opinions? I already discussed this some time ago with Ondrej, he recommended 3.

SenseException:
Does running it on 7.4 suggest incompatible changes to 7.1? What differences are there already between a run in 7.1 and 7.4 (except the found errors in the current build)?

greg0ire:
> Does running it on 7.4 suggest incompatible changes to 7.1?

It does not suggest anything more than what I put in the PR. So the difference is the errors in the current build. I will use solution 2 since it's what you seem to be hinting at.

greg0ire:
Ah I completely forgot about the deprecation detector… maybe we should remove that package, it makes sense to have it in a project, but not much on a library.

Also the BC-check detected some issues in what I did :grimacing: I think its concerns can be overruled, tell me if you think otherwise.

SenseException:
> I will use solution 2 since it's what you seem to be hinting at.

I don't have enough information to decide which one fits the best. We use solution 2 in other projects, but that doesn't mean it should be chosen. What are the advantages of 3 compared to the others?

> maybe we should remove that package, it makes sense to have it in a project, but not much on a library.

The deprecation check can help with the third party dependencies, but shouldn't it be a dev-dependency? It was introduced with commit 7e84e61a0b41aae8251f7780ba5b783c27f65921 and the message "Allow using on PHP 7.1 with Composer 2". Maybe Nicolas can tell more about the reason of this package.

> Also the BC-check detected some issues in what I did :grimacing: I think its concerns can be overruled, tell me if you think otherwise.

I agree. It shouldn't affect projects in any harmful way.

greg0ire:
You are mistaking `phpstan-deprecations-rules` with `package-versions-deprecated`, which is just a replacement for `package-versions`. It was introduced in bd44c7f41c806866a8926c5f157007c529316b9a, as part of https://github.com/doctrine/migrations/pull/739

>  What are the advantages of 3 compared to the others?

Better coverage of static analysis: some paths that cannot be analysed on one run will be analysed on the  other run, for instance here: 

https://github.com/doctrine/migrations/blob/dbbd7b33b8329ec1914eeb45741fd940ea7dec3c/lib/Doctrine/Migrations/Configuration/Configuration.php#L362-L369

greg0ire:
I will try to fix as many deprecations as possible before we take a decision.

ondrejmirtes:
@greg0ire Hi, can you phrase your question what are you asking me? :)

greg0ire:
Sorry @ondrejmirtes , here is the question (from @SenseException actually):

> What are the advantages of 3 compared to the others?

I feel that I might have not answered it fully.

ondrejmirtes:
2. Use 7.4 in the CI pipeline
3. Find a way to run PHPStan on both 7.4 and 7.1 (so have one config file per version of PHP).

You can do only 2. if you trust yourself you won't break anything on lower PHP versions. There isn't much difference between them anyway.

You can do 3. but you need multiple config files to ignore stuff that's only reported on that PHP version. You can reuse the common config parts with `includes` section in .neon files.

I'd also throw PHP 8.0 into the mix since it changes more stuff and PHPStan can be very useful there.

greg0ire:
> I'd also throw PHP 8.0 into the mix since it changes more stuff and PHPStan can be very useful there.

Might be worth it to try adding it indeed see if it reports more things on a package per package basis 👍

SenseException:
Number 3 sounds reasonable for me now. Thanks for clarification.

Issue #1106, opened 2020-12-31
PostgreSQL JSONB contains operator is broken


alvaro-octal:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 3.0.2

#### Summary

The migrations worked fine with 3.0.1, but version 3.0.2 seems to be trying to replace the `?` **PostgreSQL** operand by `$1` of queries like this:

```sql
UPDATE "user" u
SET permissions = permissions::jsonb || '["dashboard"]'::jsonb
WHERE (u.permissions_old->'dashboard'->'api')::jsonb ? 'read'
```

https://www.postgresql.org/docs/9.4/functions-json.html#FUNCTIONS-JSONB-OP-TABLE

#### Previous behavior

This migration was valid

#### Current behavior

```
[error] Migration App\Migrations\Version20201023161652 failed during Execution. Error: "An exception occurred while executing '
            UPDATE "user" u
            SET permissions = permissions::jsonb || '["dashboard"]'::jsonb
            WHERE (u.permissions_old->'dashboard'->'api')::jsonb ? 'read'
        ':
SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "$1"
LINE 4: ...ERE (u.permissions_old->'dashboard'->'api')::jsonb $1 'read'
                                                              ^"
In AbstractPostgreSQLDriver.php line 73:
                                                                              
  An exception occurred while executing '                                     
              UPDATE "user" u                                                 
              SET permissions = permissions::jsonb || '["dashboard"]'::jsonb  
              WHERE (u.permissions_old->'dashboard'->'api')::jsonb ? 'read'   
          ':                                                                  
                                                                              
  SQLSTATE[42601]: Syntax error: 7 ERROR:  syntax error at or near "$1"       
  LINE 4: ...ERE (u.permissions_old->'dashboard'->'api')::jsonb $1 'read'     
```

#### How to reproduce

Make any migration with the `?` operand on a PostgreSQL database on (at least) PHP 7.3

goetas:
This is actually a PHP bug, solved in 7.4 by https://wiki.php.net/rfc/pdo_escape_placeholders (see also https://bugs.php.net/bug.php?id=71885)

shadowhand:
I do not believe this is purely a PHP bug. The same query works just fine in doctrine/migrations v3.0.1 and breaks in v3.0.2.

adrianrudnik:
[...]

PR #1111, opened 2021-01-06
fix for jsonb contains operator

adrianrudnik:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1106 

#### Summary

Looking at the current code, the `executeQuery` usage mentions some case with a REPAIR returning a result set. It's unclear to my why this is important when the result is never returned, requested or in any other way important to a migration, except its success.

Usage of a static statement, instead of a prepared one, allows for simple JSONB contain queries again, like in 3.0.1.

As for testing, I'm unsure how to do it and I could not find any good approach to simulate a DBAL `executeQuery` vs `executeStatement` that would lead to this error to make it testable.

### DIFF
--- a/lib/Doctrine/Migrations/Version/DbalExecutor.php
+++ b/lib/Doctrine/Migrations/Version/DbalExecutor.php
@@ -295,8 +295,7 @@ final class DbalExecutor implements Executor
             $this->outputSqlQuery($query);
 
             $stopwatchEvent = $this->stopwatch->start('query');
-            // executeQuery() must be used here because $query might return a result set, for instance REPAIR does
-            $this->connection->executeQuery($query->getStatement(), $query->getParameters(), $query->getTypes());
+            $this->connection->executeStatement($query->getStatement(), $query->getParameters(), $query->getTypes());
             $stopwatchEvent->stop();
 
             if (! $configuration->getTimeAllQueries()) {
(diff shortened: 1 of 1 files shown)


andrew-demb:
See comment, why the same change was reverted
https://github.com/doctrine/migrations/pull/888#issuecomment-722318249

adrianrudnik:
Do you consider `REPAIR TABLE` and `OPTIMIZE TABLE` worthy of migrations? They do nothing to the schema and representation of the data itself. If so, I have no clue how to migrate JSONs right now.

goetas:
Well, this here is tricky, we are already going back and forth 3 times changing this line. Ideally it should be `executeUpdate`, but some vendors (mysql) have instructions that prefer the other way around.

(this because with `executeQuery` will never use `PDO::exec`, that does not try to parse `?`)

IMO we should leave this as it is, and use https://wiki.php.net/rfc/pdo_escape_placeholders to "workaround" the issue.

adrianrudnik:
Thanks @goetas , nice to know about the RFC for `??`. It works for now but will break BC for PHP < 7.4 if I read the RFC correctly. Also the dev creating the migrations needs to be aware of it as you cant just build your SQL inside a query window and simply paste it around without further editing, and some IDE support goes bad because of the bad syntax, but at least it migrates.

A warning would be nice in that case but seems hard to trigger as Doctrine and then deep down PDO itself will interpret the query, not the migrations. Maybe handle the parts of the error message `ERROR:  syntax error at or near "$1"` of `Doctrine\DBAL\Exception\SyntaxErrorException` and hint towards this situation / suggestions.

Also just rediscovered that you also could just `$this->connection->executeStatement` inside a migration if everything else fails.

Thanks for your work and time, I'll close this.

PR #1120, opened 2021-01-29
Fix of PHPStan issues

SenseException:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | -


goetas (approved):
it is interesting why phpstan was not complaining that the exclusion rule was not used...

PR #1125, opened 2021-02-07
use @internal instead of the type "@var internal"

goetas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

#### Summary

the issue was introduced in https://github.com/doctrine/migrations/pull/925 but  no one noticed, recent versions of phpstan detected this issue


goetas:
I think that we should ignore the Roave BC Check failure, that class was supposed to be internal, there was just a typo

greg0ire (approved):

greg0ire:
Thanks @goetas !

PR #1127, opened 2021-02-15
POC: config version comparator

fezfez:
Allow config version comparator + adding version comparator without namespace

|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no

#### Summary

Hello,

This is a POC, because I'm not sure that the implementation is correct.

the default versionComparator is duplicate twice (in Configuration and ConfigurationArray).

Thanks in advance for review :)

### DIFF
--- a/lib/Doctrine/Migrations/Configuration/Configuration.php
+++ b/lib/Doctrine/Migrations/Configuration/Configuration.php
@@ -8,7 +8,12 @@ use Doctrine\Migrations\Configuration\Exception\FrozenConfiguration;
 use Doctrine\Migrations\Configuration\Exception\UnknownConfigurationValue;
 use Doctrine\Migrations\Exception\MigrationException;
 use Doctrine\Migrations\Metadata\Storage\MetadataStorageConfiguration;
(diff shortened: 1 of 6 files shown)


lib/Doctrine/Migrations/Configuration/Configuration.php
```diff
@@ -135,6 +149,12 @@ public function setCustomTemplate(?string $customTemplate): void
         $this->customTemplate = $customTemplate;
     }
 
+    public function setVersionComparator(string $versionComparator): void
```
goetas, line 152: is there a need for this setter? (or functionality as part of the configurations?)

comparators are meant to be services (not configrations), thus to be setted via dependency factory class... and there this already a way to provide them. When you need external services to decide the sorting, this solution starts to be not good anymore.

(i blogged time ago on how to provide a custom comparator in https://www.goetas.com/blog/multi-namespace-migrations-with-doctrinemigrations-30/)

goetas (changes requested):

PR #1129, opened 2021-03-02
docs: address issue with incorrect code snippet

rlcurrall:
|      Q       |   A
|------------- | -----------
| Type         | docs
| BC Break     | no
| Fixed issues | NA

#### Summary

The import for `ExistingConfiguration` is fixed. Also, the `Configuration` does
not provide a *setName* method, so it throws an error when running the script.


SenseException (approved):

SenseException:
Thank you @rlcurrall

PR #1130, opened 2021-03-13
Warn users about transactions and DDL statements 

greg0ire:
#### Summary

Users should be aware that there is no guarantee all statements are
wrapped inside a transaction.


greg0ire:
>  [BC] CHANGED: Doctrine\Migrations\Version\Factory was marked "@internal"

I did that to fix the PHPStan check

derrabus (approved):

ostrolucky (approved):

Issue #1104, opened 2020-12-25
DDL migrations produce errors on PHP 8.0 + MySQL in transactional mode


lippok:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 3.0.2 (probably > 3.0)
| PHP     | 8.0
| DBAL-Driver     | PDO
| DB     | mysql 5.7 and 8.0

#### Summary

When running the command `migrations:migrate` with a ddl sql statement (e.g. `create table`) on php 8.0 I got an error:

> In Connection.php line 1761: There is no active transaction  

#### Previous behavior
Bevore php 8.0 the migration produces no errors.

#### Current behavior
DDL statements on mysql result in an implicit commit (see [mysql documentation](https://dev.mysql.com/doc/refman/8.0/en/implicit-commit.html)). This causes the [explicit commit](https://github.com/doctrine/migrations/blob/3.0.2/lib/Doctrine/Migrations/Version/DbalExecutor.php#L215) in transactional mode to fail. But only starting with php 8.0 an error is raised (probably due to php/php-src@990bb34).

To bypass the error I can disable the transactional mode by overwriting ``isTransactional`` in my migration, but in my opinion it's a breaking change in the behaviour when using php 8.

I found the same issue in a bug report for the yii framework (yiisoft/yii2#18406) that helped me to figure out what happens. They fixed it by checking [`PDO::inTransaction`](https://www.php.net/manual/de/pdo.intransaction.php) before running the commit. If I'm not wrong this method is not wrapped by the dbal package and therefore not available here.

#### How to reproduce

1. Create a migration with a ddl statement.
1. Let the transactional mode active.
1. Run the migration.

chriskapp:
It looks like this problem exists also at version `2.3.2`. This occurred on our CI server, we are using Mysql and PHP 8:

```
Fatal error: Uncaught PDOException: There is no active transaction in /home/travis/build/apioo/fusio-impl/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php on line 1761
PDOException: There is no active transaction in /home/travis/build/apioo/fusio-impl/vendor/doctrine/dbal/lib/Doctrine/DBAL/Connection.php on line 1761
Call Stack:
    0.0001     414648   1. {main}() /home/travis/build/apioo/fusio-impl/vendor/phpunit/phpunit/phpunit:0
[...]

PR #1131, opened 2021-03-13
Stop throwing exceptions on commit in PHP8+pdo_mysql combo 

ostrolucky:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1104

#### Summary

I didn't go with exception swallowing after all, because it turned out PDO has `inTransaction()` which works here.

### DIFF
--- a/lib/Doctrine/Migrations/Event/Listeners/AutoCommitListener.php
+++ b/lib/Doctrine/Migrations/Event/Listeners/AutoCommitListener.php
@@ -7,6 +7,7 @@ namespace Doctrine\Migrations\Event\Listeners;
 use Doctrine\Common\EventSubscriber;
 use Doctrine\Migrations\Event\MigrationsEventArgs;
 use Doctrine\Migrations\Events;
+use Doctrine\Migrations\Tools\TransactionHelper;
 
 /**
  * Listens for `onMigrationsMigrated` and, if the connection has autocommit
@@ -24,7 +25,7 @@ final class AutoCommitListener implements EventSubscriber
             return;
         }
 
-        $conn->commit();
+        TransactionHelper::commitIfInTransaction($conn);
     }
 
     /** {@inheritdoc} */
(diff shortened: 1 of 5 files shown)


greg0ire (approved):

derrabus (approved):

albe:
FYI: @ostrolucky @greg0ire Seeing an odd `"Transaction commit failed because the transaction has been marked for rollback only."` now with 3.1.1 on PGSQL - Tests pass with dependency fixed on 3.1.0 and nothing else changed 😬:
https://github.com/neos/neos-development-collection/pull/3311/checks?check_run_id=2141073265#step:10:30 ✔️ 
https://github.com/neos/neos-development-collection/runs/2134647627?check_suite_focus=true#step:19:19 ❌

I'm not sure why that only occurs with PGSQL and the Behat tests, and also only in the tests of our CMS and not even the underlying Flow Framework:
https://github.com/neos/flow-development-collection/pull/2365/checks#step:10:32

That hints at that it might be very edge-case scenario, but still something behaves unexpectedly different now. I will try to find out more.

albe:
I spent some time analyzing the cause. See https://github.com/neos/neos-development-collection/pull/3311#issuecomment-803560353 for the steps I did.
One of our migrations we run seems to leave a failed transaction open missing a rollback with 3.1.1
Looking at the code, the only possible explanation I have (so far) is this:
The `$wrappedConnection instanceof PDO && ! $wrappedConnection->inTransaction()` must match (we use pdo_pgsql, which isn't affected by the bug this fixes so this is a behaviour change) and prevent one transaction from committing. So this creates a discrepancy between started/committed transactions?
Also note how this happens when migrations are run on an already current schema (`doctrine:create` without marking migrations applied) - so the migration statements are logically failing. So a failing transaction that is not committed at the expected point in time is probably caught somewhere else without a rollback, leading to the "marked for rollback only" transaction error at the next `commit()`?

Edit: This hardly explains why the same doesn't happen in our Framework tests though, where we also should have applied migrations onto a freshly created schema.

Anyway, the core thing is: The condition right now slightly changes commit behaviour outside of the scope of the original issue (PHP8+pdo_mysql).

PR #1143, opened 2021-04-03
Rollback if and only if inside a transaction

### LINKED ISSUE #1139: Error "There is no active transaction" instead of Error about SQL syntax
### Bug Report

Syntax errors in SQL does not always display a syntax error message. It can display the error "There is no active transaction"

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.1.1

#### Summary

When executing one migration with two DDL statements, if the second one has a syntax error, then the error displayed by the command is not a syntax error, but the error "There is no active transaction".
Note : I also tried when executing several migrations and in this case the error message showed is a syntax error message. It seems to happen only when migrating one file.

#### Current behavior

A syntax error can display the error "There is no active transaction"

#### How to reproduce

1. Have all your migrations up to date
2. Create a migration where you have two ALTER TABLE statements, and make any syntax error in the second statement.
3. Run this migration

Code example :
`$this->addSql('ALTER TABLE article ADD publish_at DATETIME DEFAULT NULL');`
`$this->addSql('ALTER TABLE article CHANGE publish_at DATETIME NOT NULL');`

#### Expected behavior

A syntax error message should be displayed

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1139

#### Summary

Similarly to what happens when attempting to commit an already closed
connection when combining PHP 8 and pdo_mysql, an error is thrown when
attempting to rollback. We use the same solution as for commit(), that
is we check that we actually are inside a transaction when possible to do so.

Fixes #1139

# How can I test this?

```shell
composer config repositories.greg0ire vcs https://github.com/greg0ire/doctrine-migrations
composer require doctrine/migrations "dev-rollback-iff-in-transaction as 2.3.3"
```

### DIFF
--- a/lib/Doctrine/Migrations/Migrator.php
+++ b/lib/Doctrine/Migrations/Migrator.php
@@ -196,7 +196,7 @@ class Migrator
             $this->configuration->dispatchMigrationEvent(Events::onMigrationsMigrated, $direction, $dryRun);
         } catch (Throwable $e) {
             if ($allOrNothing) {
-                $connection->rollBack();
+                TransactionHelper::rollbackIfInTransaction($connection);
             }
 
             throw $e;
--- a/lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/DiffCommand.php
@@ -86,7 +86,7 @@ EOT
                 null,
                 InputOption::VALUE_OPTIONAL,
                 'Max line length of unformatted lines.',
-                120
+                '120'
             )
             ->addOption(
                 'check-database-platform',
--- a/lib/Doctrine/Migrations/Tools/Console/Command/DumpSchemaCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/DumpSchemaCommand.php
@@ -61,7 +61,7 @@ EOT
                 null,
                 InputOption::VALUE_OPTIONAL,
                 'Max line length of unformatted lines.',
-                120
+                '120'
             );
     }
 
--- a/lib/Doctrine/Migrations/Tools/TransactionHelper.php
+++ b/lib/Doctrine/Migrations/Tools/TransactionHelper.php
@@ -14,13 +14,28 @@ final class TransactionHelper
 {
     public static function commitIfInTransaction(Connection $connection): void
     {
-        $wrappedConnection = $connection->getWrappedConnection();
-
-        // Attempt to commit while no transaction is running results in exception since PHP 8 + pdo_mysql combination
-        if ($wrappedConnection instanceof PDO && ! $wrappedConnection->inTransaction()) {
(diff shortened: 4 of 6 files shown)


greg0ire:
@AntoineRoue please review and test

ostrolucky (approved):

AntoineRoue:
@greg0ire I've just seen your message, is it ok or do you still need me to test ?

greg0ire:
It's best if you test, if you can't we'll merge as is.

AntoineRoue:
Ok no problem. I had issues with composer auth token yesterday but I think there is a recent update that fixes it. I'll try this evening.

SenseException (approved):

AntoineRoue:
@greg0ire I managed to get your work, but when I run any doctrine command, I have the error
`2021-04-07T19:12:59+00:00 [critical] Uncaught Error: Too few arguments to function Doctrine\Migrations\Configuration\Configuration::__construct(), 0 passed in var\cache\dev\ContainerF0kiXSY\getDoctrine_Migrations_DependencyFactoryService.php on line 23 and at least 1 expected
exit status 255`

greg0ire:
I know that the last version of the migrations bundle is buggy, not sure if that is why you experience this. When you installed my fork, did any other package upgrade as well? In particular, did the migrations bundle upgrade?

AntoineRoue:
At first I could not install your package and the migration bundle package because of the dependency to "doctrine/migrations: ^3.1" in doctrine-migrations-bundle.
So I forked doctrine-migrations-bundle and created a branch from "3.1.x" to remove the dependency to "doctrine/migrations: ^3.1" (now that I think about it, maybe there is a composer option that ignores dependencies...)
Then I installed my 3.1.x no-dependency version of doctrine-migrations-bundle and your package. With this done, any doctrine command does not work.
EDIT : I just saw that somehow your branch rollback-iff... is 300 commits behind 3.1.x.

greg0ire:
> I just saw that somehow your branch rollback-iff... is 300 commits behind 3.1.x.

Yes, I'm targeting 2.3.x since it has the bug too

greg0ire:
I just pushed a branch that is the result of the merge of 3.1.x and this one so that you can test.

```shell
composer config repositories.greg0ire vcs https://github.com/greg0ire/doctrine-migrations
composer require doctrine/migrations "dev-upmerge-rollback-iff-in-transaction as 3.1.1"
```

AntoineRoue:
Oh I didn't realize that ! By the way, your first command should be `composer config repositories.greg0ire vcs https://github.com/greg0ire/migrations`.
I'll test this tomorrow evening.

AntoineRoue:
Ok, we did it! I was able to test it, and your fix works.

greg0ire:
Great, thanks @AntoineRoue , and good luck with that certification!

PR #1150, opened 2021-04-24
cache some metadata state to avoid querying all the time the database…

goetas:
… schema

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues |https://github.com/doctrine/DoctrineMigrationsBundle/issues/417

### DIFF
--- a/lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
+++ b/lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
@@ -37,6 +37,12 @@ use const CASE_LOWER;
 
 final class TableMetadataStorage implements MetadataStorage
 {
+    /** @var bool */
+    private $isInitialized;
+
+    /** @var bool */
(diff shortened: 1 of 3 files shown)


greg0ire (approved):

greg0ire:
Ah there are tests failures now

goetas:
now it should be better! :) (ive also added tests)

lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
```diff
@@ -177,6 +193,10 @@ private function needsUpdate(Table $expectedTable): ?TableDiff
 
     private function isInitialized(): bool
     {
+        if ($this->isInitialized) {
```
stof, line 196: shouldn't you also memoize the result of `tablesExist`, and distinguish `null` vs `false` (we know it is not initialized yet) ?

Issue #1155, opened 2021-04-27
Using replicas config


AlexZodov:
|    Q        |   A
|------------ | -----
| Version     | 3.1.2

 Other packages versions: 
doctrine/dbal 2.13.1 
doctrine/migrations 3.1.2 
doctrine/doctrine-migrations-bundle 3.1.1 
doctrine/bundle 2.3.1

### Support Question

My config looks like that:

```
doctrine:
    dbal:
        default_connection: diary
        connections:
            shop:
                url: '%env(resolve:SHOP_DATABASE_URL_MASTER)%'
                options:
                    1002: 'SET sql_mode=(SELECT REPLACE(@@sql_mode, "ONLY_FULL_GROUP_BY", ""))'
                replicas:
                    replica1:
                        url: '%env(resolve:SHOP_DATABASE_URL_SLAVE)%'
                    replica2:
                        url: '%env(resolve:SHOP_DATABASE_URL_MASTER)%' 
```

and I`m getting error

```
In PrimaryReadReplicaConnection.php line 140:
  Passing a connection name as first argument is not supported anymore. Use ensureConnectedToPrimary()/ensureConnectedToReplica() instead. 
```

on each try of command `doctrine:migrations:diff --em=shop_em` or `doctrine:migrations:diff`

eshishko:
Getting same error for `doctrine:migrations:status`. On 3.1.1 version no error.

Gemineye:
When do we get the next version 3.1.3 with this fix?

Webonaute:
I used this to fix the issue temporary in my project.

```
use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection as DoctrinePrimaryReadReplicaConnection;
class PrimaryReadReplicaConnection extends DoctrinePrimaryReadReplicaConnection
{


    /**
     * @param string|null $connectionName
     *
     * @return bool
     * @deprecated This is just to allow smooth migration.
     */
    public function connect($connectionName = null)
    {
        if ($connectionName !== null) {
            $connectionName = $connectionName === 'master' ? 'primary' : 'replica';
            $this->performConnect($connectionName);
        }

        return $this->performConnect();
    }
```

```
doctrine:
  dbal:
    wrapper_class: 'App\Component\Doctrine\DBAL\Connections\PrimaryReadReplicaConnection'
```

jlekowski:
[...]

Issue #1154, opened 2021-04-27
Using replicas in dbal configuration


AlexZodov:
|    Q        |   A
|------------ | -----
| Version     | 3.1.2

 Other packages versions: 
doctrine/dbal 2.13.1 
doctrine/migrations 3.1.2 
doctrine/doctrine-migrations-bundle 3.1.1 
doctrine/bundle 2.3.1

### Support Question

My config looks like that:

`doctrine:
    dbal:
        default_connection: diary
        connections:
            shop:
                url: '%env(resolve:SHOP_DATABASE_URL_MASTER)%'
                options:
                    1002: 'SET sql_mode=(SELECT REPLACE(@@sql_mode, "ONLY_FULL_GROUP_BY", ""))'
                replicas:
                    replica1:
                        url: '%env(resolve:SHOP_DATABASE_URL_SLAVE)%'
                    replica2:
                        url: '%env(resolve:SHOP_DATABASE_URL_MASTER)%'`

and I`m getting error

`In PrimaryReadReplicaConnection.php line 140:
  Passing a connection name as first argument is not supported anymore. Use ensureConnectedToPrimary()/ensureConnectedToReplica() instead. `

on each try of command `doctrine:migrations:diff --em=shop_em`

AlexZodov:
`
doctrine:
  dbal:
        default_connection: diary
        connections:
            shop:
                url: '%env(resolve:SHOP_DATABASE_URL_MASTER)%'
                options:
                    1002: 'SET sql_mode=(SELECT REPLACE(@@sql_mode, "ONLY_FULL_GROUP_BY", ""))'
                replicas:
                    replica1:
                        url: '%env(resolve:SHOP_DATABASE_URL_SLAVE)%'
                    replica2:
                        url: '%env(resolve:SHOP_DATABASE_URL_MASTER)%'
`

PR #1156, opened 2021-04-28
Do not pass connection name to `PrimaryReadReplicaConnection::connect()`

frankverhoeven:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1154 #1155 

#### Summary

Passing a connection name to `PrimaryReadReplicaConnection::connect()` will result in an exception being thrown.
This change will remove the `master` name from `TableMetadataStorage::isInitialized()` to prevent that exception.

### DIFF
--- a/lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
+++ b/lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
@@ -198,7 +198,7 @@ final class TableMetadataStorage implements MetadataStorage
         }
 
         if ($this->connection instanceof PrimaryReadReplicaConnection) {
-            $this->connection->connect('master');
+            $this->connection->connect();
         }
 
         return $this->schemaManager->tablesExist([$this->configuration->getTableName()]);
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
```diff
@@ -198,7 +198,7 @@ private function isInitialized(): bool
         }
 
         if ($this->connection instanceof PrimaryReadReplicaConnection) {
-            $this->connection->connect('master');
+            $this->connection->connect();
```
stof: The equivalent of the old `->connect('master')` API of MasterSlaveConnection is `->ensureConnectedToPrimary()`, not `->connect()`

stof (changes requested):

stof (approved):

SenseException (approved):

greg0ire:
Should this target 2.3.x, or does the fix not apply there?

stof:
@greg0ire no. 2.3.x is not affected. It properly uses the new API for the new version of DBAL already (but `TableMetadataStorage` does not exist at all in 2.3.x so a merge up did not fix it for 3.x)

greg0ire:
Let's merge this then! Thanks @frankverhoeven !

Issue #1153, opened 2021-04-26
RFC: config key to generate non-transactional migrations


greg0ire:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | yes
| BC Break    | no

#### Summary

To fix issues related to PHP 8 finally exposing a long living issue with implicit transactions, it would be great to give our users the necessary tool to easily generate non-transactional migrations. A simple boolean could govern the generation of an override of `AbstractMigration::isTransactional()` that would return false. They would still be able to have transactional DDL migrations by dropping that override.

When implementing and reviewing this, we should be mindful of `all_or_nothing`, which could be set to `true` iff the boolean defined in this RFC is set to `true` as well.

The silencing mechanism introduced in #1131 and #1143 could be deprecated , and then dropped in the next major release.

A shell script could be provided as part of the UPGRADE guide to help people make their existing migrations non-transactional via `sed`.

stof:
> A shell script could be provided as part of the UPGRADE guide to help people make their existing migrations non-transactional via `sed`.

Or maybe this could be contributed to rector, which might be more reliable than a sed replacement.

greg0ire:
Alternatively, we could use a setting that could take more than 2 values:

- `always`: no override method generated, disables `TransactionHelper`
- `silence_failures`: is the default setting, deprecated from the start, no override method generated, `TransactionHelper` is used 
- `never`: generates an override method, disables `TransactionHelper`

That way we can get rid of `TransactionHelper`

greg0ire:
Or maybe 2 booleans:

- `transactional`: whether to generate an override method
- `silence_transaction_failures`: setting that one to true is the default, but is deprecated.

But then we would have to deprecate `silence_transaction_failures` in 4.x and remove it in 5.x… maybe the best way is a node that is ideally boolean, but can take a special string value (`silence_failures`)

greg0ire:
Now that I think of it, I can't think of a situation where it would be useful to have `silence_transaction_failures` set to `false`, so just the first boolean will be useful.

VincentLanglet:
@greg0ire I was going to open an issue because the migration was not transactional and I just saw your issue saying it's transactionnal.

The following migration
```
        $this->addSql('DROP TABLE asv_relaunch');
        $this->addSql('ALTER TABLE asv_client DROP next_check_by_lab');
[...]

PR #1157, opened 2021-04-28
Configurable isTransactional() override

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | #1153 

#### Summary

This introduces a new configuration key that allows the user to generate
a method override that disables transaction.

The plan is to trigger a deprecation when `TransactionHelper` is in use in a subsequent PR, and then remove `TransactionHelper` in 4.0.0

### DIFF
--- a/lib/Doctrine/Migrations/Configuration/Migration/ConfigurationArray.php
+++ b/lib/Doctrine/Migrations/Configuration/Migration/ConfigurationArray.php
@@ -61,6 +61,9 @@ final class ConfigurationArray implements ConfigurationLoader
             'all_or_nothing' => static function ($value, Configuration $configuration): void {
                 $configuration->setAllOrNothing(is_bool($value) ? $value : BooleanStringFormatter::toBoolean($value, false));
             },
+            'transactional' => static function ($value, Configuration $configuration): void {
+                $configuration->setAllOrNothing(is_bool($value) ? $value : BooleanStringFormatter::toBoolean($value, true));
+            },
             'check_database_platform' =>  static function ($value, Configuration $configuration): void {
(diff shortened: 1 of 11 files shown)


greg0ire:
cc @ste93cry

ostrolucky:
I like https://github.com/doctrine/migrations/pull/1152 as it is. Enabling transactions by default for databases which don't support transactional DDL is not smart. Disabling it for such databases is worth the slight behavioural change, IMHO.

goetas:
I prefer the changes in this PR compared to mine in https://github.com/doctrine/migrations/pull/1152. This is backward compatible.

goetas (approved):

lib/Doctrine/Migrations/Configuration/Migration/ConfigurationArray.php
```diff
@@ -61,6 +61,9 @@ public function getConfiguration(): Configuration
             'all_or_nothing' => static function ($value, Configuration $configuration): void {
                 $configuration->setAllOrNothing(is_bool($value) ? $value : BooleanStringFormatter::toBoolean($value, false));
             },
+            'transactional' => static function ($value, Configuration $configuration): void {
+                $configuration->setAllOrNothing(is_bool($value) ? $value : BooleanStringFormatter::toBoolean($value, true));
```
BafS, line 65: Why don't we use the "setTransactional" setter here?
```php
$configuration->setTransactional(is_bool($value) ? $value : BooleanStringFormatter::toBoolean($value, true));
```

greg0ire: I think it was a copy/paste mistake

PR #1161, opened 2021-05-24
Bugfix: all-or-nothing from configuration must not be ignored

grachevko:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes/no
| Fixed issues | 

#### Summary

all-or-nothing from config always replaced by default options from command


greg0ire:
The build fails because of https://github.com/box-project/box/issues/555

greg0ire (approved):

greg0ire:
Thanks @grachevko !

Issue #1166, opened 2021-06-11
Setting single migration as non-transaction broken


shadowhand:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 3.1.3

#### Summary

When setting a single migration as non-transactional, the class setting is ignored and produces an error:

```
REATE INDEX CONCURRENTLY cannot run inside a transaction block             
```

This migration has:

```php
public function isTransactional(): bool
{
    return false;
}
```

#### Previous behavior

Migration class `isTransactional()` flag works as expected. Migration was working.

#### Current behavior

Migration class `isTransactional()` flag is ignored, resulting in a database error because the SQL must **NOT** be run inside a transaction.

#### How to reproduce

We are using Postgres, which has a `create index concurrently ...` expression. The `concurrently` flag cannot be run in a transaction, because it executes in a non-blocking way. So create a migration with:

```php
    public function up(Schema $schema): void
    {
        $createIndex = <<<SQL
         create index concurrently
        if not exists idx_user_name
                   on users (lower(last_name), lower(first_name))
        SQL;

        $this->addSql($createIndex);
    }

    public function down(Schema $schema): void
    {
        $dropIndex = <<<SQL
        drop index idx_user_name
        SQL;

        $this->addSql($dropIndex);
    }

    public function isTransactional(): bool
    {
        return false;
    }
```

This migration worked correctly in 3.1.2 and is broken in 3.1.3, I assume because of PR #1161.

greg0ire:
@shadowhand I suspect `all-or-nothing` is configured to `true` in your setup… can you please show us your configuration?

shadowhand:
@greg0ire you're correct:

```php
return [
    'table_storage' => [
        'table_name' => 'migrations',
        'version_column_name' => 'version',
        'version_column_length' => 191,
        'executed_at_column_name' => 'executed_at',
        'execution_time_column_name' => 'execution_time',
    ],
    'migrations_paths' => [
        'Acme\\Framework\\Migration' => 'src/Framework/Migration',
    ],
    'all_or_nothing' => true,
    'check_database_platform' => false,
];
```

This seems like a significant BC break though, given that `all-or-nothing` didn't conflict with per-migration `isTransactional: false` setting.

greg0ire:
Thanks for confirming my hunch! I suppose you tried setting it to `false` and it fixed your issue.
[...]

PR #1167, opened 2021-06-15
Check for all-or-nothing conflict in migrations

shadowhand:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1166

#### Summary

Creates a new `MigrationConfigurationConflict` exception that is thrown when `all-or-nothing` is enabled but a migration uses `isTransactional() => false`.

### DIFF
--- a/lib/Doctrine/Migrations/DbalMigrator.php
+++ b/lib/Doctrine/Migrations/DbalMigrator.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\Migrations;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\Migrations\Exception\MigrationConfigurationConflict;
 use Doctrine\Migrations\Metadata\MigrationPlanList;
 use Doctrine\Migrations\Query\Query;
 use Doctrine\Migrations\Tools\BytesFormatter;
@@ -65,6 +66,7 @@ class DbalMigrator implements Migrator
         $allOrNothing = $migratorConfiguration->isAllOrNothing();
 
         if ($allOrNothing) {
+            $this->assertAllMigrationsAreTransactional($migrationsPlan);
             $this->connection->beginTransaction();
         }
 
@@ -87,6 +89,15 @@ class DbalMigrator implements Migrator
         }
 
         return $sql;
+    }
+
+    private function assertAllMigrationsAreTransactional(MigrationPlanList $migrationsPlan): void
+    {
+        foreach ($migrationsPlan->getItems() as $plan) {
+            if (! $plan->getMigration()->isTransactional()) {
+                throw MigrationConfigurationConflict::migrationIsNotTransactional($plan->getMigration());
+            }
+        }
     }
 
     /**
--- a/lib/Doctrine/Migrations/Exception/MigrationConfigurationConflict.php
+++ b/lib/Doctrine/Migrations/Exception/MigrationConfigurationConflict.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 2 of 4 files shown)


lib/Doctrine/Migrations/DbalMigrator.php
```diff
@@ -89,6 +91,15 @@ private function executeMigrations(
         return $sql;
     }
 
+    private function assertAllMigrationsAreTransactional(MigrationPlanList $migrationsPlan): void
+    {
+        foreach ($migrationsPlan->getItems() as $plan) {
+            if ($plan->getMigration()->isTransactional() === false) {
```
greg0ire: No need for a strict comparison here, `isTransactional` is a boolean, that's enforced in the signature of the method, which means we can use the more expressive
```suggestion
            if (! $plan->getMigration()->isTransactional()) {
```

lib/Doctrine/Migrations/Exception/MigrationConfigurationConflict.php
```diff
@@ -0,0 +1,26 @@
...
+    public static function migrationIsNotTransactional(AbstractMigration $migration): self
+    {
+        return new self(
+            sprintf(
+                "Context: attempting to execute migrations with all-or-nothing enabled\n" .
+                "Problem: migration %s is marked as non-transactional\n" .
+                'Solution: disable all-or-nothing in configuration or by command-line option, or enable transactions for all migrations',
+                get_class($migration)
+            )
+        );
```
greg0ire: Let's use linebreaks only when necessary, and let's use heredoc to avoid having to concatenated 3 strings and manage line returns.
```suggestion
        return new self(sprintf(
            <<<'EXCEPTION'
Context: attempting to execute migrations with all-or-nothing enabled
Problem: migration %s is marked as non-transactional
Solution: disable all-or-nothing in configuration or by command-line option, or enable transactions for all migrations
EXCEPTION,
            get_class($migration)
        ));
```

greg0ire (commented):
Top-notch work 👍 Only a few nitpicks. When reading the docs, people might wonder how all-or-nothing plays with `isTransactional()`. Should we answer their question in a note?

lib/Doctrine/Migrations/Exception/MigrationConfigurationConflict.php
shadowhand: I didn't want to use heredoc because the PHP version requirement is >=7.2, which means we can't indent the heredoc, which is incredibly ugly to me. 🤷‍♂️

shadowhand:
@greg0ire committed those suggestions.

greg0ire:
Sorry, I think I didn't write 7.2-compatible code :sweat:

lib/Doctrine/Migrations/Exception/MigrationConfigurationConflict.php
```diff
@@ -0,0 +1,26 @@
...
+final class MigrationConfigurationConflict extends UnexpectedValueException implements MigrationException
+{
+    public static function migrationIsNotTransactional(AbstractMigration $migration): self
+    {
+        return new self(sprintf(
+            <<<'EXCEPTION'
+Context: attempting to execute migrations with all-or-nothing enabled
+Problem: migration %s is marked as non-transactional
+Solution: disable all-or-nothing in configuration or by command-line option, or enable transactions for all migrations
+EXCEPTION,
```
greg0ire:
```suggestion
EXCEPTION
                ,
```

greg0ire (approved):

greg0ire:
Thanks @shadowhand !

Issue #1169, opened 2021-06-19
Deprecate silencing


greg0ire:
Some platforms do not support transactions for DDL statements, and doctrine/migrations historically swipes the issue under the rug by not letting uninformed users know about it. Since PHP 8, we even silence the issue suddenly made visible by PDO. Let's stop doing that.

MySQL users should typically set `transactional` to `false` in their configuration and manually edit the `isTransactional` method override generated in their migrations to `return true` (or drop it entirely) if they run DDL migrations and want those to be transactional.

goetas:
Done in https://github.com/doctrine/migrations/pull/1170

PR #1170, opened 2021-06-19
Deprecate silencing transaction-related errors

greg0ire:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #1169 

#### Summary

Some platforms do not support transactions for DDL statements, and
doctrine/migrations historically swipes the issue under the rug by not
letting uninformed users know about it. Since PHP 8, we even silence the
issue suddenly made visible by PDO. Let's stop doing that.

MySQL users should typically set transactional to false in their
configuration and manually edit the isTransactional method override
generated in their migrations to return true (or drop it entirely) if
they run DDL migrations and want those to be transactional.

### DIFF
--- a/lib/Doctrine/Migrations/Tools/TransactionHelper.php
+++ b/lib/Doctrine/Migrations/Tools/TransactionHelper.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\Migrations\Tools;
 
 use Doctrine\DBAL\Connection;
+use Doctrine\Deprecations\Deprecation;
 use PDO;
 
 /**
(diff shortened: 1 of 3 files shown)


lib/Doctrine/Migrations/Tools/TransactionHelper.php
```diff
@@ -15,6 +16,18 @@ final class TransactionHelper
     public static function commitIfInTransaction(Connection $connection): void
     {
         if (! self::inTransaction($connection)) {
+            Deprecation::trigger(
```
goetas, line 19: it would be nice to know which migration triggered the error... but i guess that here there is not enough context to know that... :(

goetas (approved):

PR #1171, opened 2021-06-21
Skipped migration should not generate an error log

### LINKED ISSUE #1164: Skipped migration should not generate an error log
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.1.2

#### Summary

A skipped migration is logged as an error.

#### Current behavior

We have Sentry wired on Monolog, so any log with `level >= NOTICE` is reported to Sentry.
Whenever some migrations are skipped (as expected), we now have Sentry reports about it, since it's logged as an error:

https://github.com/doctrine/migrations/blob/8274a0a52bb801ff2424ad4533449985c30bea7c/lib/Doctrine/Migrations/Version/DbalExecutor.php#L273-L280

#### How to reproduce

Generate a new migration with:

```php
    public function preUp(Schema $schema): void 
    {
        $this->skipIf(true, 'Some normal and expected reason this should be skipped');    
    }
```

#### Expected behavior

Use another log level, more appropriated? A level INFO should be enough?
If this level is legit, we might add a dedicated channel in the Symfony bundle, so it can be excluded.

ogizanagi:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #1164 

#### Summary

Changed the log level to `NOTICE` for skipped migrations, according to #1164

### DIFF
--- a/lib/Doctrine/Migrations/Version/DbalExecutor.php
+++ b/lib/Doctrine/Migrations/Version/DbalExecutor.php
@@ -270,7 +270,7 @@ final class DbalExecutor implements Executor
     private function logResult(Throwable $e, ExecutionResult $result, MigrationPlan $plan): void
     {
         if ($result->isSkipped()) {
-            $this->logger->error(
+            $this->logger->notice(
                 'Migration {version} skipped during {state}. Reason: "{reason}"',
                 [
                     'version' => (string) $plan->getVersion(),


greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @ogizanagi !

PR #1174, opened 2021-07-03
implementing newSubset and unavailableSubset for migration lists

chr-hertel:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | https://github.com/doctrine/DoctrineMigrationsBundle/pull/421#discussion_r662788109

#### Summary

As as result of @greg0ire review in https://github.com/doctrine/DoctrineMigrationsBundle/pull/421 I added
* ExecutedMigrationList->unavailableSubset(AvailableMigrationList)
* AvailableMigrationList->newSubset(ExecutedMigrationList)

Includes tests and refactoring of CurrentMigrationStatusCalculator

### DIFF
--- a/lib/Doctrine/Migrations/Metadata/AvailableMigrationsList.php
+++ b/lib/Doctrine/Migrations/Metadata/AvailableMigrationsList.php
@@ -9,6 +9,7 @@ use Doctrine\Migrations\Exception\MigrationNotAvailable;
 use Doctrine\Migrations\Exception\NoMigrationsFoundWithCriteria;
 use Doctrine\Migrations\Version\Version;
 
+use function array_filter;
 use function array_values;
 use function count;
 
@@ -80,5 +81,12 @@ final class AvailableMigrationsList implements Countable
         }
 
         throw MigrationNotAvailable::forVersion($version);
+    }
+
+    public function newSubset(ExecutedMigrationsList $executedMigrations): self
+    {
+        return new self(array_filter($this->getItems(), static function (AvailableMigration $migration) use ($executedMigrations): bool {
+            return ! $executedMigrations->hasMigration($migration->getVersion());
+        }));
     }
 }
--- a/lib/Doctrine/Migrations/Metadata/ExecutedMigrationsList.php
+++ b/lib/Doctrine/Migrations/Metadata/ExecutedMigrationsList.php
@@ -9,6 +9,7 @@ use Doctrine\Migrations\Exception\MigrationNotExecuted;
 use Doctrine\Migrations\Exception\NoMigrationsFoundWithCriteria;
 use Doctrine\Migrations\Version\Version;
 
+use function array_filter;
 use function array_values;
 use function count;
 
@@ -81,5 +82,12 @@ final class ExecutedMigrationsList implements Countable
         }
 
         throw MigrationNotExecuted::new((string) $version);
+    }
+
+    public function unavailableSubset(AvailableMigrationsList $availableMigrations): self
+    {
+        return new self(array_filter($this->getItems(), static function (ExecutedMigration $migration) use ($availableMigrations): bool {
+            return ! $availableMigrations->hasMigration($migration->getVersion());
+        }));
     }
 }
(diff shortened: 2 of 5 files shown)


greg0ire:
I retargeted to 3.2.x since this is a new feature. I'm also working on fixing the build at https://github.com/doctrine/migrations/pull/1175

greg0ire (approved):

goetas:
I'm not 100% sure this is a good idea.

When developing the initial version of the 3.0 release I did something similar (adding some "convenience" methods in the collection classes that were allowing to filter/manipulate data related to that collection).

I've realized that such approach was not working well since such logic sometimes might require external services or uses the collection classes as placeholder for utility methods... and the number of methods in there was growing and growing every commit i was doing.

Said methods have been removed with https://github.com/doctrine/migrations/pull/858/commits/60a1d47f9faef3e95a9a7e2f934461c98c4b52ef and their functionality have been moved to the migrations calculator (that now become a much more important class in the migrations project).

Would it make sense to have a `CachedMetadataStorage` that decorates the initial `Doctrine\Migrations\Metadata\Storage\MetadataStorage` offering some cached version of the migrations status? in this way multiple calls to the MetadataStorage instance would lead to only one set of queries (that would fix globally the performance issue we are trying to solve in the bundle...?).

(this is only one possible solution... do you see any other?)

greg0ire:
IMO, since no "use" statements were added, it means the methods that were added are not out of place. At least they are not requiring external services (but do require external classes in the same namespace).

> the number of methods in there was growing and growing every commit i was doing

IMO that's for the best, it's far better to have this than to have "utility methods" in some `Util` namespace or whatever. Maybe the other methods you describe did not really belong to the collection classes?

I think we should double down and use the newly introduced methods in `CurrentMigrationStatusCalculator`

Regarding the cache solution, it would involve having an available cache pool, or should we force the usage of a per-http-request cache?

goetas:
>IMO, since no "use" statements were added, it means the methods that were added are not out of place. At least they are not requiring external services (but do require external classes in the same namespace).
>> the number of methods in there was growing and growing every commit i was doing

>IMO that's for the best, it's far better to have this than to have "utility methods" in some `Util` namespace or whatever. Maybe the other methods you describe did not really belong to the collection classes?
>I think we should double down and use the newly introduced methods in `CurrentMigrationStatusCalculator`

I do not have a strong opinion here, it was just my past experience...
If we start adding this kind of methods in collection classes, then the methods in  `CurrentMigrationStatusCalculator` in most of the cases will just proxy the call to the collection class... a downside of such approach is that in the future it will not be clear with method is the source of truth? should the user rely on the "util methods" added in the collection classes or to rely on some other service provided by the DI container having similar or same methods?

I'm ready to try it again...

> Regarding the cache solution, it would involve having an available cache pool, or should we force the usage of a per-http-request cache?

My idea here was just to have some sort of instance/array cache that stores the computation for the same class instance, that was it it.

greg0ire:
The fact that we need code similar to what's in `CurrentMigrationStatusCalculator` in the bundle should be a good indicator that we might instead want to have these methods in the collection classes. Besides, I don't think there is anything wrong with having more than adders, removers and getters in a collection class, quite the contrary: if that was the case, then we would use a generic collection class and use psalm annotations like `@template` to get some type safety.

> a downside of such approach is that in the future it will not be clear with method is the source of truth

If you have one method that proxies to another, then by definition this last one is the source of truth. It's still good to have `CurrentMigrationStatusCalculator` around though, that way the collection classes do not have to know about `MetadataStorage` or `MigrationPlanCalculator` which I believe fall into the "external services" category you mentioned before.

@chr-hertel , do you have any thoughts to share on this? It would be great to have your point of view as well.

goetas (approved):
@greg0ire your explanation makes sense, I agree with the changes in this pull request.

greg0ire:
I'm going to change it so that `CurrentMigrationStatusCalculator` reuses the new methods then 👍

greg0ire:
Thanks @chr-hertel !

PR #1180, opened 2021-07-06
Update composer.json

abdounikarim:
Hello, i'm not sure but i think we should use `symfony/cache` version like other dependencies ?

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

I think we should also use symfony/cache version in 3.4 and 4 version, but i'm not sure.
WDYT @greg0ire ?


greg0ire:
I suppose it makes sense, let's see if the build still passes.

goetas (approved):

PR #1184, opened 2021-07-18
Allow `psr/log` 2 and 3

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,7 @@
         "doctrine/deprecations": "^0.5.3",
         "doctrine/event-manager": "^1.0",
         "friendsofphp/proxy-manager-lts": "^1.0",
-        "psr/log": "^1.1.3",
+        "psr/log": "^1.1.3 || ^2 || ^3",
         "symfony/console": "^3.4 || ^4.4.16 || ^5.0",
         "symfony/stopwatch": "^3.4 || ^4.0 || ^5.0"
     },


greg0ire:
The build is still using `psr/log` 1:

```
composer why-not psr/log 2
symfony/cache  v5.3.3  requires  psr/log (^1.1)
```

derrabus:
> The build is still using `psr/log` 1:

Yes, because your CI tests against stable packages only. You can try this out locally by allowing a dev version of `symfony/cache`.

greg0ire:
Looks like waiting for the next patch version of Symfony would be enough to be sure it actually works: https://github.com/symfony/symfony/pull/42174

Let's do that?

derrabus:
Sure, there's no need to rush anything, I guess.

derrabus:
The PHPUnit job on PHP 8.0 passes with `psr/log` 3.0.0 now.

goetas (approved):

SenseException (approved):

goetas:
@greg0ire since this is marked as improvement, shouldn't target `3.3.x`?

derrabus:
I'm not familiar with the process here, sorry. Shall I relabel it as a bugfix? I don't think that simply widening the version range of a dependency really requires a minor version bump.

greg0ire:
It's more about stability than anything else: huge changes in the tests directory could go to the stable branch. If @derrabus says this change is not a big deal, then I think it's fine to merge it here.

greg0ire (approved):

greg0ire:
Thanks @derrabus!

PR #1186, opened 2021-07-27
Make sure to run migrations up if there are not executed migrations

### LINKED ISSUE #1182: Migrations out of order are skipped if latest is run
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes - it is for us, but I'm not sure if it's deliberate
| Version     | 3.0.1 (3.0.3 of Doctrine Migrations Bundle)

#### Summary

We have multiple developers working on a single monolith. It is not uncommon for developers to be adding sets of migrations (of 1 or more migration each) in different branches and for them to be subsequently merged - and for the ordering to make sense and be applied correctly in a branch, but not be strict when combined together. I.e. it doesn't matter if branches get interleaved with migrations on different areas of the platform.

We've hit a very specific case where if the latest executed migration does not exist in the branch, and there are migrations missing, the direction of migration is assumed to be DOWN rather than UP in the `findDirection` function, and as a result we end up not executing migrations. (The migrations aren't removed either, they do not execute).

This used to work, but seemingly does not anymore. It seems to work most of the time, but there appears to be some very specific cases where it does not.

A test case for this used to exist, as created here: https://github.com/doctrine/migrations/pull/244

#### Current behavior

Migrations with timestamps (or names) that are before the most recent migration will be ignored if they appear in the code base (i.e. are executed) in a different order to when they're created. The migrations do not run at all.

To further add to the confusion, `doctrine:migrations:up-to-date` will also report the migration has not been run and is therefore not up to date, so the following bash code won't work:

```
if ! doctrine:migrations:up-to-date; then
     doctrine:migrations:migrate 
fi
```

#### How to reproduce

I think I've got a test that reproduces the behaviour, and I think it's possible to reproduce with three migrations:

1. Apply a migration with a timestamp. This timestamp is the largest/latest timestamp (migration C)
2. Create a new branch without migration C, and create a new migration with an earlier timestamp (migration B). Apply this too. Your DB will now have migrations B and C applied, but your branch knows only of the code for migration B.
3. Now, create a final branch which contains an earlier still migration (migration A) and also contains the code for migration B (but not C). 

As migration C is the "latest" and migration B has already run, it is assumed that migration A has also run, and therefore will not execute.

The line that seems to make the difference is https://github.com/doctrine/migrations/blob/0058994cd52ecbf0a5b96d8fc28e7dd9c0a7e133/lib/Doctrine/Migrations/Version/SortedMigrationPlanCalculator.php#L110-L117. This will calculate that because the ExecutedMigrations includes the latest migration, that the direction must be down, which leads following states to return the wrong set of migrations and do the wrong step.

I've got a test here too: https://github.com/doctrine/migrations/compare/3.2.x...edhgoose:test-for-migrations-missing-when-using-latest

#### Expected behavior

Any missing migration added before the latest one should also be run. This is what used to happen before v3.

goetas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/migrations/issues/1182

### DIFF
--- a/lib/Doctrine/Migrations/Version/SortedMigrationPlanCalculator.php
+++ b/lib/Doctrine/Migrations/Version/SortedMigrationPlanCalculator.php
@@ -83,10 +83,10 @@ final class SortedMigrationPlanCalculator implements MigrationPlanCalculator
             throw MigrationClassNotFound::new((string) $to);
         }
 
-        $availableMigrations = $this->getMigrations();
+        $availableMigrations = $this->getMigrations(); // migrations are sorted at this point
         $executedMigrations  = $this->metadataStorage->getExecutedMigrations();
 
-        $direction = $this->findDirection($to, $executedMigrations);
+        $direction = $this->findDirection($to, $executedMigrations, $availableMigrations);
 
         $migrationsToCheck = $this->arrangeMigrationsForDirection($direction, $availableMigrations);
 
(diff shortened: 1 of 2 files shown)


greg0ire (approved):

greg0ire:
Thanks @goetas !

PR #1187, opened 2021-07-27
The write-sql option can be a filename

### LINKED ISSUE #1162: Error : The path "/tmp/migration.sql" not writeable!
running 
```
bin/console doctrine:migration:migrate --write-sql=/tmp/migration.sql --dry-run -vvv
```

generating error :
```
[ERROR] The path "/tmp/migration.sql" not writeable!
```

using _is_writable_ in the command return false if the file does not exist.
is it intentional to force the file to exist?

The workaround is to perform a touch on the file before run the script.

goetas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/migrations/issues/1162

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
@@ -12,8 +12,10 @@ use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 use function array_map;
+use function dirname;
 use function getcwd;
 use function implode;
+use function is_dir;
 use function is_string;
 use function is_writable;
 use function sprintf;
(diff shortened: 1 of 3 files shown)


lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
```diff
@@ -130,7 +132,8 @@ protected function execute(InputInterface $input, OutputInterface $output): int
             : Direction::UP;
 
         $path = $input->getOption('write-sql') ?? getcwd();
-        if (is_string($path) && ! is_writable($path)) {
+
+        if (is_string($path) && (! is_writable($path) && (! is_dir($path) && ! is_writable(dirname($path))))) {
```
SenseException: Maybe this (and the other check in MigrateCommand.php) should be put into a method to make it more readable. Something like `$this->isDirectoryWritable($path)`

goetas: good idea, done. I would like also to find a way to reduce the code duplication but I really do not know where to put that shared method... `DoctrineCommand` seems too generic for this very specific check

SenseException: I think this should be fine for now.

SenseException (approved):

PR #1192, opened 2021-08-16
Allow version aliases to be used with execute command

Padam87:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

This PR plugs in the alias resolver to the execute command in order to simplify executing a migration.
In practice, this would improve DX, because the last migration could be easily `--down`-ed, when spotting a mistake.

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
@@ -38,7 +38,10 @@ final class ExecuteCommand extends DoctrineCommand
             ->addArgument(
                 'versions',
                 InputArgument::REQUIRED | InputArgument::IS_ARRAY,
-                'The versions to execute.',
+                sprintf(
+                    'The versions to execute. Use the FQCN of the migration, or the following reserved aliases: %s',
+                    implode(', ', ['first', 'current', 'prev', 'next', 'latest'])
+                ),
                 null
             )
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
```diff
@@ -38,7 +38,10 @@ protected function configure(): void
             ->addArgument(
                 'versions',
                 InputArgument::REQUIRED | InputArgument::IS_ARRAY,
-                'The versions to execute.',
+                sprintf(
+                    'The versions to execute. Use the FQCN of the migration, or the following reserved aliases: %s',
+                    implode(', ', ['first', 'current', 'prev', 'next', 'latest'])
```
goetas, line 43: does it make sense to get this list from the interface somehow?

goetas:
Since this is a new feature, ive changed the base branch to `3.3.x`

lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php
Padam87: A `getSupportedAliases` method would be easy to impletement, but using it could be trouble, because of the verbatim return, and the `current-1` formats.
Should I go for it? If so, should it be enforced with a `supports` method, or just a simple listing?

PR #1194, opened 2021-08-22
expose all attempted execution migrations

goetas:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no
| Fixed issues | -

### DIFF
--- a/lib/Doctrine/Migrations/Metadata/ExecutedMigration.php
+++ b/lib/Doctrine/Migrations/Metadata/ExecutedMigration.php
@@ -26,11 +26,22 @@ final class ExecutedMigration
      */
     public $executionTime;
 
-    public function __construct(Version $version, ?DateTimeImmutable $executedAt = null, ?float $executionTime = null)
-    {
+    /** @var string|null */
+    private $reason;
+
+    /** @var string|null */
+    private $reasonDescription;
+
+    public function __construct(
+        Version $version,
+        ?DateTimeImmutable $executedAt = null,
+        ?float $executionTime = null,
+        ?string $reason
+    ) {
         $this->version       = $version;
         $this->executedAt    = $executedAt;
         $this->executionTime = $executionTime;
+        $this->reason        = $reason;
     }
 
     public function getExecutionTime(): ?float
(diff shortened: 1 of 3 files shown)


lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
```diff
@@ -11,8 +11,17 @@ interface MetadataStorage
...
+    /**
+     * Returns only the migrations that have been successfully executed (reason=executed)
+     */
     public function getExecutedMigrations(): ExecutedMigrationsList;
 
+    /**
+     * Returns all the for which there was an execution attempt
+     * Includes skipped, errored executed migrations.
+     */
+    public function getAllExecutedMigrations(): ExecutedMigrationsList;
```
goetas, line 23: what would be here the best way to add such functionality to this interface in a backward compatible way?

lib/Doctrine/Migrations/Metadata/ExecutedMigration.php
```diff
@@ -26,11 +26,22 @@ final class ExecutedMigration
...
+    private $reason;
+
+    /** @var string|null */
+    private $reasonDescription;
+
+    public function __construct(
+        Version $version,
+        ?DateTimeImmutable $executedAt = null,
+        ?float $executionTime = null,
+        ?string $reason
```
SenseException, line 39: Why did you decide not to use
```suggestion
        ?string $reason = null
```
?

lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
```diff
@@ -11,8 +11,17 @@ interface MetadataStorage
 {
     public function ensureInitialized(): void;
 
+    /**
+     * Returns only the migrations that have been successfully executed (reason=executed)
+     */
     public function getExecutedMigrations(): ExecutedMigrationsList;
 
+    /**
+     * Returns all the for which there was an execution attempt
```
SenseException, line 20:
```suggestion
     * Returns all the migrations for which there was an execution attempt
```

lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
SenseException: Usually something like

```
class Foo implements MetadataStorage, AllExecutedMigrationsInterfaceThingy
```

but I assume you're looking for something different?

PR #1195, opened 2021-08-24
Commands throw exception if ran twice.

Warxcell:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues |

#### Summary

Hello, we a project which is using two databases, and we have tests that runs `doctrine:migrations:migrate` for each test. 
We run commands without going out of the framework.The problem is that basically commands can be ran only single time, otherwise they throw exception. (I added test case for that).

I'm not sure how to solve that, so I accept help. :)

### DIFF
--- a/tests/Doctrine/Migrations/Tests/Tools/Console/Command/DoctrineCommandTest.php
+++ b/tests/Doctrine/Migrations/Tests/Tools/Console/Command/DoctrineCommandTest.php
@@ -5,7 +5,9 @@ declare(strict_types=1);
 namespace Doctrine\Migrations\Tests\Tools\Console\Command;
 
 use Doctrine\Migrations\Configuration\Configuration;
+use Doctrine\Migrations\Configuration\Connection\ConnectionLoader;
 use Doctrine\Migrations\Configuration\Connection\ExistingConnection;
+use Doctrine\Migrations\Configuration\Migration\ConfigurationLoader;
 use Doctrine\Migrations\Configuration\Migration\ExistingConfiguration;
 use Doctrine\Migrations\DependencyFactory;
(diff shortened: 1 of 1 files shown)


tests/Doctrine/Migrations/Tests/Tools/Console/Command/DoctrineCommandTest.php
```diff
@@ -44,6 +44,39 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+        $command = new class ($dependencyFactory) extends DoctrineCommand {
+            protected function execute(InputInterface $input, OutputInterface $output): int
+            {
+                return 0;
+            }
+        };
+        $commandTester = new CommandTester($command);
+
+        // execute once, this will freeze the dependencies.
+        $commandTester->execute(
```
goetas, line 66: im not sure if this test makes sense... is it possible to execute two times the same command ? ( i guess that you are using it as a service, not CLI, right?)

Warxcell: Yes, commands are ran from PHPUnit without going a loop throught CLI and bootstraping the project again. Basically services are instantiated, and second ran of command uses same services, and one of them (DependencyFactory) is frozen and throws exception on second ran.

Warxcell: Basically like that: https://github.com/Warxcell/files/blob/5.0/tests/Functional/AbstractFunctionalTest.php#L68-L81

Warxcell:
This is still valid issue.

PR #1197, opened 2021-09-16
Fix "transactional" configuration

BafS:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes but edge-case
| Fixed issues | 

#### Summary

The PR https://github.com/doctrine/migrations/pull/1157 introduced the "transactional" configuration but it doesn't change the configuration correctly. `$configuration->isTransactional()` is always true.


greg0ire (approved):

greg0ire:
Thanks @BafS !

PR #1199, opened 2021-09-19
run phpstan on multiple php versions

goetas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues |

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -18,6 +18,8 @@ jobs:
       matrix:
         php-version:
           - "7.2"
+          - "7.4"
+          - "8.0"
 
     steps:
       - name: "Checkout code"
--- a/lib/Doctrine/Migrations/Tools/Console/ConsoleLogger.php
+++ b/lib/Doctrine/Migrations/Tools/Console/ConsoleLogger.php
@@ -9,6 +9,7 @@ use DateTimeInterface;
 use Psr\Log\AbstractLogger;
 use Psr\Log\InvalidArgumentException;
 use Psr\Log\LogLevel;
+use Stringable;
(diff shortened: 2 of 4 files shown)


.github/workflows/static-analysis.yml
```diff
@@ -18,6 +18,8 @@ jobs:
       matrix:
         php-version:
           - "7.2"
+          - "7.4"
+          - "8.0"
```
greg0ire, line 22: This is not enough to get static analysis for DBAL v3, since the ORM  does not allow it yet, and I thought that was your goal after our private chat. Can you please clarify your intentions here?

goetas: The idea is to move one step closer to it. 
Since we were running static analysis only on 7.2, when we have added support for psr/log 2.0  (in https://github.com/doctrine/migrations/pull/1184/files) we did not notice the small imperfections about it I had to fix here

greg0ire (approved):

lib/Doctrine/Migrations/Tools/Console/ConsoleLogger.php
```diff
@@ -100,10 +103,12 @@ public function log($level, $message, array $context = []): void
     /**
      * Interpolates context values into the message placeholders.
      *
-     * @param mixed[] $context
+     * @param string|Stringable $message
```
greg0ire, line 106: I would expect PHPStan to complain about this type not existing in PHP 7, but apparently, it's not an issue… if you know why, I'm interested :slightly_smiling_face:

derrabus: `symfony/polyfill-php80` ships the interface. 🙂

derrabus (approved):

PR #1210, opened 2021-11-11
Look harder for a native connection

### LINKED ISSUE #1209: Exception on wrong transactional logic hides underlying exception
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | 
| Version     | 3.3.0

#### Summary

Flawed transaction logic throws an Exception and leads to loss of the underlying Exception thus making it hard to debug the issue.

#### Details

https://github.com/doctrine/migrations/blob/36205d5af1d00fa2348ab91fddee5f4f45511732/lib/Doctrine/Migrations/Tools/TransactionHelper.php#L60-L64

This does not work anymore, because `wrappedConnection` returns `PDO\Connection` since one of the last updates. Line 60 should be replaced by `$connection->getWrappedConnection()->getWrappedConnection();` or something like that.

This method is executed deep down in this block (line 115):
https://github.com/doctrine/migrations/blob/36205d5af1d00fa2348ab91fddee5f4f45511732/lib/Doctrine/Migrations/Version/DbalExecutor.php#L112-L118
thus resulting in trying to rollback a transaction already rolled back, therefore throwing an exception and then leads to the loss of the underlying Exception `$e`.

#### How to reproduce

Write a failing migration playing with transactions?

greg0ire:
doctrine/dbal 3 switches from inheritance to composition but still
provides an accessor for PDO, and it is named like the method in
Connection.
Unwrapping until that method no longer exists should give the innermost
connection regardless of the DBAL version

I hade to adapt some tests to ensure `getWrappedConnection` returns an object.

Fixes #1209

# How can I test this?

```shell
composer config repositories.greg0ire vcs https://github.com/greg0ire/migrations
composer require doctrine/migrations "dev-fix-compat-with-dbal-3 as 3.3.0"
```

### DIFF
--- a/lib/Doctrine/Migrations/Tools/TransactionHelper.php
+++ b/lib/Doctrine/Migrations/Tools/TransactionHelper.php
@@ -8,6 +8,8 @@ use Doctrine\DBAL\Connection;
 use Doctrine\Deprecations\Deprecation;
 use PDO;
 
+use function method_exists;
+
 /**
  * @internal
  */
@@ -57,10 +59,13 @@ DEPRECATION
 
     private static function inTransaction(Connection $connection): bool
     {
-        $wrappedConnection = $connection->getWrappedConnection();
+        $innermostConnection = $connection;
+        while (method_exists($innermostConnection, 'getWrappedConnection')) {
+            $innermostConnection = $innermostConnection->getWrappedConnection();
+        }
 
         /* Attempt to commit or rollback while no transaction is running
            results in an exception since PHP 8 + pdo_mysql combination */
-        return ! $wrappedConnection instanceof PDO || $wrappedConnection->inTransaction();
+        return ! $innermostConnection instanceof PDO || $innermostConnection->inTransaction();
     }
 }
--- a/phpstan-common.neon.dist
+++ b/phpstan-common.neon.dist
@@ -25,6 +25,11 @@ parameters:
             path: tests/Doctrine/Migrations/Tests/Stub/DoctrineRegistry.php
         - '~Call to method getVersion\(\) of deprecated class PackageVersions\\Versions\:.*~'
 
+	# https://github.com/phpstan/phpstan/issues/5982
+        -
+            message: '~^Cannot call method getWrappedConnection\(\) on class-string\|object\.~'
+            path: lib/Doctrine/Migrations/Tools/TransactionHelper.php
+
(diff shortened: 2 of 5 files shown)


derrabus:
Tagging @morozov because we currently have a discussion about deprecating the methods you use to access the PDO connection, see doctrine/dbal#4966

greg0ire:
I think in previous discussions, we established that this `TransactionHelper` was fragile, hence why I deprecated it, so IMO this is in favor of deprecating getting the wrapped connection :stuck_out_tongue:

derrabus:
Okay, in other words, this class will be broken when DBAL 4 is used eventually and that won't be a problem? In that case, the change looks good. 🙂

derrabus (approved):

greg0ire:
> Okay, in other words, this class will be broken when DBAL 4 is used eventually and that won't be a problem?

Yes. Maybe we should create `doctrine/migrations` 4.0.x and drop `TransactionHelper` to clarify that, WDYT @goetas ?

lib/Doctrine/Migrations/Tools/TransactionHelper.php
```diff
@@ -57,10 +59,13 @@ public static function rollbackIfInTransaction(Connection $connection): void
...
-        $wrappedConnection = $connection->getWrappedConnection();
+        $innermostConnection = $connection;
+        while (method_exists($innermostConnection, 'getWrappedConnection')) {
+            $innermostConnection = $innermostConnection->getWrappedConnection();
+        }
 
         /* Attempt to commit or rollback while no transaction is running
            results in an exception since PHP 8 + pdo_mysql combination */
-        return ! $wrappedConnection instanceof PDO || $wrappedConnection->inTransaction();
+        return ! $innermostConnection instanceof PDO || $innermostConnection->inTransaction();
```
morozov, line 69: If the "in transaction" method for the drivers of the platform that don't support transactional DDL is still of interest, it looks like the right course of action is to revive the [discussion](https://externals.io/message/114181) in PHP internals.

greg0ire: We are planning to drop the `TransactionHelper` class in the next major release ([ref](https://github.com/doctrine/migrations/pull/1170)), so I don't think it will be of interest.

UlrichThomasGabor:
fixes #1209

PR #1211, opened 2021-11-11
Allow Symfony 6

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

This PR allows the installation of this library in Symfony 6 projects.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -23,9 +23,6 @@ jobs:
     name: "PHPUnit"
     runs-on: "ubuntu-20.04"
 
-    env:
-      COLUMNS: 120
-
     strategy:
       matrix:
         dbal-version:
@@ -83,6 +80,12 @@ jobs:
         with:
           name: "phpunit-${{ matrix.deps }}-${{ matrix.php-version }}.coverage"
           path: "coverage.xml"
+
+  phpunit-with-symfony-6:
+    name: "PHPUnit with Symfony 6.0.x-dev"
+    uses: "doctrine/.github/.github/workflows/continuous-integration-symfony-unstable.yml@1.4.0"
+    with:
+      symfony-version-constraint: "6.0.*"
 
   upload_coverage:
     name: "Upload coverage to Codecov"
--- a/composer.json
+++ b/composer.json
@@ -31,8 +31,8 @@
         "doctrine/event-manager": "^1.0",
         "friendsofphp/proxy-manager-lts": "^1.0",
         "psr/log": "^1.1.3 || ^2 || ^3",
-        "symfony/console": "^3.4 || ^4.4.16 || ^5.0",
-        "symfony/stopwatch": "^3.4 || ^4.0 || ^5.0"
+        "symfony/console": "^3.4 || ^4.4.16 || ^5.0 || ^6.0",
+        "symfony/stopwatch": "^3.4 || ^4.0 || ^5.0 || ^6.0"
     },
     "require-dev": {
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -23,6 +23,7 @@
             "email": "contact@mikesimonson.com"
         }
     ],
+    "minimum-stability": "beta",
```
derrabus: This can be reverted when Symfony 6.0 is stable.

greg0ire: Can you maybe instead try https://github.com/doctrine/.github/pull/29 on this PR? That way if we have a dependency that is not compatible with Symfony 6 yet, we will know. Also, I'm not sure what's up with the BC check, but maybe dropping this line will fix it.

derrabus: > I'm not sure what's up with the BC check,

That check is drunk. It lists documented BC breaks of Symfony, but those are fine when bumping to a new major.

> but maybe dropping this line will fix it.

It might, but it will likely break again as soon as Symfony 6 is stable.

greg0ire: Maybe it's because this package defines Symfony commands, and thus shares some BC-breaks of Symfony?

derrabus: Yes, more precisely:

> [BC] CHANGED: The parameter $default of Symfony\Component\Console\Command\Command#addArgument() changed from no type to a non-contravariant ?mixed

This is not a breaking change, neither for the caller nor for downstream implementations.

> [BC] CHANGED: The return type of Symfony\Component\Console\Command\Command::getDefaultName() changed from no type to ?string

This is technically a BC break for code that extends one of the commands defined in the migrations package and overrides one of the methods listed there. Then again, a package that does something like that is probably aware of symfony/console because the commands are only useful when being integrated in a Symfony console application.

If we really treated this as a BC break, we would need to create a new major release for Symfony 6 support and either disallow Symfony 5 there or decalre all commands `final`. I really don't think that this is necessary.

derrabus: I have switched to your workflow. It installs the correct packages all right and the BC check failure is gone as we expected. But the tests fail with your workflow and they didn't do that with my previous `minimum-stability` hack and neither do they fail when I execute the tests locally. 🤔

greg0ire: :eyes:  The failure seems related to the `COLUMNS` env variable (which is used in `symfony/console` at some point). I think you don't reproduce the failure locally because you don't have the same terminal width or env var.

Similar error messages can be obtained locally on your branch with `COLUMNS=80 vendor/bin/phpunit`

.github/workflows/continuous-integration.yml
```diff
@@ -23,9 +23,6 @@ jobs:
     name: "PHPUnit"
     runs-on: "ubuntu-20.04"
 
-    env:
-      COLUMNS: 120
```
greg0ire, line 27: env variables are not transmitted to reused workflows, hence the failure 👍

derrabus:
Aaaaand, we're 🟢!

composer.json
derrabus: Indeed. I've moved the `COLUMNS` env to PHPUnit's config file. That's imho where the setting belongs if the tests fail without it.

greg0ire:
Should we merge https://github.com/doctrine/.github/pull/29  then?

greg0ire:
I'll let you give it a review, and then I'll tag it as 1.4.0 if it gets your approval

derrabus:
We're ready here, I guess. 🥳

greg0ire (approved):
:ok_hand:

PR #1215, opened 2021-11-21
Fix PhpFile FQCN in Advanced Configuration example

elazar:
Last namespace segment should be Migration rather than Configuration.


greg0ire (approved):

greg0ire:
Thanks @elazar !

PR #1223, opened 2022-01-10
Update configuration.rst

sukhoy94:
fix possioble option values for organize_migrations when organizing under year or year and month.
See https://symfony.com/bundles/DoctrineMigrationsBundle/current/index.html
```
    # Possible values: "BY_YEAR", "BY_YEAR_AND_MONTH", false
    organize_migrations: false
```

When using `year_and_month` or `year` option, I've got   'Invalid organize migrations mode value "year_and_month"' error.

|      Q       |   A
|------------- | -----------
| BC Break     | yes/no
| Fixed issues | 

#### Summary

fix possioble option values for organize_migrations when organizing under year or year and month.
See https://symfony.com/bundles/DoctrineMigrationsBundle/current/index.html
```
    # Possible values: "BY_YEAR", "BY_YEAR_AND_MONTH", false
    organize_migrations: false
```

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -147,7 +147,7 @@ Here are details about what each configuration option does:
 +----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
 | check_database_platform    | no         | true                         | Whether to add a database platform check at the beginning of the generated code. |
 +----------------------------+------------+------------------------------+----------------------------------------------------------------------------------+
-| organize_migrations        | no         | ``none``                     | Whether to organize migration classes under year (``year``) or year and month (``year_and_month``) subdirectories. |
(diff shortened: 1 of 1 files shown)


SenseException:
The values `year` and  `year_and_month` are used in https://github.com/doctrine/migrations/blob/3.3.x/lib/Doctrine/Migrations/Configuration/Configuration.php#L220-L236 and run by tests, like https://github.com/doctrine/migrations/blob/3.3.x/tests/Doctrine/Migrations/Tests/Configuration/Migration/YamlLoaderTest.php, with these values.

Can you please provide an example with Migrations itself instead of DoctrineMigrationsBundle where this error occurs?

sukhoy94:
@SenseException okay, got it. I thougth that it should be the same.

PR #1225, opened 2022-01-11
Replace composer/package-versions-deprecated with composer-runtime-api

stof:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Composer 2.2 requires an opt-in for plugins. So depending on the legacy plugin adds some annoyance for all users of the package.
This is dropping support for Composer 1, which is why I submit it for 3.4.x-dev rather than 3.3.x-dev.

The same change is already merged in DBAL 3.3.x-dev and there is an open PR for the ORM.
See https://github.com/symfony/symfony/issues/44726 for a discussion about that package.

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/ConsoleRunner.php
+++ b/lib/Doctrine/Migrations/Tools/Console/ConsoleRunner.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Migrations\Tools\Console;
 
+use Composer\InstalledVersions;
 use Doctrine\DBAL\Tools\Console\Helper\ConnectionHelper;
 use Doctrine\Migrations\Configuration\Connection\ExistingConnection;
 use Doctrine\Migrations\Configuration\EntityManager\ExistingEntityManager;
@@ -24,11 +25,11 @@ use Doctrine\Migrations\Tools\Console\Command\SyncMetadataCommand;
 use Doctrine\Migrations\Tools\Console\Command\UpToDateCommand;
 use Doctrine\Migrations\Tools\Console\Command\VersionCommand;
 use Doctrine\ORM\Tools\Console\Helper\EntityManagerHelper;
-use PackageVersions\Versions;
 use RuntimeException;
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Helper\HelperSet;
 
+use function assert;
 use function file_exists;
 use function getcwd;
 use function is_readable;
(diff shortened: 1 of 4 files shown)


lib/Doctrine/Migrations/Tools/Console/ConsoleRunner.php
```diff
@@ -99,7 +99,9 @@ public static function run(array $commands = [], ?DependencyFactory $dependencyF
     /** @param DoctrineCommand[] $commands */
     public static function createApplication(array $commands = [], ?DependencyFactory $dependencyFactory = null): Application
     {
-        $cli = new Application('Doctrine Migrations', Versions::getVersion('doctrine/migrations'));
+        $version = InstalledVersions::getVersion('doctrine/migrations');
+        assert($version !== null);
```
stof, line 104: `getVersion` returns `null` when the package name is provided by another package rather than being installed directly (if the package is not available in any way, an exception is thrown instead). But as we ask the version for our own package, we know that it is not provided by another package (otherwise, this LOC is not there either to execute it).

greg0ire (dismissed):

greg0ire:
Some of the build failures seem related.

stof:
@greg0ire I fixed the coding standard issue, and I also fixed the issue caused by one of the jobs downgrading composer to v1

stof:
The BC checks failure seems to be a bug in the roave package or in the docker image used for it.
The phpstan failure seems to be related to new generics added in DBAL 3

derrabus (approved):

greg0ire:
I'm working on fixing the SA issues.

greg0ire:
https://github.com/doctrine/migrations/pull/1226

greg0ire:
Also, regarding the bc check failure: https://github.com/Nyholm/roave-bc-check-docker/pull/26

stof:
@greg0ire maybe we should switch GA to install the composer package directly instead of relying on that non-official docker image to provide it.

greg0ire:
It's referenced in the README: https://github.com/Roave/BackwardCompatibilityCheck#nyholm-github-action so not although non-official, it seems endorsed. On the other hand, doing without it seems easy enough, so maybe we should indeed switch depending on how long this remains broken.

greg0ire (approved):

greg0ire:
Thanks @stof !

greg0ire:
The BC check is now repaired :)

PR #1229, opened 2022-01-12
Implement platform-aware schema comparison

### LINKED ISSUE #1227: Implement platform-aware schema comparison
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

Since `doctrine/dbal` 3.2.0, regular comparison is deprecated in favor of platform-aware-schema comparison: https://www.doctrine-project.org/2021/11/26/dbal-3.2.0.html

This should fix a number of false-positive and false-negative issues.

This involves changing this piece of code:

https://github.com/doctrine/migrations/blob/1967775546df997eb97057911d80184c91c92b9a/lib/Doctrine/Migrations/Generator/DiffGenerator.php#L95-L107

as well as

https://github.com/doctrine/migrations/blob/1967775546df997eb97057911d80184c91c92b9a/lib/Doctrine/Migrations/Provider/DBALSchemaDiffProvider.php#L46-L49

greg0ire:
This should fix many bugs because it eliminates a lot of false positive
and false negative diffs.

Closes #1227

### DIFF
--- a/lib/Doctrine/Migrations/Generator/DiffGenerator.php
+++ b/lib/Doctrine/Migrations/Generator/DiffGenerator.php
@@ -12,6 +12,7 @@ use Doctrine\DBAL\Schema\Schema;
 use Doctrine\Migrations\Generator\Exception\NoChangesDetected;
 use Doctrine\Migrations\Provider\SchemaProvider;
 
+use function method_exists;
 use function preg_match;
 use function strpos;
 use function substr;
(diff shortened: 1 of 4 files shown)


tests/Doctrine/Migrations/Tests/Generator/DiffGeneratorTest.php
```diff
@@ -92,15 +96,42 @@ static function ($name): bool {
...
+            $fromSchema->expects(self::once())
+                ->method('getMigrateToSql')
+                ->with($toSchema, $this->platform)
+                ->willReturn(['UPDATE table SET value = 2']);
+
+            $fromSchema->expects(self::once())
+                ->method('getMigrateFromSql')
+                ->with($toSchema, $this->platform)
+                ->willReturn(['UPDATE table SET value = 1']);
+        }
```
greg0ire, line 134: I tried to find a way to abide by "don't mock what you don't own" and reduce the big amount of mocking in this test, and gave up… I think it should be reworked, but I'm not sure how.

lib/Doctrine/Migrations/Generator/DiffGenerator.php
```diff
@@ -95,15 +96,27 @@ static function ($assetName) use ($filterExpression) {
 
         $toSchema = $this->createToSchema();
 
+        if (method_exists($this->schemaManager, 'createComparator')) {
+            $comparator = $this->schemaManager->createComparator();
+        }
+
+        $upSql = isset($comparator) ?
+            $comparator->compareSchemas($fromSchema, $toSchema)->toSql($this->platform) :
+            $fromSchema->getMigrateToSql($toSchema, $this->platform);
```
greg0ire, line 105: At first, I tried inlining the contents of this method to avoid the deprecation, but then I could no longer mock anything.

derrabus (approved):

morozov (approved):

Issue #1191, opened 2021-08-10
Changes are detected, and migration created, when there is no change


hgraca:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    |no
| Version     | 3.2.1

#### Summary

We generate migrations, and run them. This should put the ORM metadata and the actual DB in sync.

However, if we generate migrations again, it will detect changes and generate some stuff again.

#### Current behavior

It generates migrations when the DB and metadata are in sync.

If we apply the generated migration and generate migrations again, it generates another migration, equal to the previous one.

This seems to be related to the charset and collation.

#### How to reproduce

I created a repository to reproduce the issue: https://github.com/hgraca/migrations-shiet-show

It uses docker, so you can just checkout the repository and run the following in the CLI:

```bash
make docker-create-network
make docker-up
make docker-install-deps
make docker-generate-migration
make docker-run-migrations
make docker-generate-migration
```
Check the last migration file created, in `build/Migration/Version`.

We can try running that migration and creating a new one, surely that would put everything in sync...
```bash
make docker-run-migrations
make docker-generate-migration
```
Check the new migration created. 

It's exactly the same as the previous one, except for the class name.

##### Other info

I tried applying the changes (currently commented out) in: 

  - `config/doctrine/mappings/components/CM.GlobalTicket.Core.Component.Channel.Domain.Channel.php`
    - Tweaking this, it can actually get to the point of creating up() and down() methods that are exactly the same.
  - `config/packages/doctrine.php`

Unfortunately nothing works.

#### Expected behavior

After generating a migration and applying it, the DB and mapping should be in sync, therefore generating a new migration should result in no new migrations.

Let me know if I can be of any assistance. I tried debugging this with xdebug but it gets very confusing.
I appreciate any help you can give.
And tkx for all your work on Doctrine & friends.

goetas:
What is the SQL that is generated each time? probably there is a misconfiguration for the custom types you are adding to your app

hgraca:
Tkx for your reply @goetas, and sorry for taking long to reply

So, using the example I have above, 

the doctrine configurations this:
[Doctrine config](https://github.com/hgraca/migrations-shiet-show/blob/main/config/packages/doctrine.php)
```php
<?php

declare(strict_types=1);

use Acme\App\Infrastructure\Persistence\Doctrine\Type\ChannelConfigType;
use Acme\App\Infrastructure\Persistence\Doctrine\Type\ChannelIdType;
[...]

PR #1239, opened 2022-01-28
Fully address Comparator constructor deprecation

greg0ire:
[It seems that currently, this package is still not using platform-aware comparison](https://github.com/doctrine/migrations/issues/1191#issuecomment-1024518884).
I know this change is not sufficient for that to happen, since something needs to be done on the DBAL side,
and I don't know how necessary it is. One thing is for sure though, it addresses a deprecation.

Note that this is probably not worth releasing until a fix for https://github.com/doctrine/dbal/issues/5216 is released and there is a PR on this repository to take advantage of the new API.

### DIFF
--- a/tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
+++ b/tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
@@ -27,6 +27,7 @@ use Doctrine\Migrations\Version\ExecutionResult;
 use Doctrine\Migrations\Version\Version;
 use PHPUnit\Framework\TestCase;
 
+use function method_exists;
 use function sprintf;
 
 class TableMetadataStorageTest extends TestCase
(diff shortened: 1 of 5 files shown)


tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
```diff
@@ -500,14 +501,18 @@ protected function setUp(): void
 
     private function alterMetadataTable(): void
     {
-        $originalTable = $this->connection->getSchemaManager()
+        $schemaManager = $this->connection->getSchemaManager();
```
derrabus, line 504: Probably out of scope here, but `getSchemaManager()` is also deprecated.

tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
```diff
@@ -216,9 +217,13 @@ public function testCompleteWillAlwaysCastTimeToInteger(): void
         $config     = new TableMetadataStorageConfiguration();
         $executedAt = new DateTimeImmutable('2010-01-05 10:30:21');
 
+        $getNativeConnection = method_exists($this->getSqliteConnection(), 'getNativeConnection')
+            ? 'getNativeConnection'
+            : 'getWrappedConnection';
```
derrabus: Let's avoid `$object->$methodName` calls for the sake of readability.

```suggestion
        $connection = $this->getSqliteConnection();
        $pdo = method_exists($connection, 'getNativeConnection')
            ? $connection->getNativeConnection()
            : $connection->getWrappedConnection();
```

derrabus (approved):

tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
greg0ire: Good point, but I don't have the required energy for this right now :weary:  I also think it will be a lot easier in April when we drop support for DBAL 2.

As you see in this PR, having `phpstan-deprecation-rules` enabled makes this quite involving.

PR #1243, opened 2022-03-05
Reference - generating migrations: explain how to add a CustomSchemaP…

simon-schubert:
Don't know which Version to point to, but I guess the handling of passing a CustomSchemaProvider to the DiffCommand, when not using the ORM changed with Version 3 and the docs were not adjusted. This PR makes a suggestion to change the docs.

|      Q       |   A
|------------- | -----------
| Type         | docs improvement
| BC Break     | no
| Fixed issues | none

#### Summary

Changes the docs for "Generating Migrations" when used without the ORM: how to add a CustomSchemaProvider

### DIFF
--- a/docs/en/reference/generating-migrations.rst
+++ b/docs/en/reference/generating-migrations.rst
@@ -216,9 +216,10 @@ It simply takes a schema object to its constructor and returns it from ``createS
     $provider->createSchema() === $schema; // true
 
 By default the Doctrine Migrations command line tool will only add the diff command if the ORM is present.
-Without the ORM, you'll have to add the diff command to your console application manually, passing in your schema
-provider implementation to the diff command's constructor. Take a look at the :ref:`Custom Integration <custom-integration>`
-chapter for information on how to setup a custom console application.
+Without the ORM, you'll have to add the diff command to your console application manually and set your
+custom schema provider to the dependency factory, which will be passed to the the diff command's constructor.
+Take a look at the :ref:`Custom Integration <custom-integration>` chapter for information on how to setup a
+custom console application.
 
 .. code-block:: php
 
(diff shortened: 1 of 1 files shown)


docs/en/reference/generating-migrations.rst
```diff
@@ -217,7 +217,7 @@ It simply takes a schema object to its constructor and returns it from ``createS
 
 By default the Doctrine Migrations command line tool will only add the diff command if the ORM is present.
 Without the ORM, you'll have to add the diff command to your console application manually, passing in your schema
-provider implementation to the diff command's constructor. Take a look at the :ref:`Custom Integration <custom-integration>`
+provider implementation to the DependencyFactory. Take a look at the :ref:`Custom Integration <custom-integration>`
```
SenseException: More like "passing in your dependency factory implementation to the diff command's constructor."

SenseException (changes requested):
Thank you for this fix. 3.4 is the current lowest supported branch and should be fine.

docs/en/reference/generating-migrations.rst
simon-schubert: Thanks!

For me the part I did not understand first, was: I need to call `setDefinition(SchemaProvider::class, static fn () => $schemaProvider)` on the dependency factory. Everything else is already sufficiently explained here: https://www.doctrine-project.org/projects/doctrine-migrations/en/3.3/reference/custom-integration.html which is referenced in the sentence below "Take a look..."

What do you think about:

> Without the ORM, you'll have to add the diff command to your console application manually and set your custom schema provider to the dependency factory, which will be passed to the the diff command's constructor. Take a look ...

SenseException: I think that sounds comprehensible to me.

simon-schubert: thanks! updated

SenseException (approved):

SenseException:
Thank you @simon-schubert

PR #1248, opened 2022-03-31
PHPStan 1.5, Doctrine Coding Standard 9

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

This PR bumps PHPStan and its plugins and fixes newly discovered errors.


phpstan-common.neon.dist
```diff
@@ -35,6 +37,11 @@ parameters:
             paths:
                 - tests/Doctrine/Migrations/Tests/Generator/ClassNameGeneratorTest.php
 
+        # TODO: Be more precise about class-strings
```
derrabus, line 40: I tried to fix this one error, but it's really a rabbit hole. Someone (maybe me) could work on a fix this on a dedicated PR. For now, it should be fine to ignore the error.

lib/Doctrine/Migrations/Tools/Console/Command/CurrentCommand.php
```diff
@@ -15,7 +15,7 @@
  */
 final class CurrentCommand extends DoctrineCommand
 {
-    /** @var string */
+    /** @var string|null */
```
SenseException, line 18: Why are these nullable?

derrabus: Because they are in the parent class and types of overridden properties have to be invariant.

greg0ire (approved):

PR #1250, opened 2022-04-01
Drop support for DBAL 2, PHP 7.2, Symfony 3, Persistence 1

greg0ire:
DBAL 2 will no longer be supported tomorrow, and DBAL 3 requires PHP 7.3


lib/Doctrine/Migrations/Tools/TransactionHelper.php
```diff
@@ -72,11 +72,9 @@ private static function inTransaction(Connection $connection): bool
...
-            try {
-                return $connection->getNativeConnection();
-            } catch (LogicException $e) {
-            }
+        try {
+            return $connection->getNativeConnection();
+        } catch (LogicException $e) {
         }
 
         $innermostConnection = $connection;
```
greg0ire, line 80: Todo: check if the code below is still useful

.appveyor.yml
```diff
@@ -13,7 +13,6 @@ cache:
 
 environment:
   matrix:
-    - php: 7.2.25
```
greg0ire, line 16: Todo: determine what our policy for testing on appveyor shall be

lib/Doctrine/Migrations/Tools/TransactionHelper.php
greg0ire: From what I understood, it will be useful until DBAL 4, when we are sure that `getNativeConnection` can no longer throw a `LogicException`.

greg0ire:
:thinking: just noticed we apparently still support unmaintained versions of Symfony, we should bump that as well.

composer.json
```diff
@@ -24,32 +24,32 @@
         }
     ],
     "require": {
-        "php": "^7.2 || ^8.0",
+        "php": "^7.3 || ^8.0",
```
derrabus: Shall we drop 7.3 as well?

composer.json
```diff
@@ -24,32 +24,32 @@
...
         "composer-runtime-api": "^2",
-        "doctrine/dbal": "^2.11 || ^3.0",
+        "doctrine/dbal": "^3.0",
```
derrabus: We need to require `^3.3` here because you've dropped some compat code that we would still need for DBAL 3.2 and below, like checking for `getNativeConnection()`.

composer.json
```diff
@@ -24,32 +24,32 @@
...
         "phpstan/phpstan": "^1.5",
         "phpstan/phpstan-deprecation-rules": "^1",
         "phpstan/phpstan-phpunit": "^1.1",
         "phpstan/phpstan-strict-rules": "^1.1",
         "phpstan/phpstan-symfony": "^1.1",
-        "phpunit/phpunit": "^8.5 || ^9.4",
-        "symfony/cache": "^3.4.26 || ^4.2.12 || ^5.0 || ^6.0",
-        "symfony/process": "^3.4 || ^4.0 || ^5.0 || ^6.0",
-        "symfony/yaml": "^3.4 || ^4.0 || ^5.0 || ^6.0"
+        "phpunit/phpunit": "^9.4",
```
derrabus, line 49:
```suggestion
        "phpunit/phpunit": "^9.5",
```

PHPUnit 9.4 is dead and buried.

derrabus (changes requested):

composer.json
greg0ire: I don't see a reason to, but also, I don't see a reason not to, so sure, let's do it!

phpcs.xml.dist
```diff
@@ -17,6 +17,18 @@
...
+        <exclude-pattern>*/lib/*</exclude-pattern>
+    </rule>
+
+    <rule ref="SlevomatCodingStandard.TypeHints.ReturnTypeHint.MissingNativeTypeHint">
+        <exclude-pattern>*/lib/*</exclude-pattern>
+    </rule>
+
+    <rule ref="SlevomatCodingStandard.TypeHints.PropertyTypeHint.MissingNativeTypeHint">
+        <exclude-pattern>*/lib/*</exclude-pattern>
+    </rule>
```
greg0ire, line 22: Maybe I should remove those, `git add --patch` the parts that are backwards compatible, and restore these lines. But that would probably be best in a separate PR for the sake of review.

greg0ire: Just tried it, and it's not that complicated to untangle, I'll push it as an extra commit.

greg0ire:
I think that the 2 issues reported by the BC check can be safely ignored.

derrabus (approved):

PR #1251, opened 2022-04-04
Run tests on PHP 8.1 as well

derrabus:


greg0ire:
Should we do something similar for appveyor?

derrabus:
We can bump AppVeyor to PHP 8.1 as well, if you want. But I wouldn't create a another build matrix for different PHP versions there. That's probably a waste of resources.

greg0ire:
Ok then let's bump it, the latest PHP should be more efficient, right? 👍

derrabus:
Done.

greg0ire (approved):

PR #1254, opened 2022-04-22
Switch test setup to the ORMSetup class

derrabus:


stof:
This effectively means that the CI cannot run tests with ORM 2.11 (even though it does not do it right now, as there is no job running PHPUnit against lowest deps in the matrix), while still telling composer that there is no issue using them together.

I suggest adding a conflict rule (same than in https://github.com/doctrine/data-fixtures/pull/393#issuecomment-1106693547 btw)

greg0ire:
Fixed the cs issues :slightly_smiling_face:

greg0ire (approved):

greg0ire:
Thanks @derrabus !

PR #1262, opened 2022-07-14
fix testing on php 7.4 with the lowest composer dependencies

bendavies:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 
#### Summary

Looks to have been broken since https://github.com/doctrine/migrations/pull/1133


greg0ire:
Woops! Good catch!

derrabus (approved):

Issue #1082, opened 2020-11-26
migrate command executes migrations when --write-sql is passed, output is incomplete


donataso:
### Bug Report

According to documentation, `migrate --write-sql` should not execute any migrations and provide `doctrine_migration_versions` insert/delete queries, but it executes and does not provide version update queries.

It used to work correctly in version 2.

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 3.*

#### Summary

Described above.

#### Current behavior

* `--write-sql` does not prevent migrations from being executed
* `--write-sql` output does not include versions tracking table update queries

#### How to reproduce

Run any migration with a `--write-sql` option.

#### Expected behavior

* migration is not executed
* output contains versions tracking table update query

goetas:
did you try using he `--dry-run` parameter?

Zombaya:
Using the `--dry-run`-parameter will only solve the part of the issue where no migrations should be executed.  This is actually described in [`UPGRADE.md`](../blob/3.1.x/UPGRADE.md#L14) for doctrine version 3.1, so that is working as expected, I guess.
> The --write-sql option for migrations:migrate and migrations:execute does not imply dry-run anymore,
use the --dry-run parameter instead.

This does not solve the other problem @donataso mentions.   You still have to manually create the `INSERT INTO migration_versions (version, executed_at, execution_time) VALUES ("myversion",NOW(),NULL)` or `DELETE FROM migration_versions WHERE version = "myversion"`-query.  This used to be included in the sql-file as well but is now completely missing.

I would also prefer it to be included into the output versus updating the documentation to state that you will manually have to write that query.

goetas:
> This does not solve the other problem @donataso mentions. You still have to manually create the INSERT INTO migration_versions (version, executed_at, execution_time) VALUES ("myversion",NOW(),NULL) or DELETE FROM migration_versions WHERE version = "myversion"-query. This used to be included in the sql-file as well but is now completely missing.

> I would also prefer it to be included into the output versus updating the documentation to state that you will manually have to write that query.

makes sense

donataso:
One more additional note:
when using `--write-sql` option, PDO bind params are not included in the queries - questions marks are shown instead.

Zombaya:
> when using --write-sql option, PDO bind params are not included in the queries - questions marks are shown instead.

For reference: this was always the case, this has not been changed because of migrating to version 3.

rimvislt:
[...]

PR #1265, opened 2022-07-18
Insert/delete sql queries for metadata table during dry-run and write-sql

wiistriker:
Db user in my app doesn't have CREATE, ALTER query permissions so i cant run my migrations from php. Instead of this i write all necessary sql queries to file and run it from another user. After upgrading doctrine migrations to newer version i faced with issue, that resulted sql file doesnt contain insert/delete queries for meta table (doctrine_migration_versions). Very frustrating to insert this information myself so i ended up with this solution.

Such issue already been described in https://github.com/doctrine/migrations/issues/1082

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 1082

#### Summary

I am not sure about TableMetadataStorage::complete method. I see you using Connection::insert, Connection::delete methods. It seems like we need to replace it with Connection::execute and use already generated queries

### DIFF
--- a/lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
+++ b/lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
@@ -14,10 +14,12 @@ interface MetadataStorage
 
     public function getExecutedMigrations(): ExecutedMigrationsList;
 
+    public function complete(ExecutionResult $result): void;
+
+    public function reset(): void;
+
     /**
      * @return Query[]
      */
-    public function complete(ExecutionResult $result, bool $dryRun = false): array;
-
-    public function reset(): void;
+    public function getSql(ExecutionResult $result): array;
 }
--- a/lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
+++ b/lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
@@ -21,5 +21,5 @@ interface MetadataStorage
     /**
      * @return Query[]
      */
-    public function getSql(ExecutionResult $result): array;
+    public function getSql(ExecutionResult $result): iterable;
 }
--- a/lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
+++ b/lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\Migrations\Metadata\Storage;
 
 use Doctrine\Migrations\Metadata\ExecutedMigrationsList;
+use Doctrine\Migrations\Query\Query;
 use Doctrine\Migrations\Version\ExecutionResult;
 
 interface MetadataStorage
@@ -13,7 +14,10 @@ interface MetadataStorage
 
     public function getExecutedMigrations(): ExecutedMigrationsList;
 
-    public function complete(ExecutionResult $result, bool $dry_run = false) : array;
+    /**
+     * @return Query[]
+     */
+    public function complete(ExecutionResult $result, bool $dryRun = false): array;
 
     public function reset(): void;
 }
--- a/lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
+++ b/lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
@@ -8,6 +8,9 @@ use Doctrine\Migrations\Metadata\ExecutedMigrationsList;
 use Doctrine\Migrations\Query\Query;
 use Doctrine\Migrations\Version\ExecutionResult;
 
+/**
+ * @method iterable<Query> getSql(ExecutionResult $result);
+ */
 interface MetadataStorage
 {
     public function ensureInitialized(): void;
@@ -17,9 +20,4 @@ interface MetadataStorage
     public function complete(ExecutionResult $result): void;
 
     public function reset(): void;
-
-    /**
-     * @return Query[]
-     */
-    public function getSql(ExecutionResult $result): iterable;
 }
--- a/lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
+++ b/lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
@@ -13,7 +13,7 @@ interface MetadataStorage
 
     public function getExecutedMigrations(): ExecutedMigrationsList;
 
-    public function complete(ExecutionResult $migration): void;
+    public function complete(ExecutionResult $result, bool $dry_run = false) : array;
 
     public function reset(): void;
 }
--- a/lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
+++ b/lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
@@ -125,35 +125,36 @@ final class TableMetadataStorage implements MetadataStorage
         );
     }
 
-    public function complete(ExecutionResult $result, bool $dry_run = false): array
+    /**
+     * {@inheritDoc}
+     */
+    public function complete(ExecutionResult $result, bool $dryRun = false): array
     {
-        $sql_queries = [];
-        if ($dry_run) {
-            $sql_queries[] = new Query('-- Version ' . (string) $result->getVersion() . ' update table metadata');
+        $sql = [];
+        if ($dryRun) {
+            $sql[] = new Query('-- Version ' . (string) $result->getVersion() . ' update table metadata');
             if ($result->getDirection() === Direction::DOWN) {
-                $delete_query = 'DELETE FROM ' . $this->configuration->getTableName() . ' WHERE ';
-                $delete_query .= $this->configuration->getVersionColumnName() . ' = ' . $this->connection->quote((string) $result->getVersion());
-
-                $sql_queries[] = new Query($delete_query);
+                $query  = 'DELETE FROM ' . $this->configuration->getTableName() . ' WHERE ';
+                $query .= $this->configuration->getVersionColumnName() . ' = ' . $this->connection->quote((string) $result->getVersion());
             } else {
-                $insert_query = 'INSERT INTO ' . $this->configuration->getTableName();
-                $insert_query .= ' (' . $this->configuration->getVersionColumnName() . ', ' . $this->configuration->getExecutedAtColumnName() . ', ' . $this->configuration->getExecutionTimeColumnName() . ')';
-                $insert_query .= ' VALUES (' . $this->connection->quote((string) $result->getVersion()) . ', NOW(), 0)';
-
-                $sql_queries[] = new Query($insert_query);
+                $query  = 'INSERT INTO ' . $this->configuration->getTableName();
+                $query .= ' (' . $this->configuration->getVersionColumnName() . ', ' . $this->configuration->getExecutedAtColumnName() . ', ' . $this->configuration->getExecutionTimeColumnName() . ')';
+                $query .= ' VALUES (' . $this->connection->quote((string) $result->getVersion()) . ', NOW(), 0)';
             }
+
+            $sql[] = new Query($query);
         } else {
             $this->checkInitialization();
 
             if ($result->getDirection() === Direction::DOWN) {
                 $this->connection->delete($this->configuration->getTableName(), [
-                    $this->configuration->getVersionColumnName() => (string)$result->getVersion(),
+                    $this->configuration->getVersionColumnName() => (string) $result->getVersion(),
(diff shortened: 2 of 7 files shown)


lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
```diff
@@ -124,25 +125,45 @@ public function reset(): void
         );
     }
 
-    public function complete(ExecutionResult $result): void
+    /**
+     * {@inheritDoc}
+     */
+    public function complete(ExecutionResult $result, bool $dryRun = false): array
```
greg0ire: Having boolean arguments is a known code smell, it often results in the method doing too many things. Here we can see how it makes the complexity skyrocket. I think you should create a new method.

lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
```diff
@@ -124,25 +125,45 @@ public function reset(): void
...
+        $sql = [];
+        if ($dryRun) {
+            $sql[] = new Query('-- Version ' . (string) $result->getVersion() . ' update table metadata');
+            if ($result->getDirection() === Direction::DOWN) {
+                $query  = 'DELETE FROM ' . $this->configuration->getTableName() . ' WHERE ';
+                $query .= $this->configuration->getVersionColumnName() . ' = ' . $this->connection->quote((string) $result->getVersion());
+            } else {
+                $query  = 'INSERT INTO ' . $this->configuration->getTableName();
+                $query .= ' (' . $this->configuration->getVersionColumnName() . ', ' . $this->configuration->getExecutedAtColumnName() . ', ' . $this->configuration->getExecutionTimeColumnName() . ')';
+                $query .= ' VALUES (' . $this->connection->quote((string) $result->getVersion()) . ', NOW(), 0)';
```
greg0ire: Please use `sprintf` (with a nowdoc string for the `INSERT` query) to build these queries.

lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
wiistriker: totally agree. new method MetadataStorage::getSql() added

lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
```diff
@@ -13,7 +14,12 @@ public function ensureInitialized(): void;
...
 
-    public function complete(ExecutionResult $migration): void;
+    public function complete(ExecutionResult $result): void;
 
     public function reset(): void;
+
+    /**
+     * @return Query[]
+     */
+    public function getSql(ExecutionResult $result): array;
```
greg0ire: Adding a method to an interface is a BC break for implementors outside the package. You can use an `@method` annotation to indicate that the method _should_ be present, it will be taken into account by static analysis tools I think.

greg0ire: Also, you don't need the return value to be an `array`, iterable is enough.

lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorage.php
```diff
@@ -145,6 +146,36 @@ public function complete(ExecutionResult $result): void
...
+                $this->connection->quote((string) $result->getVersion()),
+                $this->connection->quote(($result->getExecutedAt() ?? new DateTimeImmutable())->format('Y-m-d H:i:s'))
+            );
+        }
+
+        $sql[] = new Query($query);
+
+        return $sql;
+    }
+
```
greg0ire:
```suggestion
    public function getSql(ExecutionResult $result): iterable
    {
        yield new Query('-- Version ' . (string) $result->getVersion() . ' update table metadata');
        if ($result->getDirection() === Direction::DOWN) {
            yield new Query(sprintf(
                'DELETE FROM %s WHERE %s = %s',
                $this->configuration->getTableName(),
                $this->configuration->getVersionColumnName(),
                $this->connection->quote((string) $result->getVersion())
            );

            return;
        }
        yield new Query(sprintf(
            'INSERT INTO %s (%s, %s, %s) VALUES (%s, %s, 0)',
            $this->configuration->getTableName(),
            $this->configuration->getVersionColumnName(),
            $this->configuration->getExecutedAtColumnName(),
            $this->configuration->getExecutionTimeColumnName(),
            $this->connection->quote((string) $result->getVersion()),
            $this->connection->quote(($result->getExecutedAt() ?? new DateTimeImmutable())->format('Y-m-d H:i:s'))
        ));
    }
```

tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
```diff
@@ -355,4 +355,42 @@ public function testResetWithEmptySchema(): void
...
+    {
+        $this->storage->ensureInitialized();
+
+        $result = new ExecutionResult(new Version('2230'), Direction::UP, new DateTimeImmutable('2010-01-05 10:30:21'));
+
+        $queries = $this->storage->getSql($result);
+
+        self::assertCount(2, $queries);
+        self::assertSame('-- Version 2230 update table metadata', $queries[0]->getStatement());
+        self::assertSame('INSERT INTO doctrine_migration_versions (version, executed_at, execution_time) VALUES (\'2230\', \'2010-01-05 10:30:21\', 0)', $queries[1]->getStatement());
```
greg0ire: You can avoid the escaping by using double quotes. Also, this line is too long, please use linebreaks. Same below.

tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
```diff
@@ -355,4 +355,42 @@ public function testResetWithEmptySchema(): void
...
+
+        self::assertCount(2, $queries);
+        self::assertSame('-- Version 2230 update table metadata', $queries[0]->getStatement());
+        self::assertSame('DELETE FROM doctrine_migration_versions WHERE version = \'2230\'', $queries[1]->getStatement());
+
+        foreach ($queries as $query) {
+            $this->connection->executeStatement($query);
+        }
+
+        self::assertCount(0, $this->connection->fetchAllAssociative($sql));
```
greg0ire, line 401: This is quite repetitive, please use a data provider to ensure both segments do use the same logic.

lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
wiistriker: > Adding a method to an interface for implementors outside the package. You can use an `@method` annotation to indicate that the method _should_ be present, it will be taken into account by static analysis tools I think.

what do you mean by "_should_ be present". This method always should be present in implementation because this method exists in Interface

I think about getSql() method and maybe we should introduce new interface? Like MetadataSQLStorage and move getSql method there? So in DbalExecutor we will call this method only if our Storage implement this interface. What do you think?

tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
wiistriker: Do you mean phpunit @dataProvider annotation? If so i disagree.

Queries in this test depends on each other: DOWN doesnt make sense without UP. I test exactly this situation - generate and execute UP queries, check that migration was inserted in metadata table and after that execute DOWN queries and check that migration was deleted from metadata table.

wiistriker: similar behaviour presented in other test - https://github.com/doctrine/migrations/blob/3.5.x/tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php#L339

lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
greg0ire: >      what do you mean by "should be present". 

I mean "should be present"… if the implementor wants to become compatible with the next major release of `doctrine/migrations`, where the method should be directly in the interface.

> I think about getSql() method and maybe we should introduce new interface? Like MetadataSQLStorage and move getSql method there? So in DbalExecutor we will call this method only if our Storage implement this interface. What do you think?

I think you should use `method_exists`, and then, on branch 3.6, trigger a deprecation if is not implemented. The deprecation should state that `getSql()` will become part of the interface in 4.0.0

tests/Doctrine/Migrations/Tests/Metadata/Storage/TableMetadataStorageTest.php
greg0ire: Ah I see, that makes sense.

lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
```diff
@@ -13,7 +14,12 @@ public function ensureInitialized(): void;
 
     public function getExecutedMigrations(): ExecutedMigrationsList;
 
-    public function complete(ExecutionResult $migration): void;
+    public function complete(ExecutionResult $result): void;
 
     public function reset(): void;
+
+    /**
+     * @return Query[]
```
greg0ire:
```suggestion
     * @return iterable<Query>
```

lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
wiistriker: I commit necessary changes. Hope i understand you correct

lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
```diff
@@ -5,15 +5,19 @@
...
+/**
+ * @method iterable<Query> getSql(ExecutionResult $result);
+ */
 interface MetadataStorage
 {
     public function ensureInitialized(): void;
 
     public function getExecutedMigrations(): ExecutedMigrationsList;
 
-    public function complete(ExecutionResult $migration): void;
```
wiistriker, line 16: Typo
in TableMetadataStorage class this variable named $result.

getting bc check errors

lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
stof: The first comment was incomplete:

```diff
-Adding a method to an interface for implementors outside the package.
+Adding a method to an interface is a BC break for implementors outside the package.
```

wiistriker: > The first comment was incomplete:
> 
> ```diff
> -Adding a method to an interface for implementors outside the package.
> +Adding a method to an interface is a BC break for implementors outside the package.
> ```

ok, i understand. already made necessary changes.

but i cant figure out last error in phpstan:

```
------ --------------------------------------------------------------------------------------------------------------------------------------- 
  Line   lib/Doctrine/Migrations/Version/DbalExecutor.php                                                                                       
 ------ --------------------------------------------------------------------------------------------------------------------------------------- 
  207    Call to function method_exists() with Doctrine\Migrations\Metadata\Storage\MetadataStorage and 'getSql' will always evaluate to true.  
 ------ --------------------------------------------------------------------------------------------------------------------------------------- 
```

Should i add this error to phpstan ignoreErrors ?

wiistriker:
any news on this?

lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
greg0ire: That's because it's a breaking change when you use named parameters, I think you can ignore that failure.

lib/Doctrine/Migrations/Metadata/Storage/MetadataStorage.php
greg0ire: Yes, please do.

wiistriker: done

greg0ire (approved):

wiistriker:
any news on this?

greg0ire:
No reviews… I should have some time to merge and release this this afternoon.

wiistriker:
Is there any options to have this changes in 3.0 version?

greg0ire:
> Is there any options to have this changes in 3.0 version?

Nope, sorry.

PR #1273, opened 2022-09-15
Leverage LazyProxyTrait in LazySchemaDiffProvider

nicolas-grekas:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no (LazySchemaDiffProvider is internal)
| Fixed issues | -

This PR proposes to leverage the new `LazyProxyTrait` that's going to be shipped with Symfony 6.2 to implement lazy schemas as required by LazySchemaDiffProvider. (See https://symfony.com/blog/revisiting-lazy-loading-proxies-in-php for an introduction to this trait.)

This would require bumping the minimum PHP version to 8.1 in the next minor version.
The main benefit is simplicity. As you can read in the patch, using this trait requires no proxy generation infrastructure, and no calls to `eval()` either. The other benefit is simplifying dependency management.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,15 +24,15 @@
         }
     ],
     "require": {
-        "php": "^7.4 || ^8.0",
+        "php": "^8.1",
         "composer-runtime-api": "^2",
         "doctrine/dbal": "^3.4.3",
         "doctrine/deprecations": "^0.5.3 || ^1",
         "doctrine/event-manager": "^1.0",
-        "friendsofphp/proxy-manager-lts": "^1.0",
         "psr/log": "^1.1.3 || ^2 || ^3",
         "symfony/console": "^4.4.16 || ^5.4 || ^6.0",
-        "symfony/stopwatch": "^4.4 || ^5.4 || ^6.0"
+        "symfony/stopwatch": "^4.4 || ^5.4 || ^6.0",
+        "symfony/var-exporter": "^6.2"
     },
     "conflict": {
         "doctrine/orm": "<2.12"
@@ -49,7 +49,7 @@
         "phpstan/phpstan-phpunit": "^1.1",
         "phpstan/phpstan-strict-rules": "^1.1",
         "phpstan/phpstan-symfony": "^1.1",
-        "phpunit/phpunit": "^9.5",
+        "phpunit/phpunit": "^9.5.24",
         "symfony/cache": "^4.4 || ^5.4 || ^6.0",
         "symfony/process": "^4.4 || ^5.4 || ^6.0",
         "symfony/yaml": "^4.4 || ^5.4 || ^6.0"
--- a/lib/Doctrine/Migrations/Provider/LazySchemaDiffProvider.php
+++ b/lib/Doctrine/Migrations/Provider/LazySchemaDiffProvider.php
@@ -5,10 +5,6 @@ declare(strict_types=1);
 namespace Doctrine\Migrations\Provider;
 
 use Doctrine\DBAL\Schema\Schema;
-use ProxyManager\Configuration;
-use ProxyManager\Factory\LazyLoadingValueHolderFactory;
-use ProxyManager\GeneratorStrategy\EvaluatingGeneratorStrategy;
-use ProxyManager\Proxy\LazyLoadingInterface;
 
 /**
  * The LazySchemaDiffProvider is responsible for lazily generating the from schema when diffing two schemas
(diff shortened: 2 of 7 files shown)


nicolas-grekas:
phpstan report is only about false positives:

> Property Doctrine\Migrations\Provider\LazySchema::$lazyObject* is unused

Yes it is: by the trait.
(And no, the trait cannot declare those properties because their type/modifiers can vary in classes using it.)

> 

> Result of && is always true.

No it's not: undefined properties are not in objects cast to arrays.

derrabus:
Add those errors to the `ignoreErrors` section in phpstan.neon.dist with a comment explaining why you want to ignore them. If you consider the errors PHPStan bugs and want to go the extra mile, report them to PHPStan and add a link to your issue as a comment to phpstan.neon.dist.

stof:
Merging this is blocked on the stable release of Symfony 6.2 though

stof:
Looking at https://packagist.org/packages/doctrine/migrations/php-stats#3, PHP 8.1 only represents 45% of the installations for doctrine/migrations 3.x. The bump might be too early (the wait for the stable release of Symfony 6.2 gives us 2.5 months before re-evaluating that ratio anyway)

nicolas-grekas:
phpstan green, thanks for the hint @derrabus 
The remaining failures were already there so :shrug: 

@stof indeed, but Symfony 6.2 is pretty close (end of November) so we can start considering this now. v3.5 will still be usable for ppl on php < 8.1. The bump is worth it IMHO.

derrabus:
My attempt at getting the CI green: doctrine/migrations#1274

.github/workflows/coding-standards.yml
```diff
@@ -13,3 +13,5 @@ jobs:
   coding-standards:
     name: "Coding Standards"
     uses: "doctrine/.github/.github/workflows/coding-standards.yml@1.4.1"
+    with:
+      php-version: 8.1
```
derrabus: You can instead bump to `coding-standards.yml@2.0.0`. That version defaults to PHP 8.1.

nicolas-grekas: thanks, all fixed

stof:
> > Property Doctrine\Migrations\Provider\LazySchema::$lazyObject* is unused
> 
> Yes it is: by the trait. (And no, the trait cannot declare those properties because their type/modifiers can vary in classes using it.)

@nicolas-grekas This is true for `lazyObjectReal`. But AFAICT, `lazyObjectId` is always an int. So couldn't we declare it in the trait ?

nicolas-grekas:
> lazyObjectId is always an int. So couldn't we declare it in the trait ?

I could not add it because it would make the trait incompatible with readonly classes (there's a test case about that in the component)

stof:
> I could not add it because it would make the trait incompatible with readonly classes (there's a test case about that in the component)

if the trait supports making the property readonly, maybe it should always have it readonly ? Non-readonly classes can have readonly properties.

nicolas-grekas:
Making lazyObjectId readonly breaks cloning... I opened a discussion on php-internals about this and almost nobody dared to care for now... I hope this could be resolved soon...

derrabus:
Can you please rebase? The CI on 3.6.x should be green now.

nicolas-grekas:
Rebased. I don't reproduce the deps=low failure locally, I need to understand why it happens. I thought it might be because of https://github.com/composer/composer/issues/11068 but apparently not.

nicolas-grekas:
I figured out the issue, all green.

composer.json
```diff
@@ -24,15 +24,15 @@
...
         "composer-runtime-api": "^2",
         "doctrine/dbal": "^3.4.3",
         "doctrine/deprecations": "^0.5.3 || ^1",
         "doctrine/event-manager": "^1.0",
-        "friendsofphp/proxy-manager-lts": "^1.0",
         "psr/log": "^1.1.3 || ^2 || ^3",
         "symfony/console": "^4.4.16 || ^5.4 || ^6.0",
-        "symfony/stopwatch": "^4.4 || ^5.4 || ^6.0"
+        "symfony/stopwatch": "^4.4 || ^5.4 || ^6.0",
+        "symfony/var-exporter": "^6.2@dev"
```
kbond:
```suggestion
        "symfony/var-exporter": "^6.2"
```

🎉

nicolas-grekas: :rocket:

lib/Doctrine/Migrations/Provider/LazySchemaDiffProvider.php
```diff
@@ -18,59 +14,27 @@
...
 
     public function __construct(
-        LazyLoadingValueHolderFactory $proxyFactory,
         SchemaDiffProvider $originalSchemaManipulator
     ) {
-        $this->proxyFactory              = $proxyFactory;
         $this->originalSchemaManipulator = $originalSchemaManipulator;
     }
 
-    public static function fromDefaultProxyFactoryConfiguration(
```
stof, line 33: this method should probably be deprecated instead of being removed

nicolas-grekas:
Ready for merge!
(failures are unrelated IIUC)

:rocket:

lib/Doctrine/Migrations/Provider/LazySchemaDiffProvider.php
nicolas-grekas: the class is marked `@internal` so I didn't bother doing that :)

stof: ah indeed. I missed that

greg0ire:
Just did a merge-up, it should fix the SA job ([the BC check will still be broken](https://github.com/Nyholm/roave-bc-check-docker/issues/32)).

greg0ire (approved):
A lot less code :)

derrabus (approved):

bobvandevijver:
@nicolas-grekas I'm just seeing this update, but due to the requirement on `symfony/var-exporter` on `^6.2` combined with the flex functionality to limit the Symfony packages to `5.4.*` (`symfony.require` in the composer file), the latest release isn't installable. Would that be an oversight or was it intentional to require SF 6.2?

stof:
older versions of var-exporter don't have that LazyProxyTrait

nicolas-grekas:
Thanks for the report. This should now be fixed. I did so by reporting that var-exporter didn't existing before 6.2, see
https://github.com/symfony-tools/recipes-checker/commit/a50caac0e81d7d6405c886578f76f29013034f49

This should allow installing var-exporter at v6.2 while the rest is at 5.4.
This is also important to help Doctrine ORM adopt lazy-proxies faster.
/cc @derrabus since we talked about this issue recently.

derrabus:
Thank you! That should help indeed.

bobvandevijver:
This fixed it indeed, thanks!

chrisguitarguy:
One thing to note here is that the new trait here isn't in Symfony 5.4, so anyone on an LTS release of Symfony (like me!) won't be able to upgraded to migrations 3.6

PR #1274, opened 2022-09-15
Fix tests and static analysis

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

This PR attempts to fix all current CI failures.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "require": {
         "php": "^7.4 || ^8.0",
         "composer-runtime-api": "^2",
-        "doctrine/dbal": "^3.3",
+        "doctrine/dbal": "^3.4.3",
         "doctrine/deprecations": "^0.5.3 || ^1",
         "doctrine/event-manager": "^1.0",
         "friendsofphp/proxy-manager-lts": "^1.0",
--- a/lib/Doctrine/Migrations/Configuration/Migration/XmlFile.php
+++ b/lib/Doctrine/Migrations/Configuration/Migration/XmlFile.php
@@ -87,16 +87,24 @@ final class XmlFile extends ConfigurationFile
             } elseif ($nodeName === 'storage' && $node->{'table-storage'} instanceof SimpleXMLElement) {
                 $config['table_storage'] = $this->extractParameters($node->{'table-storage'}, false);
             } elseif ($nodeName === 'migrations') {
-                $config['migrations'] = [];
-                foreach ($node->{'migration'} as $pathNode) {
-                    $config['migrations'][] = (string) $pathNode;
-                }
+                $config['migrations'] = $this->extractMigrations($node);
             } else {
                 $config[$nodeName] = (string) $node;
             }
         }
 
         return $config;
+    }
+
+    /** @return list<string> */
+    private function extractMigrations(SimpleXMLElement $node): array
+    {
+        $migrations = [];
+        foreach ($node->{'migration'} as $pathNode) {
+            $migrations[] = (string) $pathNode;
+        }
+
+        return $migrations;
     }
 
     private function validateXml(string $file): void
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -26,7 +26,7 @@
     "require": {
         "php": "^7.4 || ^8.0",
         "composer-runtime-api": "^2",
-        "doctrine/dbal": "^3.3",
+        "doctrine/dbal": "^3.4",
```
derrabus: I hope this bump to DBAL 3.4.3 is okay for a bugfix release. If not, I could revert here and skip the three failing tests on the low-deps run.

lib/Doctrine/Migrations/Configuration/Migration/XmlFile.php
```diff
@@ -87,10 +87,7 @@ private function extractParameters(SimpleXMLElement $root, bool $loopOverNodes):
             } elseif ($nodeName === 'storage' && $node->{'table-storage'} instanceof SimpleXMLElement) {
                 $config['table_storage'] = $this->extractParameters($node->{'table-storage'}, false);
             } elseif ($nodeName === 'migrations') {
-                $config['migrations'] = [];
-                foreach ($node->{'migration'} as $pathNode) {
-                    $config['migrations'][] = (string) $pathNode;
-                }
+                $config['migrations'] = $this->extractMigrations($node);
```
derrabus, line 90: Workaround for phpstan/phpstan#8015

tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
```diff
@@ -422,7 +422,7 @@ public function testExecuteMigrateCancelExecutedUnavailableMigrations(): void
         $output = $this->migrateCommandTester->getDisplay(true);
 
         self::assertStringContainsString('[WARNING] You have 1 previously executed migrations in the database that are not registered migrations.', $output);
-        self::assertStringContainsString('WARNING! You are about to execute a migration in database "<unnamed>" that could result in schema changes and data loss. Are you sure you wish to continue?', $output);
+        self::assertStringContainsString('WARNING! You are about to execute a migration in database "main" that could result in schema changes and data loss. Are you sure you wish to continue?', $output);
```
derrabus, line 425: On DBAL 3.4, the default schema on SQLite is `main` which is more correct than the empty string that was reported previously.

derrabus (commented):
💡 Let me explain my changes quickly.

composer.json
stof: if the supported PHP versions have not changed between DBAL 3.3 and 3.4 (or if the dropped PHP versions were already not supported here), it is fine IMO.

goetas (approved):

stof (approved):

composer.json
stof: And the answer to that is that DBAL 3.4 bumped the min PHP version from 7.3 to 7.4, which is fine as doctrine/migrations already requires PHP 7.4 anyway.

phpstan.neon.dist
```diff
@@ -24,6 +24,11 @@ parameters:
             message: '~^Method Doctrine\\Migrations\\Tests\\Stub\\DoctrineRegistry::getService\(\) should return Doctrine\\Persistence\\ObjectManager but returns Doctrine\\DBAL\\Connection\|Doctrine\\ORM\\EntityManager~'
             path: tests/Doctrine/Migrations/Tests/Stub/DoctrineRegistry.php
 
+        # We cannot document the closure as precise as PHPStan wants us to.
```
greg0ire, line 27: Why not? I'm on my phone and cannot easily check whether it is possible to document a closure, but it seems possible with`@var`? Or maybe with an inline comment?

derrabus: If you find a nice way to do so, be my guest. The closure that we're passing here looks fine from my POV, so I'm not sure if this error is really worth fixing.

PR #1282, opened 2022-10-21
doc(generating-migration): change ignore table function

Amorfx:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

The setFilterSchemaAssetsExpression function no longer exists. Here is a small fix the way I found in order to be able to ignore tables without going through the parameter in the command but directly in the configuration.

### DIFF
--- a/docs/en/reference/generating-migrations.rst
+++ b/docs/en/reference/generating-migrations.rst
@@ -260,7 +260,7 @@ with a schema filter.
 
 .. code-block:: php
 
-    $connection->getConfiguration()->setSchemaAssetsFilter(static function ($assetName) {
+    $connection->getConfiguration()->setSchemaAssetsFilter(static function (string|AbstractAsset $assetName): bool {
         if ($assetName instanceof AbstractAsset) {
             $assetName = $assetName->getName();
         }
--- a/docs/en/reference/generating-migrations.rst
+++ b/docs/en/reference/generating-migrations.rst
@@ -260,7 +260,13 @@ with a schema filter.
 
 .. code-block:: php
 
-    $connection->getConfiguration()->setFilterSchemaAssetsExpression("~^(?!t_)~");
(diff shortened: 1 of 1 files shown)


docs/en/reference/generating-migrations.rst
```diff
@@ -260,7 +260,13 @@ with a schema filter.
 
 .. code-block:: php
 
-    $connection->getConfiguration()->setFilterSchemaAssetsExpression("~^(?!t_)~");
+    $connection->getConfiguration()->setSchemaAssetsFilter(static function ($assetName) {
```
greg0ire:
```suggestion
    $connection->getConfiguration()->setSchemaAssetsFilter(static function (string|AbstractAsset $assetName): bool {
```

Amorfx: Thanks i forgot that !

docs/en/reference/generating-migrations.rst
```diff
@@ -260,7 +260,13 @@ with a schema filter.
 
 .. code-block:: php
 
-    $connection->getConfiguration()->setFilterSchemaAssetsExpression("~^(?!t_)~");
+    $connection->getConfiguration()->setSchemaAssetsFilter(static function (string|AbstractAsset $assetName): bool {
+        if ($assetName instanceof AbstractAsset) {
+            $assetName = $assetName->getName();
+        }
+
+        return preg_match("~^(?!t_)~", $assetName);
```
derrabus: `preg_match()` returns an integer. I believe you'll have to cast it in order to make this example work with strict types enabled.

```suggestion
        return (bool) preg_match("~^(?!t_)~", $assetName);
```

SenseException (approved):
Referring to the API of another library in the docs is a pain for keeping docs up to date. Thanks for that fix.

greg0ire:
Thanks @Amorfx !

PR #1284, opened 2022-11-06
fix(docs): organize_migrations parameters values

lugosium:
|      Q       |   A
|------------- | -----------
| Type         | docs
| BC Break     | no
| Fixed issues | none

#### Summary

👋 

Values provided for the `organize_migrations` configuration parameter are invalid, here's the good ones.


lugosium:
My bad, i was not reading the doc for the bundle.

PR #1285, opened 2022-11-07
Don't throw NoMappingFound from OrmSchemaProvider

nicolas-grekas:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | -

This is something we figured out in Symfony: when there is no mapping info available, it doesn't mean there is no migration to generate. This happens when there are schema listeners that integrate with migrations, as is the case for those:
https://github.com/symfony/symfony/tree/6.2/src/Symfony/Bridge/Doctrine/SchemaListener

Right now, one has to have entities before they can generate migrations, but those schema are entity-less.

Removing this check does the job.

### DIFF
--- a/tests/Doctrine/Migrations/Tests/Provider/OrmSchemaProviderTest.php
+++ b/tests/Doctrine/Migrations/Tests/Provider/OrmSchemaProviderTest.php
@@ -12,7 +12,6 @@ use Doctrine\ORM\EntityManager;
 use Doctrine\ORM\EntityManagerInterface;
 use Doctrine\ORM\Mapping\Driver\XmlDriver;
 use Doctrine\ORM\ORMSetup;
-use UnexpectedValueException;
 
 /**
  * Tests the OrmSchemaProvider using a real entity manager.
@@ -37,10 +36,14 @@ class OrmSchemaProviderTest extends MigrationTestCase
         }
     }
 
-    public function testEntityManagerWithoutMetadataCausesError(): void
+    /**
+     * It should be OK to use migrations to manage tables not managed by
+     * the ORM.
+     *
+     * @doesNotPerformAssertions
+     */
+    public function testEntityManagerWithoutMetadata(): void
     {
-        $this->expectException(UnexpectedValueException::class);
-
         $this->config->setMetadataDriverImpl(new XmlDriver([]));
 
         $this->ormProvider->createSchema();
--- a/lib/Doctrine/Migrations/Provider/OrmSchemaProvider.php
+++ b/lib/Doctrine/Migrations/Provider/OrmSchemaProvider.php
@@ -5,12 +5,10 @@ declare(strict_types=1);
 namespace Doctrine\Migrations\Provider;
 
 use Doctrine\DBAL\Schema\Schema;
-use Doctrine\Migrations\Provider\Exception\NoMappingFound;
 use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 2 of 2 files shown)


nicolas-grekas:
(phpstan failures are unrelated)

greg0ire:
I fixed the PHPStan failures  in https://github.com/doctrine/migrations/pull/1286

derrabus:
Can you please rebase?

nicolas-grekas:
All green in a minute.

tests/Doctrine/Migrations/Tests/Provider/OrmSchemaProviderTest.php
```diff
@@ -37,13 +36,13 @@ public function testCreateSchemaFetchesMetadataFromEntityManager(): void
         }
     }
 
-    public function testEntityManagerWithoutMetadataCausesError(): void
+    public function testEntityManagerWithoutMetadata(): void
```
greg0ire, line 45: I think this warrants an explanation alongside the test.

```suggestion
    /**
     * It should be OK to use migrations to manage tables not managed by
     * the ORM.
     *
     * @doesNotPerformAssertions
     */
    public function testEntityManagerWithoutMetadataIsFine(): void
```

tests/Doctrine/Migrations/Tests/Provider/OrmSchemaProviderTest.php
```diff
@@ -37,13 +36,13 @@ public function testCreateSchemaFetchesMetadataFromEntityManager(): void
...
     {
-        $this->expectException(UnexpectedValueException::class);
-
         $this->config->setMetadataDriverImpl(new XmlDriver([]));
 
         $this->ormProvider->createSchema();
+
+        self::addToAssertionCount(1);
```
greg0ire: Let's be explicit about this and use `@doesNotPerformAssertions` instead?
```suggestion
```

tests/Doctrine/Migrations/Tests/Provider/OrmSchemaProviderTest.php
nicolas-grekas: updated, thanks

tests/Doctrine/Migrations/Tests/Provider/OrmSchemaProviderTest.php
```diff
@@ -37,10 +36,14 @@ public function testCreateSchemaFetchesMetadataFromEntityManager(): void
         }
     }
 
-    public function testEntityManagerWithoutMetadataCausesError(): void
+    /**
+     * It should be OK to use migrations to manage tables not managed by
+     * the ORM.
+     *
+     * @doesNotPerformAssertions
```
stof, line 43: The tests should probably perform some assertion

nicolas-grekas: I had an "assertInstanceOf" but phpstan complained that this was already enforced by the return type of the method... :shrug:

greg0ire (approved):

SenseException (approved):
As far as I understand this became an expected behaviour.

tests/Doctrine/Migrations/Tests/Provider/OrmSchemaProviderTest.php
derrabus: Do we expect the schema returned by `createSchema()` to be empty? Can we assert that?

PR #1286, opened 2022-11-07
Address deprecations from doctrine/dbal

greg0ire:


phpstan.neon.dist
```diff
@@ -53,6 +53,11 @@ parameters:
             message: '~^Call to function method_exists\(\) with Doctrine\\Migrations\\Metadata\\Storage\\MetadataStorage and ''getSql'' will always evaluate to true\.$~'
             path: lib/Doctrine/Migrations/Version/DbalExecutor.php
 
+        # TODO: deprecate using the connection event manager and expose
+        # our own event manager instead.
+	-   message: '~^Call to deprecated method getEventManager\(\) of class Doctrine\\DBAL\\Connection\.$~'
+            path: lib/Doctrine/Migrations/DependencyFactory.php
```
greg0ire, line 60: This one should be addressed on 3.6.x, because I think it will involve a deprecation.

stof: indeed. Using a separate EventManager will require consumers to change how they register their listener.

greg0ire: I created a separate issue to track this: https://github.com/doctrine/migrations/issues/1287

derrabus (approved):

derrabus:
Thank you!

PR #1290, opened 2022-11-18
Workflow maintenance

greg0ire:


.github/workflows/continuous-integration.yml
```diff
@@ -2,23 +2,25 @@ name: "Continuous Integration"
...
 jobs:
-  composer_normalize:
-    name: "Composer Normalize"
-    runs-on: "ubuntu-22.04"
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          fetch-depth: 0
-
-      - uses: "docker://ergebnis/composer-normalize-action:latest"
```
greg0ire, line 18: That image hasn't receive a push in 2 years, which might explain why the build is broken :P

composer.json
```diff
@@ -34,16 +34,12 @@
...
-    "conflict": {
-        "doctrine/orm": "<2.12"
-    },
     "require-dev": {
         "ext-pdo_sqlite": "*",
         "doctrine/coding-standard": "^9",
         "doctrine/orm": "^2.12",
         "doctrine/persistence": "^2 || ^3",
         "doctrine/sql-formatter": "^1.0",
-        "ergebnis/composer-normalize": "^2.9",
```
greg0ire, line 46: I removed that dependency and its configuration because I think no-one is going to break the composer.json so badly that they will need to fix it with that tool as opposed to manually. If they do, they can require the lib, run the tool, and remove it.

SenseException: Sounds reasonable. As I can see we didn't implement it as an dependency in other projects. Its impact wasn't that big for us.

.github/workflows/static-analysis.yml
```diff
@@ -5,9 +5,19 @@ on:
...
+      - lib/**
+      - phpstan*
   push:
     branches:
       - "*.x"
+    paths:
+      - .github/workflows/static-analysis.yml
+      - composer.*
+      - lib/**
+      - phpstan*
```
SenseException: Tests-directory is still checked in this repo,

SenseException (changes requested):

SenseException (approved):

PR #1293, opened 2022-12-12
Use correct default values for VersionColumnLength in doc

eigan:
See [TableMetadataStorageConfiguration](https://github.com/doctrine/migrations/blob/5713b45c933122e509d9b31c767b420c3dfed399/lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorageConfiguration.php)

Relevant
https://github.com/doctrine/migrations/pull/960 (from 1024 -> 191)

|      Q       |   A
|------------- | -----------
| Type         | bug (documentation)
| BC Break     | no
| Fixed issues | 

#### Summary

The documentation references 1024 as default version column length, but the actual default value is 191. https://github.com/doctrine/migrations/blob/5713b45c933122e509d9b31c767b420c3dfed399/lib/Doctrine/Migrations/Metadata/Storage/TableMetadataStorageConfiguration.php

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -27,7 +27,7 @@ Now, in the root of your project place a file named ``migrations.php``, ``migrat
             'table_storage' => [
                 'table_name' => 'doctrine_migration_versions',
                 'version_column_name' => 'version',
-                'version_column_length' => 1024,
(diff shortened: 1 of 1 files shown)


SenseException (approved):

SenseException:
Thank you @eigan

Issue #1279, opened 2022-10-06
`--all-or-nothing` option is very confusing


lazka:
|    Q        |   A
|------------ | -----
| Version     | 3.1.5

I was running the following:

```
bin/console doctrine:migrations:migrate --em=my_em --no-interaction --all-or-nothing
```

Turns out just passing `--all-or-nothing` does not do anything, you have to pass `--all-or-nothing=1` or `--all-or-nothing=0` for it to have any effect: https://github.com/doctrine/migrations/blob/5713b45c933122e509d9b31c767b420c3dfed399/lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php#L24 (it will result in null, then fall back to the configuration)

This is very confusing imo and there are lots of examples out there that just pass `--all-or-nothing`.

Am I missing something?

greg0ire:
What do the docs say?

agustingomes:
I created a PR to correct the docs example: https://github.com/doctrine/migrations/pull/1296 as I just found this issue while working on automating Doctrine migrations in my work context.

@greg0ire what do you think?

greg0ire:
I'll try to look into it but have you seen https://github.com/doctrine/migrations/pull/1161 ? Before it was merged, I think the `--all-or-nothing` must have been working as expected, because back then the code looked a lot like the code for the `dry-run` option (which works as expected, right?)

I think the code in that MR might be wrong. It should not fall back on the configuration as soon as `$input->hasOption('all-or-nothing')` returns `true`, right?

greg0ire:
Cc @grachevko

agustingomes:
Added a test case in the PR confirming the current (and unexpected) behavior.

PR #1296, opened 2023-01-09
Handle absence of value for `--all-or-nothing` properly

agustingomes:
### Expected behavior:
Upon specification of the option `--all-or-nothing` wraps multiple migrations into a transaction

### Observed behavior
Specifying `--all-or-nothing` by itself does not cause the desired effect, as the result of the statement `$input->getOption('all-or-nothing')` is `null`, and therefore, in the absence of config, this value is ultimately evaluated to `false`, therefore not wrapping multiple migrations into a single transaction.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/migrations/issues/1279

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
+++ b/lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
@@ -21,12 +21,26 @@ class ConsoleInputMigratorConfigurationFactory implements MigratorConfigurationF
     {
         $timeAllQueries = $input->hasOption('query-time') ? (bool) $input->getOption('query-time') : false;
         $dryRun         = $input->hasOption('dry-run') ? (bool) $input->getOption('dry-run') : false;
-        $allOrNothing   = $input->hasOption('all-or-nothing') ? $input->getOption('all-or-nothing') : null;
-        $allOrNothing   = (bool) ($allOrNothing ?? $this->configuration->isAllOrNothing());
+        $allOrNothing   = $this->determineAllOrNothingValueFrom($input) ?? $this->configuration->isAllOrNothing();
 
         return (new MigratorConfiguration())
             ->setDryRun($dryRun)
             ->setTimeAllQueries($timeAllQueries)
             ->setAllOrNothing($allOrNothing);
+    }
+
+    private function determineAllOrNothingValueFrom(InputInterface $input): ?bool
+    {
+        $allOrNothingOption = null;
+
+        if ($input->hasOption('all-or-nothing')) {
+            $allOrNothingOption = $input->getOption('all-or-nothing');
+        }
+
+        if ($allOrNothingOption === 'notprovided') {
+            return null;
+        }
+
+        return (bool) ($allOrNothingOption ?? true);
     }
 }
--- a/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
@@ -75,7 +75,8 @@ final class MigrateCommand extends DoctrineCommand
                 'all-or-nothing',
                 null,
                 InputOption::VALUE_OPTIONAL,
-                'Wrap the entire migration in a transaction.'
+                'Wrap the entire migration in a transaction.',
+                'notprovided'
             )
             ->setHelp(<<<EOT
 The <info>%command.name%</info> command executes a migration to a specified version or the latest available version:
--- a/tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
+++ b/tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
@@ -362,7 +362,7 @@ class MigrateCommandTest extends MigrationTestCase
     }
 
     /**
-     * @psalm-param array<string, bool> $input
+     * @psalm-param array<string, bool|int|string|null> $input
      *
      * @dataProvider allOrNothing
      */
@@ -390,13 +390,22 @@ class MigrateCommandTest extends MigrationTestCase
     }
 
     /**
-     * @psalm-return Generator<array{bool, array<string, bool>, bool}>
+     * @psalm-return Generator<array{bool, array<string, bool|int|string|null>, bool}>
      */
     public function allOrNothing(): Generator
     {
         yield [false, ['--all-or-nothing' => false], false];
+        yield [false, ['--all-or-nothing' => 0], false];
+        yield [false, ['--all-or-nothing' => '0'], false];
+
         yield [false, ['--all-or-nothing' => true], true];
+        yield [false, ['--all-or-nothing' => 1], true];
+        yield [false, ['--all-or-nothing' => '1'], true];
+        yield [false, ['--all-or-nothing' => null], true];
+
         yield [true, ['--all-or-nothing' => false], false];
+        yield [true, ['--all-or-nothing' => 0], false];
+        yield [true, ['--all-or-nothing' => '0'], false];
 
         yield [true, [], true];
         yield [false, [], false];


greg0ire:
As stated in #1279 today, this looks to me more like an unintended bug introduced in #1161. Can you take a look and let me know what you think?

agustingomes:
Hi @greg0ire 

I will try to add a failing test case if possible. I'll keep you posted.

agustingomes:
@greg0ire I added an extra permutation for the data provider of the `allOrNothing` test, which confirms that the behavior is not as expected.

What would you advice as next step?

greg0ire:
- revert the doc change
- fix the bug: `$allOrNothing   = $input->hasOption('all-or-nothing') ? $input->getOption('all-or-nothing') ?? true : null;` should IMO do the trick.

agustingomes:
@greg0ire unfortunately your suggestion alone does not seem to do the trick.

I've checked the [hasOption method in Symfony](https://github.com/symfony/symfony/blob/6.3/src/Symfony/Component/Console/Input/InputDefinition.php#L267) and it seems that method is not a reliable way to know if a user passed the command option or not.

In any case, I'll do a small refactor there to stop using `hasOption`.

greg0ire:
Indeed! I think by using a default value different than `null` (like `false`?) for `all-or-nothing`, you might be able to tell whether the option was used or not with `getOption`

agustingomes:
I'm not sure @greg0ire 

I'll give it some thought over the week, since it seems like relying on the default value of the option may not work, as it can override the Doctrine project configuration.

greg0ire:
I guess if you set the default to `false`, you would then have to translate `null` to `true` (option provided), and `false` to `null` (default value, option not provided), so that the line `$allOrNothing   = (bool) ($allOrNothing ?? $this->configuration->isAllOrNothing());` continues to work as intended. Bit convoluted, maybe you'll find something more elegant than this. And maybe we should use `VALUE_NONE` in 4.0, to simplify all this.

agustingomes:
> And maybe we should use VALUE_NONE in 4.0, to simplify all this.

That sounds like a good idea.

greg0ire:
You can use `vendor/bin/phpcbf` to fix the cs issues.

agustingomes:
@greg0ire how do you want to proceed regarding the PHPStan failures? THey seem to be originating from deprecations, and the BC check error I never seen it before.

lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
```diff
@@ -21,12 +22,26 @@ public function getMigratorConfiguration(InputInterface $input): MigratorConfigu
...
             ->setDryRun($dryRun)
             ->setTimeAllQueries($timeAllQueries)
             ->setAllOrNothing($allOrNothing);
     }
+
+    private function determineAllOrNothingValueFrom(InputInterface $input): ?bool
+    {
+        try {
+            $allOrNothingOption = $input->getOption('all-or-nothing');
+        } catch (InvalidArgumentException $exception) {
```
greg0ire: For which cases is this necessary? I have an uneasy feeling about this… if it's not strictly related to the issue at hand, I'd remove it.

greg0ire:
Nice commit messages. I've seen the test failing, so you can squash both commits now, so that all commits pass the build.

greg0ire:
> @greg0ire how do you want to proceed regarding the PHPStan failures? THey seem to be originating from deprecations, and the BC check error I never seen it before.

Don't worry about those, I'll take care of it.

Regarding the BC-check, it seems to have been broken for a while: https://github.com/Nyholm/roave-bc-check-docker/issues/32
Regarding the static analysis issues, I created https://github.com/doctrine/migrations/pull/1297

lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
agustingomes: There are some tests that fail if we don't catch it... I found it weird as well, but I would not know how to proceed further otherwise.

The tests that failed when not catching the exception:

- [testWriteSql](https://github.com/doctrine/migrations/blob/3.5.x/tests/Doctrine/Migrations/Tests/Tools/Console/Command/ExecuteCommandTest.php#L50)
- [testExecute](https://github.com/doctrine/migrations/blob/3.5.x/tests/Doctrine/Migrations/Tests/Tools/Console/Command/ExecuteCommandTest.php#L101)
- [testExecuteMultiple](https://github.com/doctrine/migrations/blob/3.5.x/tests/Doctrine/Migrations/Tests/Tools/Console/Command/ExecuteCommandTest.php#L123)
- [testExecuteCancel](https://github.com/doctrine/migrations/blob/3.5.x/tests/Doctrine/Migrations/Tests/Tools/Console/Command/ExecuteCommandTest.php#L155)

greg0ire: I think this class you're changing is called for several commands. IMO you should remove this `try/catch`, and resort to `hasOption` here, to determine if the command that is called has that option (as in, the option is defined as opposed to its value provided). If it doesn't, then I guess we can omit calling `setAllOrNothing`, or call it with `false`, whatever is the most simple for you.

agustingomes: The last push I did seems to be passing all tests. Using the `hasOption` did help.

lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
```diff
@@ -21,12 +21,26 @@ public function getMigratorConfiguration(InputInterface $input): MigratorConfigu
...
+
+        if ($input->hasOption('all-or-nothing')) {
+            $allOrNothingOption = $input->getOption('all-or-nothing');
+        }
+
+        if ($allOrNothingOption === 'notprovided') {
+            return null;
+        }
+
+        return (bool) ($allOrNothingOption ?? true);
```
greg0ire, line 44: If the command does not have the option, the value will default to `true`, but I guess it does not matter since `all-or-nothing` is not used when the option is not defined.

greg0ire (approved):

greg0ire:
Thanks @agustingomes !

greg0ire:
> > And maybe we should use VALUE_NONE in 4.0, to simplify all this.
> 
> That sounds like a good idea.

@agustingomes if you want to work on this, you can open a PR on 3.6.x to deprecate passing a value :)

Here is an example: https://github.com/doctrine/migrations/blob/01f89a1082b6971ba53ba76f95741fc8ad4a5fd2/lib/Doctrine/Migrations/Tools/TransactionHelper.php#L22-L32

agustingomes:
@greg0ire I'll look into it the next days 👍🏼

lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
e-jevdokimovs: This is a problem, the default value of `MigrationConfiguration` property `allOrNothing` is `false`, these changes always set the value to true if the option isn't passed at all AND the command doesn't have an `all-or-nothing` option, for example `ExecuteCommand`.

greg0ire: You mean the option is used although it's not defined in the command? Where?

e-jevdokimovs: The option itself isn't used, but [ConsoleInputMigratorConfigurationFactory->getMigratorConfiguration()](https://github.com/doctrine/migrations/blob/3.5.x/lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php#L20) is being called in [`ExecuteCommand->execute()`](https://github.com/doctrine/migrations/blob/3.5.x/lib/Doctrine/Migrations/Tools/Console/Command/ExecuteCommand.php#L115) which in turn creates a `MigratorConfiguration` with `allOrNothing` set to `true` because `ExecuteCommand` doesn't have an `all-or-nothing` option which results in the value not defaulting to `notprovided`.

e-jevdokimovs: Basically this method should look like this:
```
        if (!$input->hasOption('all-or-nothing')) {
            return null;
        }

        $allOrNothingOption = $input->getOption('all-or-nothing');

        if ($allOrNothingOption === 'notprovided') {
            return null;
        }

        return (bool) ($allOrNothingOption ?? true);
```

e-jevdokimovs: Created a PR with the fix: https://github.com/doctrine/migrations/pull/1308

PR #1298, opened 2023-01-10
Adds success message for migrations command

cpats007:
- Added success message before new line for successful migrations
- updated unit tests

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

When executing a successful migration, the only output from the command when not using verbosity levels is a blank new line. When running from something like ansible, it's sometimes nice to see an output. This PR adds a single line output to denote a successful migration to version XXX

Examples:

Previously you would need to run the migration twice to ensure that the update was successful:

image

This is how it will look on first run with the new change:

image

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -27,7 +27,7 @@ Now, in the root of your project place a file named ``migrations.php``, ``migrat
             'table_storage' => [
                 'table_name' => 'doctrine_migration_versions',
                 'version_column_name' => 'version',
-                'version_column_length' => 1024,
+                'version_column_length' => 191,
                 'executed_at_column_name' => 'executed_at',
                 'execution_time_column_name' => 'execution_time',
             ],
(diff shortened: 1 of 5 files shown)


greg0ire:
Needs a rebase because of #1297

greg0ire:
Also, a before/after screenshot would be nice.

greg0ire:
Oh and this should target 3.6.x, since it's an improvement.

cpats007:
> Oh and this should target 3.6.x, since it's an improvement.

Updated the target branch as the change will be the same in that branch also - no conflicts or additional code to alter that I can see and also added some screenshots

cpats007:
> Needs a rebase because of #1297

not sure what needs to be done for this?

greg0ire:
? Why did you close?

cpats007:
Having issues locally - was easier to close and create a new PR from the 3.6.x branch - will be incoming

PR #1300, opened 2023-01-11
Drop support for silencing errors about DDL in transactions

greg0ire:
I hesitated between this and removing the transaction helper entirely, but I think this solution will result in less support for us. Let me know if you think I should remove the transaction helper.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,3 +1,8 @@
+# Upgrade to 4.0
+
+It is no longer possible to relying on silencing to enable transactional
+migrations when the platform does not allow DDL inside transactions.
+
 # Upgrade to 3.1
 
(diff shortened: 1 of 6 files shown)


SenseException:
If the helper class is kept in 4.0, it can only be removed in a 5.0 release, right? It doesn't seem to be widely in use now.

greg0ire:
I think the choice is between removing it in 4.0, or keeping it forever, for DX reasons.

derrabus:
The method naming is odd now. The names end with `IfInTransaction` which suggests that I can also safely call the method without an active transaction. But that's not the case anymore.

greg0ire:
Ok, I'll switch to just `commit` and `rollback` then.

PR #1303, opened 2023-01-11
Added success message for Command Output

cpats007:
- Added success message before new line for successful migrations
- updated unit tests

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | none

#### Summary

When executing a successful migration, the only output from the command when not using verbosity levels is a blank new line. When running from something like ansible, it's sometimes nice to see an output. This PR adds a single line output to denote a successful migration to version XXX

Examples:

Previously you would need to run the migration twice to ensure that the update was successful:

image

This is how it will look on first run with the new change:

image

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php
@@ -215,6 +215,10 @@ EOT
             $writer->write($path, $plan->getDirection(), $sql);
         }
 
+        $this->io->success(sprintf(
+            'Successfully migrated to version : %s',
+            $version
+        ));
         $this->io->newLine();
 
         return 0;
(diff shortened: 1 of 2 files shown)


greg0ire (approved):

SenseException (approved):

SenseException:
Failed BC check looks unrelated.

greg0ire:
Thanks @cpats007 !

cpats007:
> Thanks @cpats007 !

No problem, thanks for merging :)

ToshY:
As pedantic as this may sound, a colon is **never** preceded by a white space in the **English** language.

greg0ire:
At the risk of sounding even more pedantic, you should _never_ use bold if you intend the text you write to be read linearly.

ToshY:
> At the risk of sounding even more pedantic, you should _never_ use bold if you intend the text you write to be read linearly.

Duly noted.

Back to the issue at hand: space preceding a colon. I assume there are no objections to fixing it then.

greg0ire:
No objections. If you want to send a PR, you might want to run `grep -rn -F ' :'` to look for other violations.

Issue #1304, opened 2023-01-15
Deprecate passing arguments for `--all-or-nothing` command option


agustingomes:
During the resolution of a bug with the `--all-or-nothing` command option, it was suggested to deprecate passing values with the option, and as consequence, treat the presence of the option as `true` from version `4.0.x` onwards. 

[reference](https://github.com/doctrine/migrations/pull/1296#issuecomment-1376282500)

agustingomes:
@greg0ire since the related PR was merged, should this issue be closed?

greg0ire:
Yes, thanks again for this!

PR #1305, opened 2023-01-15
Deprecate using `--all-or-nothing` with values

agustingomes:
|      Q       |   A
|------------- | -----------
| Type         | deprecation
| BC Break     | yes, in the major version 4
| Fixed issues | https://github.com/doctrine/migrations/issues/1304

#### Summary
During the iteration of the PR https://github.com/doctrine/migrations/pull/1296 an idea to deprecate passing values to this option was suggested, given how complicated was to handle using that option without any value, which led to a bug and inconsistency with the documentation.

As specified in the deprecation message, using this option from 4.0.x onwards in the CLI will be treated as `true`, and the option itself won't accept values.

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
+++ b/lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Migrations\Tools\Console;
 
+use Doctrine\Deprecations\Deprecation;
 use Doctrine\Migrations\Configuration\Configuration;
 use Doctrine\Migrations\MigratorConfiguration;
 use Symfony\Component\Console\Input\InputInterface;
@@ -28,16 +29,30 @@ class ConsoleInputMigratorConfigurationFactory implements MigratorConfigurationF
 
     private function determineAllOrNothingValueFrom(InputInterface $input): bool|null
     {
-        if (! $input->hasOption('all-or-nothing')) {
-            return null;
+        $allOrNothingOption        = null;
+        $wasOptionExplicitlyPassed = $input->hasOption('all-or-nothing');
+
+        if ($wasOptionExplicitlyPassed) {
+            $allOrNothingOption = $input->getOption('all-or-nothing');
         }
 
-        $allOrNothingOption = $input->getOption('all-or-nothing');
+        if ($wasOptionExplicitlyPassed && $allOrNothingOption !== null) {
+            Deprecation::trigger(
+                'doctrine/migrations',
+                'https://github.com/doctrine/migrations/issues/1304',
+                <<<'DEPRECATION'
+                    Context: Passing values to option `--all-or-nothing`
+                    Problem: Passing values is deprecated
+                    Solution: From version 4.0.x, `--all-or-nothing` option won't accept any value, 
+                    and the presence of the option will be treated as `true`.
+                    DEPRECATION,
+            );
+        }
 
         if ($allOrNothingOption === 'notprovided') {
             return null;
         }
 
-        return (bool) ($allOrNothingOption ?? true);
+        return (bool) ($allOrNothingOption ?? false);
     }
 }
--- a/lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
+++ b/lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
@@ -43,8 +43,8 @@ class ConsoleInputMigratorConfigurationFactory implements MigratorConfigurationF
                 <<<'DEPRECATION'
                     Context: Passing values to option `--all-or-nothing`
                     Problem: Passing values is deprecated
-                    Solution: From version 4.0.x, `--all-or-nothing` option won't accept any value, 
-                    and the presence of the option will be treated as `true`.
+                    Solution: If you need to disable the behavior, omit the option,
+                    otherwise, pass the option without a value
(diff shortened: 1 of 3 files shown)


lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
```diff
@@ -31,12 +32,26 @@ public function getMigratorConfiguration(InputInterface $input): MigratorConfigu
...
+        if ($wasOptionExplicitlyPassed) {
             $allOrNothingOption = $input->getOption('all-or-nothing');
         }
 
+        if ($wasOptionExplicitlyPassed && $allOrNothingOption !== null) {
+            Deprecation::trigger(
+                'doctrine/migrations',
+                'https://github.com/doctrine/migrations/issues/1304',
+                <<<'DEPRECATION'
+                    Context: Passing values to the `--all-or-nothing`
```
greg0ire:
```suggestion
                    Context: Passing values to option `--all-or-nothing`
```

lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
```diff
@@ -31,12 +32,26 @@ public function getMigratorConfiguration(InputInterface $input): MigratorConfigu
...
+                    Problem: The mechanism to read the option value is unnecessarily complicated.
```
greg0ire: The problem must describe what the user is doing wrong, not why we decided to deprecate passing values.

```suggestion
                    Problem: Passing values is deprecated
```

lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
```diff
@@ -31,12 +32,26 @@ public function getMigratorConfiguration(InputInterface $input): MigratorConfigu
...
+                    Solution: From version 4.0.x, `--all-or-nothing` option won't accept any value,
```
greg0ire: The solution should describe how to avoid the deprecation.

```suggestion
                    Solution: If you need to disable the behavior, omit the option,
                    otherwise, pass the option without a value
```

tests/Doctrine/Migrations/Tests/Tools/Console/Command/MigrateCommandTest.php
```diff
@@ -402,7 +411,7 @@ public function allOrNothing(): Generator
         yield [false, ['--all-or-nothing' => true], true];
         yield [false, ['--all-or-nothing' => 1], true];
         yield [false, ['--all-or-nothing' => '1'], true];
-        yield [false, ['--all-or-nothing' => null], true];
+        yield [false, ['--all-or-nothing' => null], true, false];
```
greg0ire: You should update the `psalm-return` annotation to add `bool?` I think.

greg0ire:
Please add an `Upgrade to 3.6` section to `UPGRADE.md`

agustingomes:
@greg0ire I'll apply the suggestions later today

lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
agustingomes: > If you need to disable the behavior, omit the option,

From my understanding, this won't be the case. If the configuration has this option set to `true`, omitting the CLI option does not change it to `false`, or is this what is planned for the 4.x version?

greg0ire (commented):
@agustingomes sorry I forgot about this… can you rebase please :pray: ?

greg0ire:
ping @agustingomes :)

agustingomes:
@greg0ire No worries. I'll do the rebase and push it afterwards.

agustingomes:
@greg0ire I rebased it against `3.6.x`, let me know if this ok for you, or if I should rebase it against other branch.

lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
agustingomes: I updated the Deprecation with your suggestion.

greg0ire:
Deprecations should go on the next minor branch. Right now that's 3.7.x, sorry.

agustingomes:
@greg0ire no problem, I'll rebase my branch on `3.7.x`

agustingomes:
@greg0ire I rebased the branch, but I don't know how to check the BC Break locally, as it seems the tool that does that check is not getting installed when doing `composer install`.

greg0ire (approved):
The bc check failure looks unrelated, I think it can be ignored. I shall try to look into it at some point

SenseException (approved):

greg0ire:
Thanks @agustingomes !

vitormattos:
After this merge the follow row started to trigger the deprecated message:

https://github.com/doctrine/migrations/blob/3.7.x/lib/Doctrine/Migrations/Tools/Console/Command/MigrateCommand.php#L81

I think that as a follow up of this PR will be necessary remove this row.

greg0ire:
I don't understand. Please open a full fledged issue.

agustingomes:
@vitormattos if I understand correctly, this will fix the problem, as the logic to throw/not throw the deprecation was incomplete:
https://github.com/doctrine/migrations/pull/1381

It already has an issue associated to it, which has been linked in the PR.

cc @greg0ire

greg0ire:
Oh right, forgot this wasn't merged already.

PR #1308, opened 2023-01-17
Fixed `all-or-nothing` console input value determination.

e-jevdokimovs:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes

#### Summary

This fixes a bug that was introduced in #1296 where the configuration value of `MigratorConfiguration::allOrNothing` got set to `true` if the command didn't have an `all-or-nothing` option. Now the value is set from input options only if the command that is being executed has an `all-or-nothing` option.


agustingomes (commented):
Nice catch!

agustingomes (approved):

greg0ire (approved):

greg0ire:
Thanks @e-jevdokimovs !

PR #1326, opened 2023-02-23
Make it possible to `addSql()` that is executed as a statement

### LINKED ISSUE #1325: Cannot add `CREATE TRIGGER` statements with the DBAL `mysqli` driver
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | guess no
| Version     | 3.5.5

More specifically,

* doctrine/migrations 3.5.5
* doctrine/dbal 3.6.0
* PHP 8.1.6
* ext-mysqli with client API library version "mysqlnd 8.1.16"

#### Summary

Passing a `CREATE TRIGGER ...` statement to `$this->addSql()` in a migration fails when the SQL is executed.

#### Current behavior

The migration fails with the error message:

`An exception occurred while executing a query: This command is not supported in the prepared statement protocol yet`

#### How to reproduce

Create a migration:

```php
class Version... { ...
    {
        $this->addSql('CREATE TRIGGER some_trigger BEFORE UPDATE ON some_table FOR EACH ROW UPDATE other_table SET some_column = NEW.other_column WHERE fk = NEW.id');
    }
}
```

Then, execute the migration using the `mysqli` driver.

#### Expected behavior

Migration being executed with no error, trigger created.

#### Additional information

The `DbalExecutor` calls `Connection::executeQuery()`:

https://github.com/doctrine/migrations/blob/dc441bb16c2f92ec2d095d7ba6e9ad91b5d6f7c5/lib/Doctrine/Migrations/Version/DbalExecutor.php#L293-L298

This, in turn, will call `\Doctrine\DBAL\Driver\Connection::query()` (on the DBAL Driver/Connection).

https://github.com/doctrine/dbal/blob/4173dfc23de02e2ac0c99cfb16fcd8af5fedc23b/src/Driver/Mysqli/Connection.php#L68-L71

In `\Doctrine\DBAL\Driver\Mysqli\Connection::prepare()`, we ultimately get to `\mysqli::prepare()`, which is where the initial exception is thrown.

Things work when I use `Connection::executeStatement()` on the DBAL connection, since that will use `\Doctrine\DBAL\Driver\Connection::exec()` on the DBAL Driver/Connection, at least when no parameters need to be bound.

`executeStatement` also seems more appropriate, at least for trigger creation. The code snippet quoted above includes a comment why `executeQuery()` is called, but in the Git history I can only trace that back to a reverted use of `executeUpdate()`; so basically it has been `executeQuery()` since a long time.

#1326 adds a new parameter to `addSql()` to indicate that the statement should be executed with `executeStatement()`, which would solve the issue.

mpdude:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | fixes #1325

#### Summary

When the DBAL connection uses `mysqli` as the driver, prepared statements are sent to the MySQL server through a dedicated protocol mechanism. For example, `CREATE TRIGGER` statements are not possible in this case.

To use SQL statements like `CREATE TRIGGER`, the `DbalExecutor` may not (at least in the case of the `mysqli` driver) use `Connection::executeQuery()`, but has to call `Connection::executeStatement()`. See #1325 for more details.

This PR adds a new `executeAsStatement` parameter to `\Doctrine\Migrations\AbstractMigration::addSql()`, which defaults to `false` (current behaviour). 

By setting it to `true`, a migration can pass the information to the `DbalExecutor` that the statement must be executed with `executeStatement()` instead of `executeQuery()`.

### DIFF
--- a/lib/Doctrine/Migrations/AbstractMigration.php
+++ b/lib/Doctrine/Migrations/AbstractMigration.php
@@ -126,9 +126,10 @@ abstract class AbstractMigration
     protected function addSql(
         string $sql,
         array $params = [],
-        array $types = []
+        array $types = [],
+        bool $executeAsStatement = false
     ): void {
-        $this->plannedSql[] = new Query($sql, $params, $types);
+        $this->plannedSql[] = new Query($sql, $params, $types, $executeAsStatement);
     }
 
     /** @return Query[] */
--- a/lib/Doctrine/Migrations/Query/Query.php
+++ b/lib/Doctrine/Migrations/Query/Query.php
@@ -21,19 +21,22 @@ final class Query
     /** @var mixed[] */
     private array $types;
 
+    private bool $executeAsStatement;
+
     /**
      * @param mixed[] $parameters
      * @param mixed[] $types
      */
-    public function __construct(string $statement, array $parameters = [], array $types = [])
+    public function __construct(string $statement, array $parameters = [], array $types = [], bool $executeAsStatement = false)
     {
         if (count($types) > count($parameters)) {
             throw InvalidArguments::wrongTypesArgumentCount($statement, count($parameters), count($types));
         }
 
-        $this->statement  = $statement;
-        $this->parameters = $parameters;
-        $this->types      = $types;
+        $this->statement          = $statement;
+        $this->parameters         = $parameters;
+        $this->types              = $types;
+        $this->executeAsStatement = $executeAsStatement;
     }
 
     public function __toString(): string
@@ -56,5 +59,10 @@ final class Query
     public function getTypes(): array
     {
         return $this->types;
+    }
+
+    public function getExecuteAsStatement(): bool
+    {
+        return $this->executeAsStatement;
     }
 }
--- a/lib/Doctrine/Migrations/Version/DbalExecutor.php
+++ b/lib/Doctrine/Migrations/Version/DbalExecutor.php
@@ -290,8 +290,13 @@ final class DbalExecutor implements Executor
             $this->outputSqlQuery($query, $configuration);
 
             $stopwatchEvent = $this->stopwatch->start('query');
-            // executeQuery() must be used here because $query might return a result set, for instance REPAIR does
-            $this->connection->executeQuery($query->getStatement(), $query->getParameters(), $query->getTypes());
+
+            if ($query->getExecuteAsStatement()) {
+                $this->connection->executeStatement($query->getStatement(), $query->getParameters(), $query->getTypes());
+            } else {
+                $this->connection->executeQuery($query->getStatement(), $query->getParameters(), $query->getTypes());
+            }
+
             $stopwatchEvent->stop();
 
             if (! $configuration->getTimeAllQueries()) {
--- a/tests/Doctrine/Migrations/Tests/Version/ExecutorTest.php
+++ b/tests/Doctrine/Migrations/Tests/Version/ExecutorTest.php
@@ -17,6 +17,7 @@ use Doctrine\Migrations\Provider\SchemaDiffProvider;
 use Doctrine\Migrations\Query\Query;
 use Doctrine\Migrations\Tests\TestLogger;
 use Doctrine\Migrations\Tests\Version\Fixture\EmptyTestMigration;
+use Doctrine\Migrations\Tests\Version\Fixture\TestMigrationWithStatement;
 use Doctrine\Migrations\Tests\Version\Fixture\VersionExecutorTestMigration;
 use Doctrine\Migrations\Version\DbalExecutor;
 use Doctrine\Migrations\Version\Direction;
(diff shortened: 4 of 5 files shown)


mpdude:
Please advise what to do about the PHPStan error.

stof (approved):

stof:
Ideally, most things should be executed as statements, but using a different default would be a BC break.

lib/Doctrine/Migrations/AbstractMigration.php
```diff
@@ -126,9 +126,10 @@ public function down(Schema $schema): void
     protected function addSql(
         string $sql,
         array $params = [],
-        array $types = []
+        array $types = [],
+        bool $executeAsStatement = false
     ): void {
-        $this->plannedSql[] = new Query($sql, $params, $types);
+        $this->plannedSql[] = new Query($sql, $params, $types, $executeAsStatement);
     }
```
greg0ire, line 133: I think boolean flags are a code smell. Let's add a new method called `addSqlStatement()` instead?

lib/Doctrine/Migrations/Query/Query.php
```diff
@@ -57,4 +60,9 @@ public function getTypes(): array
     {
         return $this->types;
     }
+
+    public function getExecuteAsStatement(): bool
```
greg0ire, line 64: This implies that this is an SQL query, but for some reason we want to execute it as a statement. Apparently, it's the opposite: https://stackoverflow.com/questions/4735856/difference-between-a-statement-and-a-query-in-sql

I would create a parent class called `Statement` to model this better, and then use `instanceof`.

greg0ire (changes requested):

lib/Doctrine/Migrations/Query/Query.php
stof: Well, the hard part being that even though queries are statements in the understanding of SQL in this SO answer, this is not entirely true for the DBAL API.
If you use `executeStatement` to execute a statement that returns a record set (i.e. to execute a query), some drivers will fail due to receiving a non expected record set. For instance, PDO only supports having 1 buffer at a time for record sets, so you must consume or close the result of a query before executing another query: https://github.com/doctrine/migrations/pull/888#issuecomment-722318249

stof: Also, DBAL (and also the ORM) seem to use Query as the more generic name (see `QueryBuilder` as name)

lib/Doctrine/Migrations/AbstractMigration.php
mpdude: You mean, like a static constructor method ob `Query`?

greg0ire: No, I mean like a new method next to `addSql`, called `addSqlStatement`. 

```php
protected function addSqlStatement(
        string $sql,
        array $params = [],
        array $types = [],
): void {
    $this->plannedSql[] = new Query($sql, $params, $types, true);
}

lib/Doctrine/Migrations/Query/Query.php
greg0ire: Still, it feels weird to leak a DBAL implementation detail in the public API of this value object, or even for that value object to have knowledge about it. 

According to the SO answer, queries are statements, and not the other way around. According to DBAL, queries and statements are different things (so would be best modeled as 2 classes extending a common one I suppose?).

> Also, DBAL (and also the ORM) seem to use Query as the more generic name (see QueryBuilder as name)

I don't think `Doctrine\ORM\Query` and `Doctrine\ORM\Query\AST\*Statement` can really be compared if that is what you are referring to, and how the ORM treats one as more generic as the other.

As for `QueryBuilder`… I guess it's badly named (although it's a mistake I could have made myself before today)? Other parts of the DBAL don't seem to treat query as the more generic name (otherwise `Doctrine\DBAL\Statement` would not have that name). Anyway, if there is a naming/modeling mistake in other packages, I don't see a huge benefit in mimicking it for the sake of consistency. It's IMO fine for `doctrine/migrations` to have its own model of statements and queries, to fulfill its own purposes, regardless of whether it's using DBAL or another API.

I suggest we model the SQL standard instead of mimicking the DBAL API, that way the Query class is not modeled after a lower-level API, but against a higher-level, common abstraction that the DBAL also knows about: SQL.

greg0ire:
Looking at the code, it's hard to tell why this is needed at all. Please create a document under `docs/en/explanation` explaining this WTF. TL;DR: we need to make the distinction because there isn't an API that allows us to send statements without caring whether they are queries, and a buffer needs to be consumed, or whether they are just statements and we need to avoid preparing the statement. It might also be worth opening an issue on `doctrine/dbal`, because this issue you are facing seems to happen only because `Mysqli\Connection` uses `prepare`, unlike connections for other drivers.

greg0ire:
Regarding the DBAL issue, it seems that the implementation uses `prepare()` since the beginning (12 years ago): https://github.com/doctrine/dbal/pull/72

I don't know if that has been challenged since then, but maybe it should, and maybe this should be closed in favor of a DBAL PR calling  `mysqli::query()` (which does not prepare anything, and can be used to execute statements I suppose).

derrabus:
This is a MySQLi oddity and we should try to solve it in the DBAL. 

> I don't know if that has been challenged since then, but maybe it should, and maybe this should be closed in favor of a DBAL PR calling `mysqli::query()` (which does not prepare anything, and can be used to execute statements I suppose).

Yes. The problem with that is that we would basically need a second `Result` implementation for MySQLi. `mysqli::query()` returns a `mysqli_result` instance while our `Result` class operates on an executed `mysqli_statement` instance. Yay, MySQLi.

stof:
@greg0ire this is not specific to mysqli. PDO also has this behavior (see the link I gave in the discussion)

greg0ire:
@stof #1325 is specific to MySQLi. The issue you linked (https://github.com/doctrine/migrations/pull/888#issuecomment-722318249) is the opposite one: it's about needing to call `executeQuery()`, when using `REPAIR`. I believe that issue is solved right now, BTW.

mpdude:
Feels like I stirred up a hornets' nest. I don't think I can follow your discussion, please advise what to do :-)

derrabus:
> please advise what to do

Fix this method so it does not prepare a statement.

https://github.com/doctrine/dbal/blob/63646ffd71d1676d2f747f871be31b7e921c7864/src/Driver/Mysqli/Connection.php#L68-L71

See my comment above on why this is more challenging that it might look at first glance.

PR #1328, opened 2023-02-25
Apply assets filter on sequences

ottaviano:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Asset filter is used for tables and sequences on `fromSchema`, but only for tables on `toSchema`. It generates `CREATE SEQUENCE` migration query. This PR applies the filter for sequences on `toSchema`.

### DIFF
--- a/src/Generator/DiffGenerator.php
+++ b/src/Generator/DiffGenerator.php
@@ -120,6 +120,16 @@ class DiffGenerator
 
                 $toSchema->dropTable($tableName);
             }
+
+            foreach ($toSchema->getSequences() as $sequence) {
+                $sequenceName = $sequence->getName();
+
+                if ($schemaAssetsFilter($sequenceName)) {
+                    continue;
+                }
+
+                $toSchema->dropSequence($sequenceName);
+            }
         }
 
         return $toSchema;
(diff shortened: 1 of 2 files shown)


SenseException:
Thank you for the PR, @ottaviano. Did you run into a specific problem that resulted into this change?

ottaviano:
@SenseException,
I have a partitioned table `project_result_item` with 3 sub-tables: `project_result_item_0`, `project_result_item_1`, and `project_result_item_2`. 
When I run the `doctrine:migration:diff` command, it tries to drop the 3 sub-tables. 

The "solution" I found is to completely ignore the `project_result_item` table with `schema_filter`, but Doctrine migration generates the sequence for the ignored table at every diff : 
```CREATE SEQUENCE project_result_item_id_seq INCREMENT BY 1 MINVALUE 1 START 1```

This happens because the filter does not apply to sequences during the `toSchema` phase.

SenseException (changes requested):
Please put the test for your changes into its own test case instead of mixing them with the one with the table name.

Issue #1329, opened 2023-03-08
Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated.


o5:
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | no
| RFC         | no
| BC Break    | no

#### Summary

Do not use deprecated `$defaultName`.

```bash
* Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Doctrine\Migrations\Tools\Console\Command\CurrentCommand" class instead.

* Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Doctrine\Migrations\Tools\Console\Command\DiffCommand" class instead.

* Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Doctrine\Migrations\Tools\Console\Command\DumpSchemaCommand" class instead.

* Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Doctrine\Migrations\Tools\Console\Command\ExecuteCommand" class instead.

* Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Doctrine\Migrations\Tools\Console\Command\GenerateCommand" class instead.

* Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Doctrine\Migrations\Tools\Console\Command\LatestCommand" class instead.

* Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Doctrine\Migrations\Tools\Console\Command\ListCommand" class instead.

* Since symfony/console 6.1: Relying on the static property "$defaultName" for setting a command name is deprecated. Add the "Symfony\Component\Console\Attribute\AsCommand" attribute to the "Doctrine\Migrations\Tools\Console\Command\MigrateCommand" class instead.

[...]

PR #1330, opened 2023-03-10
Add attributes to commands

gordinskiy:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #1329 

Add usage of `Symfony\Component\Console\Attribute\AsCommand` attributes

### DIFF
--- a/lib/Doctrine/Migrations/Tools/Console/Command/CurrentCommand.php
+++ b/lib/Doctrine/Migrations/Tools/Console/Command/CurrentCommand.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace Doctrine\Migrations\Tools\Console\Command;
 
 use Doctrine\Migrations\Exception\MigrationClassNotFound;
+use Symfony\Component\Console\Attribute\AsCommand;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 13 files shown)


derrabus:
Thank you. Can you please also add the `description:` parameter to all attributes?

derrabus (approved):

pkruithof:
@derrabus any progress on this?

greg0ire:
The backward compat check can be ignored, but the static analysis check should get fixed before this can be merged.

gordinskiy:
> Ignored error pattern ^Call to function method_exists\(\) with Doctrine\\Migrations\\Metadata\\Storage\\MetadataStorage and 'getSql' will always evaluate to true\.$ in path /home/runner/work/migrations/migrations/lib/Doctrine/Migrations/Version/DbalExecutor.php was not matched in reported errors.

How to fix this error correctly?
Remove the check, and move the method from the annotation to the interface itself?

greg0ire:
Rebasing might be enough: it does not seem to come from your changes. I'll try closing and reopening in case that does the trick.

greg0ire (approved):

greg0ire:
Thanks @gordinskiy !

PR #1331, opened 2023-03-14
build: Fix cleanup of the backup composer lock file

theofidry:
The trap function was registered as the last bash instruction and as a result was not executed if anything failed before, e.g. the composer update command or building the PHAR itself.

In the event the build PHAR script fails, the temporary backup composer.lock file may be created and not removed.

Also add the backup composer.lock file to the gitignore for good measure.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ dist/
 tests/phpunit.xml
 vendor/
 /composer.lock
+/composer.lock.back
 phpunit.xml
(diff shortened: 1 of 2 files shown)


greg0ire (approved):

greg0ire:
Thanks @theofidry!

PR #1332, opened 2023-03-14
build: Use the new compactors

theofidry:
Those compactors are stricly identical to the old ones, only the FQCN changed (and they are removed in Box 4.x).

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no


greg0ire (approved):

greg0ire:
Thanks @theofidry!

PR #1346, opened 2023-07-29
Exclude phpstan file with .gitattributes

VincentLanglet:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

Same idea @greg0ire 

There is also possibly
- download-box.sh
- build-phar.sh

but I don't know what there are doing

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,5 @@ build.xml               export-ignore
 /phpcs.xml.dist         export-ignore
 /phpstan.neon.dist      export-ignore
 composer.lock           export-ignore
+/download-box.sh        export-ignore
+/build-phar.sh          export-ignore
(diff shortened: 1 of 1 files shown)


greg0ire:
They are used when running tests or running the phar, I think it is safe to remove them as well.

VincentLanglet:
> They are used when running tests or running the phar, I think it is safe to remove them as well.

I added them

greg0ire (approved):

greg0ire:
Thanks!

VincentLanglet:
> They are used when running tests or running the phar, I think it is safe to remove them as well.

I just saw the `.doctrine-project.json` were downloaded too on every project
https://github.com/doctrine/migrations/blob/3.6.x/.doctrine-project.json

Do you know the purpose of this file ? Can we add it too to the gitattributes ?

Issue #1348, opened 2023-08-11
DependencyFactory::getEventDispatcher triggers a deprecation


VincentLanglet:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | x.y.z

#### Summary

`DependencyFactory::getEventDispatcher` is implemented this way:
```
return $this->getDependency(EventDispatcher::class, function (): EventDispatcher {
    return new EventDispatcher(
        $this->getConnection(),
        $this->getConnection()->getEventManager(),
    );
});
```
but `Connection::getEventManager` is deprecated by dbal since https://github.com/doctrine/dbal/issues/5784

Not sure how to fix this @greg0ire @derrabus

derrabus:
Deprecate calling `fromConnection()` without passing an event manager?

VincentLanglet:
Seems like `EntityManagerInterface::getEventManager` exists ; I'll try this instead.

derrabus:
> Seems like `EntityManagerInterface::getEventManager` exists ; I'll try this instead.

Yes, but that'll only work if you have an entity manager. You can use migrations with a connection only.

stof:
`fromConnection` should indeed allow passing an EventManager explicitly, using `Connection::getEventManager` only as fallback (triggering a deprecation telling you to pass an EventManager explicitly).
`fromEntityManager` would register the EventManager using the ORM EventManager.

`DependencyFactory::getEventDispatcher` would then use a `DependencyFactory::getEventManager` method reading the configured event manager.

VincentLanglet:
DoctrineCommand is doing `DependencyFactory::fromConnection` does it mean we have to add `EventManager` to the constructor of all doctrine command ?

stof:
To be compatible with CLI commands, we might need a concept of an event manager loader instead of passing directly an event manager.

VincentLanglet:
I feel like this is solved since I use dbal 4.x without any issues.

PR #1349, opened 2023-08-11
Stop using deprecated method Connection::getEventManager

VincentLanglet:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #1348


stof:
-1 for this PR, as it breaks the usage of the library without an entity manager.

VincentLanglet:
I updated but i still need some changes, cf https://github.com/doctrine/migrations/issues/1348#issuecomment-1674773109

PR #1351, opened 2023-08-18
Modernize codebase for PHP 8

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

We require PHP 8.1 already, so let's use what PHP 8 has to offer for us.

### DIFF
--- a/lib/Doctrine/Migrations/Configuration/Migration/YamlFile.php
+++ b/lib/Doctrine/Migrations/Configuration/Migration/YamlFile.php
@@ -35,7 +35,7 @@ final class YamlFile extends ConfigurationFile
 
         try {
             $config = Yaml::parse($content);
-        } catch (ParseException $e) {
+        } catch (ParseException) {
             throw YamlNotValid::malformed();
         }
 
--- a/lib/Doctrine/Migrations/SchemaDumper.php
+++ b/lib/Doctrine/Migrations/SchemaDumper.php
@@ -16,16 +16,13 @@ use function array_merge;
 use function count;
 use function implode;
 use function preg_last_error;
+use function preg_last_error_msg;
 use function preg_match;
 use function restore_error_handler;
 use function set_error_handler;
 use function sprintf;
 
-use const PREG_BACKTRACK_LIMIT_ERROR;
-use const PREG_BAD_UTF8_ERROR;
-use const PREG_BAD_UTF8_OFFSET_ERROR;
 use const PREG_INTERNAL_ERROR;
-use const PREG_RECURSION_LIMIT_ERROR;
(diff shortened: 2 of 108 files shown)


lib/Doctrine/Migrations/Configuration/Migration/YamlFile.php
```diff
@@ -35,7 +35,7 @@ public function getConfiguration(): Configuration
 
         try {
             $config = Yaml::parse($content);
-        } catch (ParseException $e) {
+        } catch (ParseException) {
```
stof, line 38: shouldn't we rather attach it as a previous exception, so that the user can see the actual parsing issue ?

lib/Doctrine/Migrations/SchemaDumper.php
```diff
@@ -166,27 +149,14 @@ private static function pregMatch(string $pattern, string $subject, ?array &$mat
...
-                    $error = 'Offset doesn\'t correspond to the begin of a valid UTF-8 code point.';
-                    break;
-                default:
-                    $error = 'Error.';
-            }
-
-            throw new InvalidArgumentException($error);
+            throw new InvalidArgumentException(match (preg_last_error()) {
+                PREG_INTERNAL_ERROR => sprintf('Internal PCRE error, please check your Regex. Reported errors: %s.', implode(', ', $errorMessages)),
+                PREG_BACKTRACK_LIMIT_ERROR => 'pcre.backtrack_limit reached.',
```
stof: this could be replaced by using `preg_last_error_message` instead

lib/Doctrine/Migrations/Configuration/Migration/YamlFile.php
derrabus: Yes, but that's out if scope for this PR. We can do this in a follow-up.

lib/Doctrine/Migrations/SchemaDumper.php
derrabus: I guess.

derrabus:
BC check failure:

> [BC] CHANGED: The parameter $service of Doctrine\Migrations\DependencyFactory#setService() changed from no type to a non-contravariant object|callable
> [BC] CHANGED: The parameter $service of Doctrine\Migrations\DependencyFactory#setService() changed from no type to object|callable

I think those changes are safe. We've only ever accepted those types anyway and existing child classes that override this method without parameter types should continue to work.

greg0ire (approved):

SenseException (approved):

PR #1352, opened 2023-08-18
Drop support for Symfony 4

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

Symfony 4 is EOL. Let's make our lives easier by removing it from our test matrix.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,8 +30,8 @@
         "doctrine/deprecations": "^0.5.3 || ^1",
         "doctrine/event-manager": "^1.2 || ^2.0",
         "psr/log": "^1.1.3 || ^2 || ^3",
-        "symfony/console": "^4.4.16 || ^5.4 || ^6.0",
-        "symfony/stopwatch": "^4.4 || ^5.4 || ^6.0",
+        "symfony/console": "^5.4 || ^6.0",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -30,8 +30,8 @@
         "doctrine/deprecations": "^0.5.3 || ^1",
         "doctrine/event-manager": "^1.2 || ^2.0",
         "psr/log": "^1.1.3 || ^2 || ^3",
-        "symfony/console": "^4.4.16 || ^5.4 || ^6.0",
-        "symfony/stopwatch": "^4.4 || ^5.4 || ^6.0",
+        "symfony/console": "^5.4 || ^6.0",
```
stloyd, line 33: Symfony 6.3 is the one with support (from 6.x line) right now so it can be bumped as well

derrabus: Right, but as long as we support 5.4, we don't gain much from dropping 6.0-6.2.

stof (approved):

stloyd (approved):

greg0ire (approved):

SenseException (approved):

PR #1354, opened 2023-08-18
Ditch action in favor of running tool directly

### LINKED ISSUE #1353: The roave BC check CI is broken
See https://github.com/Nyholm/roave-bc-check-docker/issues/32 for the report on the docker image being used to run that step.

According to https://github.com/Nyholm/roave-bc-check-docker/issues/32#issuecomment-1488287882, running the tool on the runner itself is also faster, so it might make sense to move away from the action.

greg0ire:
Closes #1353

This may be replaced by a Doctrine workflow once we have figured out what flexibility we need by doing this on several repositories.

I stole part of this from https://github.com/yiisoft/actions/blob/master/.github/workflows/bc.yml

### DIFF
--- a/.github/workflows/bc-check.yml
+++ b/.github/workflows/bc-check.yml
@@ -24,7 +24,13 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: "Roave BC Check"
-        uses: "docker://nyholm/roave-bc-check-ga"
+      - name: Install PHP with extensions.
+        uses: shivammathur/setup-php@v2
         with:
-          args: "--from=${{ github.event.pull_request.base.sha }}"
+          php-version: "8.2"
+
+      - name: Install roave/backward-compatibility-check.
+        run: composer require --dev roave/backward-compatibility-check
+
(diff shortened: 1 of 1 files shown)


greg0ire:
Breaking change detected: https://github.com/doctrine/migrations/actions/runs/5906000004/job/16021134090?pr=1354#step:5:230

It works!

.github/workflows/bc-check.yml
```diff
@@ -24,7 +24,11 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: "Roave BC Check"
-        uses: "docker://nyholm/roave-bc-check-ga"
-        with:
-          args: "--from=${{ github.event.pull_request.base.sha }}"
+      - name: Install PHP with extensions.
+        uses: shivammathur/setup-php@v2
```
derrabus, line 28: We should pin the PHP version.

greg0ire: Addressed :slightly_smiling_face:

derrabus (approved):

PR #1364, opened 2023-10-15
Use github actions format for BC check

greg0ire:
It is a better experience than having to check the logs.


derrabus:
> The "--format" option does not exist.

🤔

greg0ire:
Well, that's embarassing :sweat_smile:

greg0ire:
Oh right, I added it to the Composer command line :facepalm:

derrabus (approved):

PR #1368, opened 2023-10-26
Remove unmaintained symfony libraries

Shidarr:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     |no
| Fixed issues | 

#### Summary

Drop unmaintained symfony 4.4 libraries

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,9 +30,9 @@
         "doctrine/deprecations": "^0.5.3 || ^1",
         "doctrine/event-manager": "^1.2 || ^2.0",
         "psr/log": "^1.1.3 || ^2 || ^3",
-        "symfony/console": "^5.4 || ^6.0",
-        "symfony/stopwatch": "^5.4 || ^6.0",
-        "symfony/var-exporter": "^6.2"
+        "symfony/console": "^5.4 || ^6.0 || ^7.0",
+        "symfony/stopwatch": "^5.4 || ^6.0 || ^7.0",
+        "symfony/var-exporter": "^6.2 || ^7.0"
     },
     "require-dev": {
         "ext-pdo_sqlite": "*",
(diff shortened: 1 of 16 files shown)


stof:
This should be submitted in the 3.7.x branch to remove support for Symfony 4.4 in the next minor version rather than in the next patch version IMO.

Shidarr:
Good Idea. Maybe drop support for Versions 5.* also? Because if you update symfony to latest (6.3) Version, you will get some deprecation warnings.

stof:
I don't understand your proposal. Why wanting to drop support for 5.x (where 5.4 is still maintained for 1 year plus an extra year of security fixes) ? you're saying that you get some deprecation warnings when updating to 6.3, but I don't understand how this relates to the support of 5.x

Shidarr:
How to get rid of the deprecation warnings then? 
They are mentioned here also: https://github.com/doctrine/migrations/issues/1329
But the AsCommand Attribute is already added. Deleting the Parameter results in errors in older versions. 
I can also add the Name in configure function and delete parameter. That should work too.

derrabus:
Did you check if this allows us to drop any compat code?

Shidarr:
Ok I just updated the PR to use setName Function instead of $defaultName. This works for old and new symfony/console Versions. Maybe this is better? Then 5.4 is still supported.

stof:
the AsCommand attribute is supported starting in Symfony 5.3. So you don't need to drop support for 5.4.

Shidarr:
Yes with my latest changes 5.4 does not need to be dropped and there is no deprecation warning anymore.

derrabus:
Dropping Symfony 4 was done in #1352 already. The changes you've made to the commands should not be necessary since the `AsCommand` attribute should suppress the deprecation regarding overriding the static properties.

derrabus:
Closing in favor of #1370.

PR #1377, opened 2023-11-20
Fix formatParameter() for boolean

jderusse:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | yes
| Fixed issues | n/a

#### Summary

This PR fixes BC break when formatting a boolean parameters (and type is not defined)

```
        $this->addSql('INSERT INTO XXX (..., is_bar) VALUES (..., :is_bar)', [
            'is_bar' => true,
        ]);
```

### DIFF
--- a/lib/Doctrine/Migrations/InlineParameterFormatter.php
+++ b/lib/Doctrine/Migrations/InlineParameterFormatter.php
@@ -53,7 +53,7 @@ final class InlineParameterFormatter implements ParameterFormatter
         return sprintf('with parameters (%s)', implode(', ', $formattedParameters));
     }
 
-    private function formatParameter(mixed $value, string|int $type): string|int|float|null
+    private function formatParameter(mixed $value, string|int $type): string|int|bool|float|null
     {
         if (is_string($type) && Type::hasType($type)) {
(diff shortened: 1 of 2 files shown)


lib/Doctrine/Migrations/InlineParameterFormatter.php
```diff
@@ -53,7 +53,7 @@ public function formatParameters(array $params, array $types): string
         return sprintf('with parameters (%s)', implode(', ', $formattedParameters));
     }
 
-    private function formatParameter(mixed $value, string|int $type): string|int|float|null
+    private function formatParameter(mixed $value, string|int $type): string|int|bool|float|null
```
stof, line 56: Why this change ? `parameterToString` returns a string.

jderusse: `parameterToString` does, but `Type::getType('string')->convertToDatabaseValue(...)` don't

jderusse:
friendly reminder

derrabus (approved):

derrabus:
Thank you!

Issue #1379, opened 2023-11-28
Deprecation always triggers with or without `--all-or-nothing` option


endelwar:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.7.1

#### Summary

When running `bin/console doctrine:migrations:migrate` deprecation for passing values to `--all-or-nothing` always triggers because option has default value.

#### Current behavior

Due to https://github.com/doctrine/migrations/pull/1296 a default option for `--all-or-nothing` is always set ('notprovided'), making check on https://github.com/doctrine/migrations/blob/78484f91fef85f2d579cd0a4a662d77f3b742398/lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php#L33-L34 to always set `$allOrNothingOption` to a not null value, which then triggers the deprecation:

```
User Deprecated: Context: Passing values to option `--all-or-nothing` Problem: Passing values is deprecated Solution: If you need to disable the behavior, omit the option, otherwise, pass the option without a value (ConsoleInputMigratorConfigurationFactory.php:44 called by ConsoleInputMigratorConfigurationFactory.php:22, https://github.com/doctrine/migrations/issues/1304, package doctrine/migrations) {"exception":"[object] (ErrorException(code: 0): User Deprecated: Context: Passing values to option `--all-or-nothing`\nProblem: Passing values is deprecated\nSolution: If you need to disable the behavior, omit the option,\notherwise, pass the option without a value (ConsoleInputMigratorConfigurationFactory.php:44 called by ConsoleInputMigratorConfigurationFactory.php:22, https://github.com/doctrine/migrations/issues/1304, package doctrine/migrations) at /Users/manuel/Sites/myproject/vendor/doctrine/deprecations/lib/Doctrine/Deprecations/Deprecation.php:210)
```

#### How to reproduce

Run `bin/console doctrine:migrations:migrate` and look at the logs

#### Expected behavior

I expect no deprecation triggered when not setting the `--all-or-nothing` option

#### Proposed solution

Remove the default value of `all-or-nothing` option in `Doctrine\Migrations\Tools\Console\Command\MigrateCommand::configure`

greg0ire:
Cc @agustingomes

greg0ire:
> Proposed solution

> Remove the default value of all-or-nothing option in Doctrine\Migrations\Tools\Console\Command\MigrateCommand::configure

[...]

PR #1381, opened 2023-11-30
GH-1379: Improve Deprecation thrown check + logic

agustingomes:
Unfortunately as part of the previous improvement to determine whether a deprecation is thrown, validating that the deprecation was not thrown ended up not being added explicitly.

In addition, adds the explicit expectation of not throwing the deprecation whenever the `--all-or-nothing` is not indicated, which was the root cause originating the issue.

|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/migrations/issues/1379

#### Summary

When the deprecation was added, the check whether the deprecation is NOT triggered was missing. This PR should solve the underlying issue.


greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @agustingomes !

PR #1385, opened 2023-12-15
ci: fix variable reference

simPod:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

`deps` does not exist. The key in matrix is `dependencies`


derrabus (approved):

derrabus:
Good catch, thank you.

PR #1391, opened 2024-01-21
Flatten directory tree

greg0ire:


derrabus (approved):

SenseException (approved):
As long as everyone is using this lib with the composer autoloader, this should be fine. I assume we don't support the non-composer way. It would be a pain i we did.

PR #1408, opened 2024-03-06
fix: Allow enum param types: ArrayParameterType and ParameterType

### LINKED ISSUE #1407: DBAL 4.0 and ArrayParameterType
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes/no
| Version     | 3.7.3

#### Summary

Tying to migrate to DBAL 4 (4.0.1) and in v3 we had `Connection::PARAM_INT_ARRAY` deprecated in favour of `ArrayParameterType::INTEGER`.

Then in v4 `ArrayParameterType` has been changed to enum, so it is a type itself.

Example query we have in the migrations:

```php
$this->addSql('DELETE FROM books WHERE id IN (:ids)', ['ids' => $booksIds], ['ids' => ArrayParameterType::INTEGER]);
```

#### Current behavior

Exception:

```
Fatal error: Uncaught TypeError: Doctrine\Migrations\InlineParameterFormatter::formatParameter(): Argument #2 ($type) must be of type string|int, Doctrine\DBAL\ArrayParameterType given, called in [...]/vendor/doctrine/migrations/lib/Doctrine/Migrations/InlineParameterFormatter.php on line 46 and defined in [...]/vendor/doctrine/migrations/lib/Doctrine/Migrations/InlineParameterFormatter.php on line 56

TypeError: Doctrine\Migrations\InlineParameterFormatter::formatParameter(): Argument #2 ($type) must be of type string|int, Doctrine\DBAL\ArrayParameterType given, called in [...]/vendor/doctrine/migrations/lib/Doctrine/Migrations/InlineParameterFormatter.php on line 46 in [...]/vendor/doctrine/migrations/lib/Doctrine/Migrations/InlineParameterFormatter.php on line 56
```

#### How to reproduce

As I see the typehint here: 
https://github.com/doctrine/migrations/blob/3.7.x/lib/Doctrine/Migrations/InlineParameterFormatter.php#L56

is just `string|int $type`, so it would not work with DBAL 4.0

It is kinda similar to: https://github.com/doctrine/dbal/issues/6309 

#### Expected behavior

No errors.

michalbundyra:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1407

#### Summary

Fixes #1407

A bit lame fix, but it looks like we did not really use other int types before anyway

With this change we can use DBAL v4 new enum types: `ArrayParameterType` and `ParameterType`.

~~Bumping version of DBAL to 3.6 as before that `ArrayParameterType` did not exist.~~

Just to note - the problem itself is ONLY with DBAL 4.0+ as on 3.6 these types were classes with constants, not enums.

### DIFF
--- a/lib/Doctrine/Migrations/InlineParameterFormatter.php
+++ b/lib/Doctrine/Migrations/InlineParameterFormatter.php
@@ -4,7 +4,9 @@ declare(strict_types=1);
 
 namespace Doctrine\Migrations;
 
+use Doctrine\DBAL\ArrayParameterType;
 use Doctrine\DBAL\Connection;
+use Doctrine\DBAL\ParameterType;
 use Doctrine\DBAL\Types\Type;
 
 use function array_map;
@@ -53,7 +55,7 @@ final class InlineParameterFormatter implements ParameterFormatter
         return sprintf('with parameters (%s)', implode(', ', $formattedParameters));
     }
 
-    private function formatParameter(mixed $value, string|int $type): string|int|bool|float|null
+    private function formatParameter(mixed $value, string|int|ArrayParameterType|ParameterType $type): string|int|bool|float|null
     {
         if (is_string($type) && Type::hasType($type)) {
             return Type::getType($type)->convertToDatabaseValue(
--- a/lib/Doctrine/Migrations/InlineParameterFormatter.php
+++ b/lib/Doctrine/Migrations/InlineParameterFormatter.php
@@ -4,9 +4,7 @@ declare(strict_types=1);
 
 namespace Doctrine\Migrations;
 
-use Doctrine\DBAL\ArrayParameterType;
 use Doctrine\DBAL\Connection;
-use Doctrine\DBAL\ParameterType;
 use Doctrine\DBAL\Types\Type;
 
 use function array_map;
@@ -55,7 +53,7 @@ final class InlineParameterFormatter implements ParameterFormatter
         return sprintf('with parameters (%s)', implode(', ', $formattedParameters));
     }
 
-    private function formatParameter(mixed $value, string|int|ArrayParameterType|ParameterType $type): string|int|bool|float|null
+    private function formatParameter(mixed $value, mixed $type): string|int|bool|float|null
     {
         if (is_string($type) && Type::hasType($type)) {
             return Type::getType($type)->convertToDatabaseValue(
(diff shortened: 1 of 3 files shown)


derrabus:
Thank you. Please add a test that covers your change.

lib/Doctrine/Migrations/InlineParameterFormatter.php
```diff
@@ -53,7 +55,7 @@ public function formatParameters(array $params, array $types): string
         return sprintf('with parameters (%s)', implode(', ', $formattedParameters));
     }
 
-    private function formatParameter(mixed $value, string|int $type): string|int|bool|float|null
+    private function formatParameter(mixed $value, string|int|ArrayParameterType|ParameterType $type): string|int|bool|float|null
```
derrabus: I wonder if we should change the type to `mixed`. After all, the method discards any value that is not a string anyway.

michalbundyra: @derrabus pushed - if we ok with mixed, then I can even change back the bump in composer.json, but then we would not have a test case for ArrayParameterType as it does not exist in DBAL 3.5

tests/Doctrine/Migrations/Tests/InlineParameterFormatterTest.php
```diff
@@ -54,11 +58,13 @@ public function testFormatParameters(): void
             'unknown',
             'unknown',
             'unknown',
+            ParameterType::STRING,
+            ArrayParameterType::INTEGER,
```
derrabus:
```suggestion
            class_exists(ArrayParameterType::class) ? ArrayParameterType::INTEGER : Connection::PARAM_INT_ARRAY,
```

This way we can avoid to bump dependencies in a bugfix release.

michalbundyra: pushed, and this is what I was worried about - phpstan failure... :O do you want to me to add ignore on line?

derrabus: Yes, we can ignore the error.

michalbundyra: @derrabus pushed, green now.

any chance to get it released soon? thanks

derrabus (approved):

michalbundyra:
@derrabus thanks!

derrabus:
Thank you!

derrabus:
Let's do the bump in #1411.

PR #1409, opened 2024-03-06
Raise PHPStan level to 8

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

Let's make our static analysis a little bit stricter.

### DIFF
--- a/src/DependencyFactory.php
+++ b/src/DependencyFactory.php
@@ -145,7 +145,7 @@ class DependencyFactory
     {
         if ($this->configuration === null) {
             $this->configuration = $this->configurationLoader->getConfiguration();
-            $this->freeze();
+            $this->frozen = true;
         }
 
         return $this->configuration;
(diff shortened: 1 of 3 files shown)


src/DependencyFactory.php
```diff
@@ -145,7 +145,7 @@ public function getConfiguration(): Configuration
     {
         if ($this->configuration === null) {
             $this->configuration = $this->configurationLoader->getConfiguration();
-            $this->freeze();
+            $this->frozen = true;
```
SenseException, line 148: Why these changes where the `frozen()` method gets replaced with the direct set to true?

derrabus: PHPStan treats the method calls as impure and complains that `$this->configuration` might be `null` when we return it.

SenseException: I was playing around with annotations and other configurations, but it seems like PHPStan can't be convinced that `$this->freeze()` does nothing evil with other property states.

SenseException (approved):

src/DependencyFactory.php
derrabus: Yes, and this is fine imho. `freeze()` is impure, so PHPStan's behavior is correct.

PR #1418, opened 2024-03-17
Remove migration table from generated comparison schema

### LINKED ISSUE #1406: doctrine:schema:validate returns DROP TABLE doctrine_migration_versions;
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 3.0.0

#### Summary

When I updated doctrine/orm to version 3.0.0, my CI started to fail due to the doctrine:schema check:validate, which returned DROP TABLE doctrine_migration_versions;

I found that adding schema_filter can fix this problem: "~^(?!doctrine_migration_versions$)~" in the dbal configuration, indeed, then the error no longer appears, but after that my migrations stopped tracking the current state and think that at the moment no migration has been performed.

In version 3.0, the saveMode argument was removed in \Doctrine\ORM\Tools\SchemaTool::getUpdateSchemaSql and because of this, doctrine:schema:validate tracks deleted tables and returns sql for this action.
telegram-cloud-photo-size-2-5384566650814190356-y

#### Previous behavior

image

#### Current behavior

image

#### How to reproduce

1. composer require doctrine/doctrine-migrations-bundle
2. composer require doctrine/orm-pack
3. bin/console doctrine:migration:diff
4. bin/console doctrine:migration:migrate
5. bin/console doctrine:schema:validate -v

wmouwen:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1406

#### Summary

> doctrine/orm#11374
> 
> Schemas generated from metadata cause the event `ToolEvents::postGenerateSchema` to be triggered in the method `SchemaTool::getSchemaFromMetadata`. It would be helpful to do the same for schemas generated for comparison in `SchemaTool::createSchemaForComparison`. This PR introduces the new event ToolEvents::postGenerateComparisonSchema.

This new event, given that the PR in `doctrine/orm` proceeds, can be utilized in a listener to remove the metadata table from the comparison schema.

PR relates to an attempt to fix doctrine/migrations#1406. It requires doctrine/orm#11374 and is a requirement of doctrine/DoctrineMigrationsBundle#529.

### DIFF
--- a/lib/Doctrine/Migrations/Event/Listeners/RemoveMigrationTableFromSchemaListener.php
+++ b/lib/Doctrine/Migrations/Event/Listeners/RemoveMigrationTableFromSchemaListener.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Doctrine\Migrations\Event\Listeners;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\DependencyFactory;
+use Doctrine\Migrations\Metadata\Storage\TableMetadataStorageConfiguration;
+use Doctrine\ORM\Tools\Event\GenerateSchemaEventArgs;
+
+class RemoveMigrationTableFromSchemaListener
+{
+    private TableMetadataStorageConfiguration|null $configuration = null;
+
+    public function __construct(
+        private readonly DependencyFactory $dependencyFactory,
+    ) {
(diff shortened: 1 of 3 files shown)


wmouwen:
Closing due to inactivity. Given that doctrine/orm#11409 takes the maintainer's preference over doctrine/orm#11374, this PR isn't going to happen.

PR #1421, opened 2024-04-04
Compare versions of migrations without namespace

beregovoy:
Strip namespace, so namespace do not affect version comparsion.

|      Q       |   A
|------------- | -----------
| BC Break     | no
| Fixed issues | 

### Summary
Previous behaviour
Migrations\Version2022 will always migrate AFTER Brand\Migrations\Version2024, but based on pure versions - Version2022 SHOULD be migrated BEFORE Version2024. 

ZBrand\Migrations\Version2024 - will migrate AFTER Migrations\Version2022. So comparsion was not actually performed on version, but on namespace first. Which is kind of wrong, non intuitive, and break namespace functional.

New behaviour ignore namespace, and compare versions only by version. Default behaviour should not depend on namespace. If someone need to override it - he should did it explicitly for project.

### DIFF
--- a/lib/Doctrine/Migrations/Version/AlphabeticalComparator.php
+++ b/lib/Doctrine/Migrations/Version/AlphabeticalComparator.php
@@ -10,6 +10,12 @@ final class AlphabeticalComparator implements Comparator
 {
     public function compare(Version $a, Version $b): int
     {
-        return strcmp((string) $a, (string) $b);
+        return strcmp($this->stripNamespace($a), $this->stripNamespace($b));
+    }
+
+    private function stripNamespace(Version $version): string
+    {
+        $path = explode('\\', (string) $version);
+        return end($path);
     }
 }


beregovoy:
I may add tests, if concept will be accepted

SenseException:
Hi @beregovoy

it does make sense to have some expectations about the order of executed, but changing this in version 3.7 would be a bc break. I'd expect that this could be the default comparator in 4.0. A new comparator instance could still be injected manually in older versions though.

HenryVolkmer:
Hi folks,
Hi @Ocramius,

Regarding the BCs, I propose to implement the fix as outlined [here](https://github.com/HenryVolkmer/migrations/blob/bugfix/1144-version-sorting/src/Version/Version.php#L15-L24) rather than modifying the Comparator.

This approach only changes the string representation of Version-Class, which ensures that already migrated data remains unaffected. Additionally, the alpha-comparator now sorts versions correctly.

As a result, the version string will look like this: `Version20250114110700 (Foo\Bar\Migrations)` 
(Note: This change applies only to V3 versions, legacy versions remain untouched.)

.. and 5 tests will need to be adjusted accordingly

Issue #1431, opened 2024-05-23
Please throw a warning/exception if `addSql` is used in post methods.


VincentLanglet:
Hi, I'd like to kinda restart the discussion in https://github.com/doctrine/migrations/issues/369 (cc @stof).

I just encounter a production bug because I wrote a migration with a `addSql` call in a postUp migration and it wasn't executed.

I understand there are reason for not rewriting the DbalExecutor in order to supports `addSql` calls in post methods.
But in this case would it be possible to throw an error when such call are made ?

In my case such exception would have been thrown locally and/or on my CI and I would have fix the issue before it was deployed in production.

derrabus:
I would be open to that. Can you prepare a PR that we can discuss on?

VincentLanglet:
> I would be open to that. Can you prepare a PR that we can discuss on?

Sure, I proposed an implementation here https://github.com/doctrine/migrations/pull/1432

PR #1432, opened 2024-05-23
Throw an exception if addSql is used in post methods

VincentLanglet:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | ?
| Fixed issues | #1431

#### Summary

Using `addSql` in post methods should be reported since the sql will be ignored during the migration.


greg0ire (approved):

greg0ire:
Why are you targeting 3.7.x :thinking: That's not a bugfix, is it?

VincentLanglet:
> Why are you targeting 3.7.x 🤔 That's not a bugfix, is it?

I target the stable/default branch by habit. ^^

SenseException (approved):

greg0ire:
Thanks @VincentLanglet !

PR #1433, opened 2024-06-13
Enlarge table to avoid to be truncated

kouinkouin:
|      Q       |   A
|------------- | -----------
| Type         | bug (documentation)
| BC Break     | no
| Fixed issues | 

#### Summary

On this page (https://www.doctrine-project.org/projects/doctrine-migrations/en/3.0/reference/configuration.html), from v3.0 to current version, in the table under "Here are details about what each configuration option does:", the lines "migrations_paths" and "organize_migrations" are truncated.

The PR permits to show all the text of the cell.

The patch should be merged in v3.0+ branches.

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -116,21 +116,21 @@ Please note that if you want to use the YAML configuration option, you will need
 
 Here are details about what each configuration option does:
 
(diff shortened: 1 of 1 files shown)


greg0ire:
Sorry, we do not maintain branches older than 3.7.x, even for docs.

PR #1436, opened 2024-06-26
Reuse the SqlFormatter instance

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

`SqlFormatter` has a heavy constructor that compiles a couple of regular expressions. We can achieve a little speedup by reusing the formatter instance.


derrabus:
Failing PHPStan job addressed in #1437.

greg0ire (approved):

Issue #1443, opened 2024-08-23
`--all-or-nothing` without value computes to `false`


MatTheCat:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | not sure
| Version     | 3.7.4

#### Summary

There seems to be a discrepancy between this deprecation: https://github.com/doctrine/migrations/blob/535a70dcbd88b8c6ba945be050977457f4f4c06c/src/Tools/Console/ConsoleInputMigratorConfigurationFactory.php#L46-L49
and this test: https://github.com/doctrine/migrations/blob/535a70dcbd88b8c6ba945be050977457f4f4c06c/tests/Tools/Console/Command/MigrateCommandTest.php#L384
which asserts that `--all-or-nothing` without value will compute to `false`.

#### Current behavior

Running a migration with the `--all-or-nothing` option without value, “all or nothing” mode is disabled (whatever the value from the configuration).

#### How to reproduce

Run a migration passing the `--all-or-nothing` option without value.

#### Expected behavior

“All or nothing” mode is enabled.

greg0ire:
@agustingomes can you please look into it?

agustingomes:
@greg0ire I'll try during the weekend. Likely is the deprecation text that should be adjusted.

MatTheCat:
Maybe this is due to the fact I’m using the DoctrineMigrationsBundle which defaults `all_or_nothing` to `false`?

Either way passing this option to the CLI and not having it enabled came as a surprise.

MatTheCat:
Note that it seems the current behavior doesn’t match [the documentation](https://www.doctrine-project.org/projects/doctrine-migrations/en/3.8/reference/configuration.html#from-the-command-line) either:

> You can also set this option from the command line with the `migrate` command and the `--all-or-nothing` option:
> ```
> ./vendor/bin/doctrine-migrations migrate --all-or-nothing
> ```

agustingomes:
@MatTheCat @greg0ire I created a PR with an initial iteration of the possible fix: https://github.com/doctrine/migrations/pull/1444

Let's discuss in the PR how to move forward.

agustingomes:
I believe this issue may now be closed with the new `3.8.1` release

https://github.com/doctrine/migrations/releases/tag/3.8.1

MatTheCat:
That was fast! Thanks ♥️

PR #1444, opened 2024-08-23
GH-1443: Fix default option when `--all-or-nothing` option is used as intended

agustingomes:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/migrations/issues/1443
#### Summary

After a review of the issue reported (linked above), it seems the default behavior of specifying the option is opposite of what was intended, and also what is currently documented. This PR applies the correct default when the user specifies the `--all-or-nothing` option in the command line.

### Background:

Previous PR's and their linked issues contain more context regarding each iteration so far. They are presented below in the chronological order in which they were merged:
- https://github.com/doctrine/migrations/pull/1296
- https://github.com/doctrine/migrations/pull/1308
- https://github.com/doctrine/migrations/pull/1305
- https://github.com/doctrine/migrations/pull/1381

### DIFF
--- a/docs/en/reference/configuration.rst
+++ b/docs/en/reference/configuration.rst
@@ -243,18 +243,23 @@ Setting ``transactional`` to ``false`` will disable that.
 From the Command Line
 ~~~~~~~~~~~~~~~~~~~~~
 
-You can also set this option from the command line with the ``migrate`` command and the ``--all-or-nothing`` option:
+To override the configuration and explicitly enable All or Nothing Transaction from the command line,
+use the ``--all-or-nothing`` option:
 
 .. code-block:: sh
 
     $ ./vendor/bin/doctrine-migrations migrate --all-or-nothing
 
-If you have it enabled at the configuration level and want to change it for an individual migration you can
-pass a value of ``0`` or ``1`` to ``--all-or-nothing``.
+.. note::
+
+    Passing options to --all-or-nothing is deprecated from 3.7.x, and will not be allowed in 4.x
+
+To override the configuration and explicitly disable All or Nothing Transaction from the command line,
+use the ``--no-all-or-nothing`` option:
 
 .. code-block:: sh
 
-    $ ./vendor/bin/doctrine-migrations migrate --all-or-nothing=0
+    $ ./vendor/bin/doctrine-migrations migrate --no-all-or-nothing
 
 Connection Configuration
 ------------------------
--- a/src/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
+++ b/src/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
@@ -31,30 +31,55 @@ class ConsoleInputMigratorConfigurationFactory implements MigratorConfigurationF
 
     private function determineAllOrNothingValueFrom(InputInterface $input): bool|null
     {
-        $allOrNothingOption        = null;
+        $enableAllOrNothingOption  = self::ABSENT_CONFIG_VALUE;
+        $disableAllOrNothingOption = null;
+
+        if ($input->hasOption('no-all-or-nothing')) {
+            $disableAllOrNothingOption = $input->getOption('no-all-or-nothing');
+        }
+
         $wasOptionExplicitlyPassed = $input->hasOption('all-or-nothing');
 
         if ($wasOptionExplicitlyPassed) {
-            $allOrNothingOption = $input->getOption('all-or-nothing');
+            /**
+             * Due to this option being able to receive optional values, its behavior is tricky:
+             * - when `--all-or-nothing` option is not provided, the default is set to self::ABSENT_CONFIG_VALUE
+             * - when `--all-or-nothing` option is provided without values, this will be `null`
+             * - when `--all-or-nothing` option is provided with a value, we get the provided value
+             */
+            $enableAllOrNothingOption = $input->getOption('all-or-nothing');
         }
 
-        if ($wasOptionExplicitlyPassed && ($allOrNothingOption !== null && $allOrNothingOption !== self::ABSENT_CONFIG_VALUE)) {
+        $enableAllOrNothingDeprecation = match ($enableAllOrNothingOption) {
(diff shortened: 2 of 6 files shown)


lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
```diff
@@ -45,15 +45,14 @@ private function determineAllOrNothingValueFrom(InputInterface $input): bool|nul
                 <<<'DEPRECATION'
                     Context: Passing values to option `--all-or-nothing`
                     Problem: Passing values is deprecated
-                    Solution: If you need to disable the behavior, omit the option,
-                    otherwise, pass the option without a value
+                    Solution: how we should deal with this option being `true` in config but user want to disable it in command line for 4.x?
```
agustingomes: Added intentionally this line, since I'm wondering how to handle this.

Also as an extra thought, since we've added more coverage to how the option is determined, I wonder if it still makes sense to deprecate this ability, or if we still want to go ahead with disallowing values to be passed for this option.

agustingomes:
@greg0ire The CI/CD failures do not seem directly related to my changes as far as I could see. Can you take a look if it's something I need to fix in my PR?

lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
MatTheCat: What about a new `--no-all-or-nothing` option?

greg0ire:
You were targeting a branch that is no longer maintained. Please rebase.

greg0ire:
Still, the errors occur on 3.8.x, I'll try to address them.

agustingomes:
@greg0ire rebasing the branch and the extra commit fixed the CI issues.

greg0ire:
Sorry, I forgot to mention I made an attempt to address the SA issues in #1445

lib/Doctrine/Migrations/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
greg0ire: > Also as an extra thought, since we've added more coverage to how the option is determined, I wonder if it still makes sense to deprecate this ability, or if we still want to go ahead with disallowing values to be passed for this option.

I don't think there's much value in having several ways of doing the same thing. It just adds complexity.

> What about a new --no-all-or-nothing option?

Might be good, and is easy to implement with https://symfony.com/blog/new-in-symfony-5-3-negatable-command-options

agustingomes: @greg0ire that's a good idea, but it seems for our particular use case it won't work. looking at the docs:

> Negatable options are only available for options that don't allow passing any value to them (their previous type should be InputOption::VALUE_NONE).

For the remainder of the 3.x lifetime, it is better the users continue passing the optional value as it's currently allowed, and we introduce the negatable option in 4.x.

greg0ire: Maybe we can use a negatable option in 4.x, and a full-fledged option in 3.x, literally named `no-all-or-nothing` to help with the transition?

agustingomes: indeed, we can do that. my latest push should have a good starting point.

src/Tools/Console/ConsoleInputMigratorConfigurationFactory.php
```diff
@@ -31,30 +31,50 @@ public function getMigratorConfiguration(InputInterface $input): MigratorConfigu
...
+            /**
+             * Due to this option being able to receive optional values, its behavior is tricky:
+             * - when `--all-or-nothing` option is not provided, the default is set to self::ABSENT_CONFIG_VALUE
+             * - when `--all-or-nothing` option is provided without values, this will be `null`
+             * - when `--all-or-nothing` option is provided with a value, we get the provided value
+             */
+            $enableAllOrNothingOption = $input->getOption('all-or-nothing');
+        }
+
+        if ($enableAllOrNothingOption !== self::ABSENT_CONFIG_VALUE && $disableAllOrNothingOption === true) {
```
agustingomes, line 58: I will improve this logic, as it currently won't throw if we pass a value to `--all-or-nothing` option.

Edit: it was already correct. my bad.

docs/en/reference/configuration.rst
```diff
@@ -243,18 +243,23 @@ Setting ``transactional`` to ``false`` will disable that.
 From the Command Line
 ~~~~~~~~~~~~~~~~~~~~~
 
-You can also set this option from the command line with the ``migrate`` command and the ``--all-or-nothing`` option:
+To enable all the migrations to complete together, you can also set this option from the command line with
```
greg0ire: I'm not sure people will understand what is meant by "completing together". I think we might not want to shy away from being technical here, talking about transactions should make clarify things IMO.

greg0ire:
The SA issues are fixed, you can now rebase and drop your second commit.

docs/en/reference/configuration.rst
agustingomes: Pushed a new phrasing, let me know if it's ok the way is written now.

agustingomes:
@greg0ire with the release of PHPStan 1.12 today a new error surfaced. How do you want to proceed?

greg0ire (approved):

SenseException (approved):

SenseException:
> with the release of PHPStan 1.12 today a new error surfaced. How do you want to proceed?

It affects code that you haven't touched. It needs to be fixed, but not by this PR. See #1453.

greg0ire:
@agustingomes you may rebase :)

greg0ire (approved):

PR #1445, opened 2024-08-24
Add phpdoc for LogInterface::log() $message argument

greg0ire:
PHPStan complains that it is not specified. Looking at the method signature of Psr\Log\LoggerTrait, it appears that $level uses a phpdoc, and both $message and $context use native type declarations, so [the corresponding phpdoc was removed in 3.0.1](https://github.com/php-fig/log/pull/80)

### DIFF
--- a/src/Tools/Console/ConsoleLogger.php
+++ b/src/Tools/Console/ConsoleLogger.php
@@ -71,7 +71,8 @@ final class ConsoleLogger extends AbstractLogger
     /**
      * {@inheritDoc}
      *
-     * @param mixed[] $context
+     * @param string|Stringable $message
+     * @param mixed[]           $context
      */
     public function log($level, $message, array $context = []): void
     {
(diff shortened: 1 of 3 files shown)


derrabus:
Maybe the doc blocks shouldn't have been removed in the first place. 🤔 

cc @lyrixx

greg0ire:
Maybe, but now that this ship has sailed, I wouldn't revert it I think.

goetas:
Shouldn't require 3.0.1 as minimum version fix the issue?

greg0ire:
I don't think it would, because the issue appears with 3.0.1 IMO

SenseException (approved):
`{@inheritDoc}` should actually do the trick for phpstan. I don't know why the phpdocs need to be added to make it work. :thinking:

greg0ire:
@SenseException since 3.0.1, there is nothing to inherit, the phpdoc was dropped in the PR I linked in my original message.

derrabus (approved):

lyrixx (commented):
👍🏼 It's better now

PR #1449, opened 2024-08-26
Remove ConsoleLogger

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no (removed class was `@internal`)
| Fixed issues | N/A

#### Summary

We don't need to maintain our own version of the `ConsoleLogger` class. Symfony already has one, so let's use that. This will also allow us to revert #1446.


greg0ire (approved):
> This will also allow us to revert https://github.com/doctrine/migrations/pull/1446

Why would you want to revert it?

derrabus:
> > This will also allow us to revert #1446
> 
> Why would you want to revert it?

Because once we merge this PR, we don't ship our own psr/log implementation anymore. Thus, we don't need to have an opinion on which psr/log version is installed.

greg0ire:
OK, but you'd still keep the native type declarations, right?

derrabus:
> OK, but you'd still keep the native type declarations, right?

My plan is that we don't implement that interface ourselves at all. See #1450 for the remaining implementations.

PR #1450, opened 2024-08-26
Remove test logger implementations

derrabus:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

I'd like to leverage the `fig/log-test` package so we don't have to invent our own mock implementations of the psr/log interfaces.


greg0ire:
You're only touching tests here, so why not target 3.8.x? We don't want new tests to use the previous way.

greg0ire (approved):

derrabus:
Good point, I'm rebasing.

PR #1453, opened 2024-08-27
Assert there is no failure

greg0ire:
The pattern is not dynamic, and we know it works. This addresses an issue reported by PHPStan.

### DIFF
--- a/src/Generator/Generator.php
+++ b/src/Generator/Generator.php
@@ -9,6 +9,7 @@ use Doctrine\Migrations\Generator\Exception\InvalidTemplateSpecified;
 use Doctrine\Migrations\Tools\Console\Helper\MigrationDirectoryHelper;
 use InvalidArgumentException;
 
+use function assert;
 use function explode;
 use function file_get_contents;
 use function file_put_contents;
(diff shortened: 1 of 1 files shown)


SenseException (approved):

src/Generator/Generator.php
```diff
@@ -74,11 +75,14 @@ public function generateMigration(
         string|null $up = null,
         string|null $down = null,
     ): string {
-        $mch = [];
-        if (preg_match('~(.*)\\\\([^\\\\]+)~', $fqcn, $mch) === 0) {
+        $mch         = [];
+        $matchResult = preg_match('~(.*)\\\\([^\\\\]+)~', $fqcn, $mch);
+        if ($matchResult === 0) {
```
derrabus, line 80: I've had the same problem in DBAL. If you change this check to…

```php
if ($matchResult !== 1) {
```

… you should not need the assert. Alternatively, we could think about using the composer/pcre library which has a saner API.

greg0ire: @derrabus that's the solution I went first at first, but then I felt wrong because there's a tiny chance that  `false` might occur and misdiagnose it as an invalid FQCN. `composer/pcre` might indeed be a good solution.

PR #1463, opened 2024-10-10
Compensate difference between DBAL schema introspection and ORM generation

### LINKED ISSUE #1415: `CREATE SCHEMA public` is added to all down migrations in Postgres
### Bug Report

|    Q        |   A
|------------ | ------
| Version     | 3.8.3

#### Summary

I'm using the bundle with Postrgres 16. The database URL is `postgresql://$DB_USER:$DB_PASSWORD@db:5432/$DB_NAME?serverVersion=16.2&charset=utf8`. But on every `doctrine:migrations:diff` it creates migrations like the following:

```php
final class Version20240311103945 extends AbstractMigration
{
    public function getDescription(): string
    {
        return '';
    }

    {
        // this up() migration is auto-generated, please modify it to your needs

    }

    public function down(Schema $schema): void
    {
        // this down() migration is auto-generated, please modify it to your needs
        $this->addSql('CREATE SCHEMA public');
    }
}
```

No matter what are other changes, it ALWAYS adds the `$this->addSql('CREATE SCHEMA public');` line to all migrations.

How to fix it?

morozov:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no

Fixes #1415.

There is a discrepancy between how the DBAL represents the introspected schema and how the ORM generates the schema that represents entities:
- The DBAL includes all existing namespaces into the introspected schema, including the default ones like `public` for PostgreSQL.
- The ORM doesn't include any namespaces into the generated schemas, even if the schema contains tables in those namespaces. This behavior is incorrect, but fixing it isn't trivial (https://github.com/doctrine/orm/pull/11658).

The above discrepancy leads to false differences between the schemas, which we want to compensate here:
- As of DBAL 4.0, the PostgreSQL schema manager sets the introspected schema name to the name of the current schema
- If it's set, Migrations will create the corresponding namespace in the schema generated by the ORM.

### DIFF
--- a/src/Generator/DiffGenerator.php
+++ b/src/Generator/DiffGenerator.php
@@ -12,6 +12,7 @@ use Doctrine\DBAL\Schema\Schema;
 use Doctrine\Migrations\Generator\Exception\NoChangesDetected;
 use Doctrine\Migrations\Provider\SchemaProvider;
 
+use function method_exists;
 use function preg_match;
 use function strpos;
 use function substr;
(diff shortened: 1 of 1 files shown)


greg0ire (approved):
It works :tada: ! Thanks a lot for fixing this!!!

PR #1473, opened 2024-11-29
Resolves #939 replaced single quotes with nowdoc

### LINKED ISSUE #939: Nowdoc strings
### Feature Request

Generate multiline SQL (from `--formatted`) statements using nowdoc strings instead of current single quoted strings.

|    Q        |   A
|------------ | ------
| New Feature | yes
| RFC         | no
| BC Break    | no

#### Summary

The main advantage of nowdoc strings is that you don't need to escape characters inside, so a raw query can be put inside. It makes sense for multiline queries (like create table, for example), because escaping breaks SQL highlighting in IDE (PHPStorm, for example, highlights SQL strings, but if there are escaped quotes (`'... COMMENT \'(DC2Type:carbondatetime)\' ...'`), the whole highlighting breaks and reports it as a syntax error).

Not a big deal, but IMO nice to have.

averatec1337:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | #939

### DIFF
--- a/src/Generator/SqlGenerator.php
+++ b/src/Generator/SqlGenerator.php
@@ -14,10 +14,11 @@ use function array_unshift;
 use function count;
 use function get_class;
 use function implode;
+use function preg_replace;
 use function sprintf;
+use function str_repeat;
 use function stripos;
 use function strlen;
-use function var_export;
 
 /**
  * The SqlGenerator class is responsible for generating the body of the up() and down() methods for a migration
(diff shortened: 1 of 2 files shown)


greg0ire:
Thanks for working on this. It seems there are CI jobs failing. Please take a look at [this guide](https://www.doctrine-project.org/contribute/#dealing-with-checks-and-tools) for more on how to handle those.

averatec1337:
@greg0ire thanks, I had to fix units.

greg0ire:
You need to fix cs as well, and you might want to retarget on `origin/3.9.x`

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @kamil-karkus !!!

averatec1337:
@greg0ire Thank you for taking care of this issue. I was planning to get back to it on Monday, following the guide you sent.

vinise:
Really sad not to see any option on this feature.
could have been great to be able to disable it, or an option to activate this either on formatted sql only or based on --line-length=LINE-LENGTH (or other option)
(I might create a PR for this if I find some time)
 
 ;)

PR #1476, opened 2024-12-13
Unable to create migration: The namespace with name "public" already exists.

VictorKrasnov:
fix: unable to make migration on Postgres

Error message "The namespace with name "public" already exists." on d:m:diff or migration:make fixed.

### DIFF
--- a/src/Generator/DiffGenerator.php
+++ b/src/Generator/DiffGenerator.php
@@ -70,7 +70,7 @@ class DiffGenerator
             ! method_exists($this->schemaManager, 'getSchemaSearchPaths')
             && $this->platform->supportsSchemas()
         ) {
-            $defaultNamespace = $toSchema->getNamespaceName();
+            $defaultNamespace = $toSchema->getNamespaceName() ?? '';
             if ($defaultNamespace !== '') {
                 $toSchema->createNamespace($defaultNamespace);
             }
--- a/src/Generator/DiffGenerator.php
+++ b/src/Generator/DiffGenerator.php
@@ -70,7 +70,7 @@ class DiffGenerator
             ! method_exists($this->schemaManager, 'getSchemaSearchPaths')
             && $this->platform->supportsSchemas()
         ) {
-            $defaultNamespace = $toSchema->getName();
+            $defaultNamespace = $toSchema->getNamespaceName();
             if ($defaultNamespace !== '') {
                 $toSchema->createNamespace($defaultNamespace);
             }


greg0ire:
Sorry but… what was wrong with this message, and under which circumstances does it occur?

VictorKrasnov:
I updated Symfony 6.4 to 7.2 and Doctrine 2 to 3.3 yesterday. After this I tried to run commands like make:migration or doctrine:migration:diff. Any attempt to create migration ended with error: "The namespace with name "public" already exists". I debugged your code and found that a value of variable $defaultNamespace is setting with schema name. So as I understood, maybe mistaken, you tried to use schema name for namespace name. I don't know what is namespace in PostgreSQL, but changes that I published here made my application work.

Sorry, but I can't make some reproduction for you, my project is private and not small at all. However, if necessary, I can share, for example, my doctrine configuration.

greg0ire:
This was introduced very recently, in https://github.com/doctrine/migrations/pull/1463 cc @morozov maybe you will understand what's going on?

VictorKrasnov:
And one more thing. I also tried to figure out my issue another way. I found the table "shemata" in the database "information_shema". I found record here with shema "public", and I checked that schema owner is the same user, that is used in Symfony. So, I had schema public already, but Doctrine tried to create another one when I tried to create some new migration.

greg0ire:
Regarding the confusion about "schema", I recommend you read this: https://github.com/doctrine/dbal/blob/c91166a91b8639a1d7bfe0f88bca0273bdbe9d2d/src/Schema/Schema.php#L21-L44

greg0ire:
~Also, I think this is [a known issue](https://github.com/doctrine/migrations/issues/1415#issuecomment-2449682796). I recommend you upgrade to DBAL 4~

You are already using DBAL 4, right?

VictorKrasnov:
Thanks for your explanation of namespaces in doctrine. It's clear, so my pull request seems wrong. 

My doctrine versions are:

```
# composer show doctrine/*
doctrine/cache                      2.2.0  PHP Doctrine Cache library is a popular cache implementation that supports many different drivers such as redis, memcache, apc, mongodb and others.
doctrine/collections                2.2.2  PHP Doctrine Collections library that adds additional functionality on top of PHP arrays.
doctrine/common                     3.4.5  PHP Doctrine Common project is a library that provides additional functionality that other Doctrine projects depend on such as better reflection s...
doctrine/dbal                       4.2.1  Powerful PHP database abstraction layer (DBAL) with many features for database schema introspection and management.
doctrine/deprecations               1.1.4  A small layer on top of trigger_error(E_USER_DEPRECATED) or PSR-3 logging with options to disable all deprecations or selectively for packages.
doctrine/doctrine-bundle            2.13.1 Symfony DoctrineBundle
doctrine/doctrine-migrations-bundle 3.3.1  Symfony DoctrineMigrationsBundle
doctrine/event-manager              2.0.1  The Doctrine Event Manager is a simple PHP event system that was built to be used with the various Doctrine projects.
doctrine/inflector                  2.0.10 PHP Doctrine Inflector is a small library that can perform string manipulations with regard to upper/lowercase and singular/plural forms of words.
doctrine/instantiator               2.0.0  A small, lightweight utility to instantiate objects in PHP without invoking their constructors
doctrine/lexer                      3.0.1  PHP Doctrine Lexer parser library that can be used in Top-Down, Recursive Descent Parsers.
doctrine/migrations                 3.8.2  PHP Doctrine Migrations project offer additional functionality on top of the database abstraction layer (DBAL) for versioning your database schema...
doctrine/orm                        3.3.0  Object-Relational-Mapper for PHP
doctrine/persistence                3.4.0  The Doctrine Persistence project is a set of shared interfaces and functionality that the different Doctrine object mappers share.
doctrine/sql-formatter              1.5.1  a PHP SQL highlighting library
```

greg0ire:
> The DBAL includes all existing namespaces into the introspected schema, including the default ones like public for PostgreSQL.

If you are getting that error, then it must mean that in your case, `public` is not included in the introspected schema for some reason. It should be present in both `$fromSchema` and `$toSchema`, but right now, it only seems to be included in `$toSchema`.

VictorKrasnov:
Снимок экрана 2024-12-14 в 13 04 58

greg0ire:
Oh I see, this isn't an SQL error, this is a DBAL exception because you already have `public` inside `$toSchema->namespaces` :thinking: 

Reading the code, It seems we expect to either have `_name` set to `'public'` or `namespaces` set to `['public']`, but not both :thinking:

greg0ire:
Do any of your ORM mapping files/configuration mention `public`?

VictorKrasnov:
No, I can't find anything.

greg0ire:
Well what you can do is start a brand new project with all the latest libraries, and using Postgres, and test that you don't have the issue, since people at https://github.com/doctrine/migrations/issues/1415#issuecomment-2449682796 say it works for them.

If you don't reproduce the issue, use your debugger on the new project to see how exactly the variables differ from what you have in your project, in particular `$toSchema`.

You could also start from https://github.com/greg0ire/sample-pg-app

VictorKrasnov:
OK, thanks for your help. I'll try to do that a little later.

derrabus:
Can we please find a meaningful title for this PR? Thanks.

VictorKrasnov:
> Can we please find a meaningful title for this PR? Thanks.

Done.

src/Generator/DiffGenerator.php
```diff
@@ -70,7 +70,7 @@ static function ($assetName) use ($filterExpression) {
             ! method_exists($this->schemaManager, 'getSchemaSearchPaths')
             && $this->platform->supportsSchemas()
         ) {
-            $defaultNamespace = $toSchema->getName();
+            $defaultNamespace = $toSchema->getNamespaceName() ?? '';
```
morozov, line 73: This doesn't look right. The schema _is_ the namespace for its tables. The schema itself cannot be associated with a namespace (only tables, views and sequences can). Currently it's possible to call `getNamespaceName()` on any `AbstractAsset` (which is a design flaw) but in DBAL 5, it will be possible only with the objects mentioned earlier.

The only situation in which currently a schema can be associated with a namespace is when its name contains a dot and the name isn't quoted (which is also a flaw and shouldn't be allowed).

MatTheCat:
FWIW I just had the same issue when upgrading a project to DBAL 4 and the cause was that a dependency registered a subscriber similar to https://github.com/doctrine/dbal/issues/1110#issuecomment-352177498.

melroy89:
As you see see we at Mbin also are facing similar issues now here: https://github.com/MbinOrg/mbin/pull/1873

MatTheCat:
@melroy89 same issue than [mentioned above](https://github.com/doctrine/migrations/pull/1476#issuecomment-2823714530): you’d need to remove your `DoctrineMigrationsListener`.

I think this PR can be closed since there doesn’t seem to be an issue anymore?

PR #1477, opened 2024-12-14
Fix test failing on Windows

srsbiz:
Removed checking PHP_OS_FAMILY in favor of using is_link on symlinked directory. 

|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Symilnks works fine under Windows, so we should not assert that symlinked files will not work in this OS family.


greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @srsbiz

PR #1456, opened 2024-08-28
Simplify handling of `--all-or-nothing` and `--no-all-or-nothing` options

agustingomes:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | yes
| Fixed issues | 
#### Summary

Allowing passing values to options for `--all-or-nothing` option is now removed, which allows the configuration of the negatable option `--no-all-or-nothing`, whihc in its turn allows to remove the added option in 3.8.1.

The outcome is a much simpler logic to override the configuration of `all_or_nothing` via the CLI command.

PR #1478, opened 2024-12-14
Introduce addDeferredSql method in migration class

srsbiz:
|      Q       |   A
|------------- | -----------
| Type         | feature
| BC Break     | no

#### Summary

This will allow to add queries that needs to be executed after changes made to schema object.

#### Reason for this change

When people need to write data in newly created table / added column, they probably resorts to calling  `$this->connection->executeStatement` in `postUp`, and this may be wrong in at least two scenarios:

1. When testing migration with `--dry-run` option, `executeStatement` is always called, since migration does not know about dry-run mode
2. Such SQLs are not visible if executed with `-vv` or `--write-sql` options, since `executeStatement` does not log queries like `addSql`.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -5,6 +5,9 @@
   to enable wrapping all migrations in a single transaction. To disable it, you can use the `--no-all-or-nothing` 
   option instead. Both options override the configuration value.
 
+## Migration classes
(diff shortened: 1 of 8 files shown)


srsbiz:
Failing check is caused by #1456

srsbiz:
@greg0ire @stof @SenseException @derrabus Can I have any comment on this PR?

Issue #1487, opened 2025-02-08
Migrations Diff and ORM Schema commands use schema_filter config differently


cristi-contiu:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.8.2

#### Summary

With PostgreSQL, the migrations diff command does not include expected create table queries for tables not in `public` schema when using `doctrine.dbal.schema_filter` config. The ORM's schema update and validate commands include these queries.

#### Current behavior

When building the diff, the [`DiffGenerator->createToSchema()`](https://github.com/doctrine/migrations/blob/3.8.x/src/Generator/DiffGenerator.php#L120) removes the schema name from the table name (in [`DiffGenerator->resolveTableName`](https://github.com/doctrine/migrations/blob/3.8.x/src/Generator/DiffGenerator.php#L147)) before passing it to the `$schemaAssetsFilter` to be matched against the regex in `schema_filter`. The ORM's schema generation keeps the schema name if it's not the [default (search path) schema](https://www.postgresql.org/docs/16/ddl-schemas.html#DDL-SCHEMAS-PATH) (usually the `public` schema), so two different names for the same table will be matched against the regex, with and without the schema name.

I would suggest aligning the way the `schema_filter` is used by not removing the schema name (if any) from the table name before sending it to the `$schemaAssetsFilter`. This can be done simply by not calling [`DiffGenerator->resolveTableName`](https://github.com/doctrine/migrations/blob/3.8.x/src/Generator/DiffGenerator.php#L147) in [`DiffGenerator->createToSchema()`](https://github.com/doctrine/migrations/blob/3.8.x/src/Generator/DiffGenerator.php#L120). I'm not sure about backwards compatibility implications on other database vendors.

[...]

PR #1488, opened 2025-02-09
Update generated diff to match ORM schema diff

cristi-contiu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no/minor
| Fixed issues | #1487

#### Summary

Ports https://github.com/doctrine/orm/pull/7875 into Migrations to improve consistency between queries generated by Migrations diff and the ones generated by the ORM schema update/validate commands. In most cases, the queries where already identical, but in some rare cases (like #1487), the migrations generated from diff did not contain the same queries as the ORM schema update and validate commands.

It also removes filtering the schema generated from metadata (`$toSchema`) using DBAL's `schemaAssetsFilter` and the `doctrine.dbal.schema_filter` config, which should only be used to exclude tables in the schema generated from the existing database (`$fromSchema` - DBAL and DoctrineBundle already cover this). This might be considered a minor BC-break, but if the metadata contains unwanted tables, they shouldn't be included in the mappings in the first place.

The decoration of the existing `schemaAssetsFilter` from the DBAL configuration is done in https://github.com/doctrine/orm/pull/7875 to whitelist assets present in the `$toSchema` so that they don't get removed from the `$fromSchema` by the default `schemaAssetsFilter` (usually based on the `doctrine.dbal.schema_filter` config) - a difference which would lead to always generating `CREATE TABLE` queries.

I think this last part is best tested in a functional/integration test, but I would need some guidance on how to set it up.

If it's a wanted change, I can also add a note to `UPGRADE.md`, depending on the branch it would be included in and if it's considered a minor BC break.

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -35,7 +35,7 @@ parameters:
             path: src/DependencyFactory.php
 
         -
-            message: '~^Strict comparison using !== between callable\(\)\: mixed and null will always evaluate to true\.$~'
+            message: '~^Strict comparison using === between callable\(\)\: mixed and null will always evaluate to false\.$~'
             path: src/Generator/DiffGenerator.php
 
         -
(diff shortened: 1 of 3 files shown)


cristi-contiu:
Accidentaly closed this PR, opened a new one here: https://github.com/doctrine/migrations/pull/1490

PR #1489, opened 2025-02-09
PHPStan: Remove ignored XML variable property access error that is no longer matched

cristi-contiu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | 

#### Summary

PHPStan fails with the following message:
> Error: Ignored error pattern `~^Variable property access on SimpleXMLElement\.$~` in path /home/runner/work/migrations/migrations/src/Configuration/Migration/XmlFile.php was not matched in reported errors.

Could be that a newer version of PHPStan or a dependency no longer triggers this error.


greg0ire (approved):

greg0ire:
Thanks @cristi-contiu !

PR #1490, opened 2025-02-14
DiffGenerator: fix applying regex pattern filter to mappings schema

cristi-contiu:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no/minor
| Fixed issues | #1487

#### Summary

DiffGenerator [applies regex pattern filtering to the metadata-based schema](https://github.com/doctrine/migrations/blob/3.8.x/src/Generator/DiffGenerator.php#L124) (`$toSchema`) using DBAL's `schemaAssetsFilter` (and DoctrineBundle's `doctrine.dbal.schema_filter`, if applicable), which was meant for filtering the introspected schema (`$fromSchema` ) to prevent dropping existing tables from the database  (DBAL and DoctrineBundle already cover this). This is also [explained in the docs](https://www.doctrine-project.org/projects/doctrine-migrations/en/3.8/reference/generating-migrations.html#ignoring-custom-tables) : "custom tables which are not managed by Doctrine" = not in the Doctrine mappings, so if an asset is in the metadata ("managed by Doctrine"), it should be included in both `$toSchema` and `$fromSchema` (if it exists), regardless of `doctrine.dbal.schema_filter` config.

This extra filtering also generates differences in the diff created by Migrations vs the ones created by [ORM's SchemaTool](https://github.com/doctrine/orm/blob/3.3.x/src/Tools/SchemaTool.php#L887), as explained in #1487 . 

The logic not to exclude tables in the introspected schema that exist in the mappings was ported from https://github.com/doctrine/orm/pull/7875 and prevents adding unwanted "CREATE TABLE" queries.

### DIFF
--- a/src/Generator/DiffGenerator.php
+++ b/src/Generator/DiffGenerator.php
@@ -14,8 +14,6 @@ use Doctrine\Migrations\Provider\SchemaProvider;
 
 use function method_exists;
 use function preg_match;
-use function strpos;
-use function substr;
 
 /**
  * The DiffGenerator class is responsible for comparing two Doctrine\DBAL\Schema\Schema instances and generating a
@@ -46,14 +44,22 @@ class DiffGenerator
         bool $checkDbPlatform = true,
         bool $fromEmptySchema = false,
     ): string {
+        $toSchema = $this->createToSchema();
+
         if ($filterExpression !== null) {
+            // whitelist assets we already know about in $toSchema, use the existing $filterExpression otherwise
+            // @see https://github.com/doctrine/orm/pull/7875
             $this->dbalConfiguration->setSchemaAssetsFilter(
-                static function ($assetName) use ($filterExpression) {
+                static function ($assetName) use ($filterExpression, $toSchema): bool {
                     if ($assetName instanceof AbstractAsset) {
                         $assetName = $assetName->getName();
                     }
 
-                    return preg_match($filterExpression, $assetName);
+                    if ($toSchema->hasTable($assetName) || $toSchema->hasSequence($assetName)) {
+                        return true;
+                    }
+
+                    return (bool) preg_match($filterExpression, $assetName);
                 },
             );
         }
@@ -61,8 +67,6 @@ class DiffGenerator
         $fromSchema = $fromEmptySchema
             ? $this->createEmptySchema()
             : $this->createFromSchema();
-
-        $toSchema = $this->createToSchema();
 
         // prior to DBAL 4.0, the schema name was set to the first element in the search path,
         // which is not necessarily the default schema name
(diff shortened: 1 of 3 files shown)


src/Generator/DiffGenerator.php
```diff
@@ -112,42 +110,37 @@ private function createEmptySchema(): Schema
...
+     */
+    private function createFromSchema(Schema $toSchema): Schema
     {
-        $toSchema = $this->schemaProvider->createSchema();
+        // backup schema assets filter
+        $previousFilter = $this->dbalConfiguration->getSchemaAssetsFilter();
 
-        $schemaAssetsFilter = $this->dbalConfiguration->getSchemaAssetsFilter();
+        if ($previousFilter === null) {
+            return $this->schemaManager->introspectSchema();
```
cristi-contiu: This is kept for compatibility with DBAL v3. In DBAL v4, it can no longer be `null`. Might trigger a codecov warning

src/Generator/DiffGenerator.php
```diff
@@ -53,16 +51,16 @@ static function ($assetName) use ($filterExpression) {
                         $assetName = $assetName->getName();
                     }
 
-                    return preg_match($filterExpression, $assetName);
+                    return (bool) preg_match($filterExpression, $assetName);
```
cristi-contiu, line 62: Casting to `bool` for [docs ](https://www.doctrine-project.org/projects/doctrine-migrations/en/3.8/reference/generating-migrations.html#ignoring-custom-tables) and [DoctrineBundle](https://github.com/doctrine/DoctrineBundle/blob/2.13.x/src/Dbal/RegexSchemaAssetFilter.php#L25) alignment

greg0ire (commented):
Can you please add a failing test?

cristi-contiu:
@greg0ire My use case is the DiffCommand being called without an explicit `--filter-expression` , but the DBAL already having a SchemaAssetsFilter callback configured (by DoctrineBundle and the `doctrine.dbal.schema_filter` config).

[Failing test](https://github.com/doctrine/migrations/pull/1492/commits/a76ce4df2d66e846d6c99b17074e44792d52eda2) and [pipeline](https://github.com/doctrine/migrations/actions/runs/13582755552/job/37971578875)

There are differences in the way Migrations and DBAL apply the filter: DBAL applies it during schema introspection on what it thinks it's the most accurate table name using ["portableTableDefinition"](https://github.com/doctrine/dbal/blob/3.9.x/src/Schema/AbstractSchemaManager.php#L340) which has platform-specific checks like schemas support or current / default search-path schemas (for example `some_schema.my_table` on PostgreSQL), while Migrations removes the schema and applies the filter only on the unqualified table name (for the same table as earlier, it would be only `my_table`). 

The Migrations library should at least apply the regex filtering on the exact table name provided by the user in the mappings, it should not assume that `some_schema.my_table` and `my_table` reference the same table . I [suggested a fix](https://github.com/doctrine/migrations/pull/1492/commits/e48897d9086918b65ab5ad73d827646b3afb8a0b) that keeps the schema name for platforms that support schemas.

cristi-contiu:
Then there is the broader issue (addressed in this PR): if the DiffGenerator should manipulate/filter the mapped `$toSchema` at all (the ORM SchemaTool does not).

The metadata schema does not have the advanced platform-specific checks that the introspected schema has, and any apparent differences are addressed by the Comparator during diff.

For example, a mapped table with name `public.my_table` OR a table with both name `my_table` and explicit schema/namespace `public` OR a table just with name `my_table` can reference the same table in PostgreSQL, but filtering just by the provided name as string without platform-specific knowledge would produce mixed results. A safer approach would be to remove that table from the mappings instead on relying on Migrations library to filter it out.

greg0ire:
I don't know why there is a filter at the migrations level. @goetas , maybe you know?

Issue #1458, opened 2024-09-03
Error with transaction nesting and rollback and problems with error handling when running a migration


savemetenminutes:
### Bug Report

|    Q            |   A
|------------ | ------
| BC Break   | no
| Version     | 3.8.1

#### Summary

```
In ExceptionConverter.php line 118:

  [Doctrine\DBAL\Exception\DriverException (1305)]
  An exception occurred while executing a query: SQLSTATE[42000]: Syntax error or access violation: 1305 SAVEPOINT DOCTRINE_2 does not exist


Exception trace:
  at /var/www/content/adex-sales/vendor/doctrine/dbal/src/Driver/API/MySQL/ExceptionConverter.php:118
 Doctrine\DBAL\Driver\API\MySQL\ExceptionConverter->convert() at /var/www/content/adex-sales/vendor/doctrine/dbal/src/Connection.php:1939
 Doctrine\DBAL\Connection->handleDriverException() at /var/www/content/adex-sales/vendor/doctrine/dbal/src/Connection.php:1881
 Doctrine\DBAL\Connection->convertExceptionDuringQuery() at /var/www/content/adex-sales/vendor/doctrine/dbal/src/Connection.php:1213
 Doctrine\DBAL\Connection->executeStatement() at /var/www/content/adex-sales/vendor/doctrine/dbal/src/Connection.php:1634
 Doctrine\DBAL\Connection->rollbackSavepoint() at /var/www/content/adex-sales/vendor/doctrine/dbal/src/Connection.php:1531
 Doctrine\DBAL\Connection->rollBack() at /var/www/content/adex-sales/vendor/doctrine/migrations/src/Tools/TransactionHelper.php:56
 Doctrine\Migrations\Tools\TransactionHelper::rollbackIfInTransaction() at /var/www/content/adex-sales/vendor/doctrine/migrations/src/DbalMigrator.php:59
 Doctrine\Migrations\DbalMigrator->executeMigrations() at /var/www/content/adex-sales/vendor/doctrine/migrations/src/DbalMigrator.php:134
 Doctrine\Migrations\DbalMigrator->migrate() at /var/www/content/adex-sales/vendor/doctrine/migrations/src/Tools/Console/Command/MigrateCommand.php:225
 Doctrine\Migrations\Tools\Console\Command\MigrateCommand->execute() at /var/www/content/adex-sales/vendor/symfony/console/Command/Command.php:298
[...]

PR #1503, opened 2025-05-29
https://github.com/doctrine/migrations/issues/1458

savemetenminutes:
https://github.com/doctrine/migrations/issues/1458


greg0ire:
I think the thing that actually needs to be refactor is https://github.com/doctrine/migrations/blob/dfaf1bc2e829ef093f5a9afce3b3a447cf7f3d6e/src/Version/DbalExecutor.php#L71-L89

You should do something similar to https://github.com/doctrine/orm/pull/11646

PR #1504, opened 2025-06-01
Add nowdoc option, by default activated for formatted output

vinise:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     |no

#### Summary

Improvement following [PR-1473](https://github.com/doctrine/migrations/pull/1473) adding nowdoc by default for formated output only, add an option --nowdoc for full nowdoc output

### DIFF
--- a/src/Tools/Console/Command/DiffCommand.php
+++ b/src/Tools/Console/Command/DiffCommand.php
@@ -64,6 +64,12 @@ EOT)
                 'Format the generated SQL.',
             )
             ->addOption(
+                'nowdoc',
+                null,
+                InputOption::VALUE_NEGATABLE,
+                'Output the generated SQL as a nowdoc string (always active for formatted queries).',
+            )
+            ->addOption(
                 'line-length',
                 null,
                 InputOption::VALUE_REQUIRED,
@@ -102,6 +108,7 @@ EOT)
         }
 
         $formatted       = filter_var($input->getOption('formatted'), FILTER_VALIDATE_BOOLEAN);
+        $nowdocOutput    = filter_var($input->getOption('nowdoc'), FILTER_VALIDATE_BOOLEAN);
         $lineLength      = (int) $input->getOption('line-length');
         $allowEmptyDiff  = $input->getOption('allow-empty-diff');
         $checkDbPlatform = filter_var($input->getOption('check-database-platform'), FILTER_VALIDATE_BOOLEAN);
@@ -135,6 +142,7 @@ EOT)
                 $fqcn,
                 $filterExpression,
                 $formatted,
+                $nowdocOutput,
                 $lineLength,
                 $checkDbPlatform,
                 $fromEmptySchema,
--- a/src/Tools/Console/Command/DumpSchemaCommand.php
+++ b/src/Tools/Console/Command/DumpSchemaCommand.php
@@ -14,8 +14,11 @@ use Symfony\Component\Console\Output\OutputInterface;
 
 use function addslashes;
 use function class_exists;
+use function filter_var;
 use function sprintf;
 use function str_contains;
+
+use const FILTER_VALIDATE_BOOLEAN;
 
 /**
  * The DumpSchemaCommand class is responsible for dumping your current database schema to a migration class. This is
@@ -50,6 +53,12 @@ EOT)
                 'Format the generated SQL.',
             )
             ->addOption(
+                'nowdoc',
+                null,
+                InputOption::VALUE_NONE,
+                'Output the generated SQL as a nowdoc string (always active for formatted queries).',
+            )
+            ->addOption(
                 'namespace',
                 null,
                 InputOption::VALUE_REQUIRED,
@@ -75,8 +84,9 @@ EOT)
         InputInterface $input,
         OutputInterface $output,
     ): int {
-        $formatted  = $input->getOption('formatted');
-        $lineLength = (int) $input->getOption('line-length');
+        $formatted    = filter_var($input->getOption('formatted'), FILTER_VALIDATE_BOOLEAN);
+        $nowdocOutput = filter_var($input->getOption('nowdoc'), FILTER_VALIDATE_BOOLEAN);
+        $lineLength   = (int) $input->getOption('line-length');
 
         $schemaDumper = $this->getDependencyFactory()->getSchemaDumper();
 
@@ -98,6 +108,7 @@ EOT)
             $fqcn,
             $input->getOption('filter-tables'),
             $formatted,
+            $nowdocOutput,
             $lineLength,
         );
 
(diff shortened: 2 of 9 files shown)


src/Tools/Console/Command/DiffCommand.php
```diff
@@ -63,6 +63,12 @@ protected function configure(): void
                 InputOption::VALUE_NONE,
                 'Format the generated SQL.',
             )
+            ->addOption(
+                'nowdoc',
+                null,
+                InputOption::VALUE_NONE,
+                'Output the generated SQL as a nowdoc string (always active for formated queries).',
```
andrew-demb:
```suggestion
                'Output the generated SQL as a nowdoc string (always active for formatted queries).',
```

src/Tools/Console/Command/DumpSchemaCommand.php
```diff
@@ -49,6 +52,12 @@ protected function configure(): void
                 InputOption::VALUE_NONE,
                 'Format the generated SQL.',
             )
+            ->addOption(
+                'nowdoc',
+                null,
+                InputOption::VALUE_NONE,
+                'Output the generated SQL as a nowdoc string (always active for formated queries).',
```
andrew-demb:
```suggestion
                'Output the generated SQL as a nowdoc string (always active for formatted queries).',
```

src/Tools/Console/Command/DiffCommand.php
vinise: Thanks

tests/Generator/SqlGeneratorTest.php
```diff
@@ -51,24 +51,37 @@ public function testGenerate(): void
...
+    {
+        $this->configuration->setCheckDatabasePlatform(true);
+
+        $expectedCode = $this->prepareGeneratedCode(
+            <<<'CODE'
+                $this->addSql('SELECT 1');
+                $this->addSql('SELECT 2');
+                $this->addSql('%s');
+                CODE,
+            formated: false,
```
greg0ire: Use search and replace please.

src/Tools/Console/Command/DiffCommand.php
```diff
@@ -63,6 +63,12 @@ protected function configure(): void
                 InputOption::VALUE_NONE,
                 'Format the generated SQL.',
             )
+            ->addOption(
+                'nowdoc',
+                null,
+                InputOption::VALUE_NONE,
```
greg0ire: How about providing `no-nowdoc` as well?

```suggestion
                InputOption::VALUE_NEGATABLE,
```

tests/Generator/SqlGeneratorTest.php
```diff
@@ -51,24 +51,37 @@ public function testGenerate(): void
...
         );
 
-        $code = $migrationSqlGenerator->generate($this->sql, true, 80);
+        $code = $migrationSqlGenerator->generate($this->sql, true, false, 80);
 
         self::assertSame($expectedCode, $code);
     }
 
-    public function testGenerationWithoutCheckingDatabasePlatform(): void
+    public function testGenerationWithoutFormating(): void
```
greg0ire:
```suggestion
    public function testGenerationWithoutFormatting(): void
```

src/Tools/Console/Command/DiffCommand.php
vinise: Why not.
What about formatted output ? do you prefer to always keep it as nowdoc no mater this option or not?

greg0ire: I'm not sure why exactly one would want to disable nowdoc in the first place. Can you elaborate? You mentioned that it's "sad" that it cannot be disabled, but why is that exactly? Depending on your answer, it might be easier to make a decision.

greg0ire: I think `--formatted` should imply `--nowdoc`, but that if we're going to provide a means to disable nowdoc, it would be weird to ignore it.

vinise: @greg0ire even with formatted I don't think imply `--nowdoc` is a great idea for all query
I would adopt the same way of working of `--formatted` which is limited by `--line-length`

I no formatting is applied, this will result in a one-line query that provides no value for nowdoc.
As a quick example as well, adding one "legacy_id" column to 25 tables will generate 25 short querry, i think it should stay on one line, avoiding to have 75 lines (times 2 for up and down) that don't bring anything 

To resume what I propose: 
- By default, no nowdoc
- If `--formatted` enabled, all formatted output will be by default in nowdoc (regarding `--line-length` option)
- if `--nowdoc` enabled, all output will be in nowdoc
- if `--no-nowdoc` enabled, all output will be as previously formatted without nowdoc, event formatted ones

If we agree on this behavior, I'll implement it like this.

theofidry: To clarify, what is the difference between formatted and nowdoc then?

vinise: @theofidry
- default

```php
$this->addSql('CREATE TABLE "user" (id SERIAL PRIMARY KEY, firstname VARCHAR(100) NOT NULL, lastname VARCHAR(100) NOT NULL )');
$this->addSql('ALTER TABLE platform ADD test_number INT NOT NULL');
```
- `--nowdoc`
```php
  $this->addSql(<<<'SQL'
      CREATE TABLE "user" (id SERIAL PRIMARY KEY, firstname VARCHAR(100) NOT NULL, lastname VARCHAR(100) NOT NULL)
  SQL);
  $this->addSql(<<<'SQL'
      ALTER TABLE platform ADD test_number INT NOT NULL
  SQL);
  ```

- `--formatted`
```php
$this->addSql(<<<'SQL'
  CREATE TABLE "user" (
    id SERIAL PRIMARY KEY,
    firstname VARCHAR(100) NOT NULL,
    lastname VARCHAR(100) NOT NULL
  )
SQL);
$this->addSql('ALTER TABLE platform ADD test_number INT NOT NULL');
```

- `--formatted --nowdoc`
```php
$this->addSql(<<<'SQL'
  CREATE TABLE "user" (
    id SERIAL PRIMARY KEY,
    firstname VARCHAR(100) NOT NULL,
    lastname VARCHAR(100) NOT NULL
  )
SQL);
$this->addSql(<<<'SQL'
    ALTER TABLE platform ADD test_number INT NOT NULL
SQL);
```
- `--formatted --no-nowdoc`
```php
$this->addSql('
  CREATE TABLE "user" (
    id SERIAL PRIMARY KEY,
    firstname VARCHAR(100) NOT NULL,
    lastname VARCHAR(100) NOT NULL
  )
');
$this->addSql('ALTER TABLE platform ADD test_number INT NOT NULL');
```

greg0ire: I think I'm OK with this proposal.

greg0ire:
Please rebase rather than merge, and take this as an occasion to squash some of the commits (e.g. the "Fix typo" commits, we probably don't need to know about those when reviewing the history).

greg0ire (approved):

SenseException (approved):

greg0ire:
Thanks @vinise !

PR #1514, opened 2025-06-30
Use native lazy objects when possible

greg0ire:


src/Provider/LazySchemaDiffProvider.php
```diff
@@ -34,6 +45,16 @@ public function createToSchema(Schema $fromSchema): Schema
             return LazySchema::createLazyProxy(static fn () => $originalSchemaManipulator->createToSchema($fromSchema));
         }
 
+        if (PHP_VERSION_ID >= 80400) {
+            $reflector = new ReflectionClass(Schema::class);
+
+            if ($reflector->isUninitializedLazyObject($fromSchema)) {
+                return $reflector->newLazyProxy(
+                    static fn () => $originalSchemaManipulator->createToSchema($fromSchema),
```
greg0ire: PHP complains that this object is lazy, and I do not understand why. I don't think it is.

Another thing I do not understand: why is there a logical or here: https://github.com/php/php-src/blob/4a98b36416a8ebe0964d2e6ceca435c0d7f3f065/Zend/zend_lazy_objects.c#L517

greg0ire: OK actually, it was, although `isUninitializedObject()` says the oppositte, because it's a clone: https://github.com/doctrine/migrations/blob/dfc28aea72cc28da34dafdd4c9e3e24abd703947/src/Provider/DBALSchemaDiffProvider.php#L35-L38

So now I just don't understand why `isUninitializedObject()` said the opposite, but anyway, there is no need to check, because `initializeLazyObject()` can be used on non-lazy objects (it's a no-op).

greg0ire: OK, it makes a bit more sense now: https://3v4l.org/ZqQpZ#v8.4.9

a proxy can still be an initialized lazy object, and be lazy (because it still is a proxy?)

SenseException (approved):

PR #1515, opened 2025-06-30
Fix command options mode (`InputOption::VALUE_REQUIRED`)

gharlan:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | 

#### Summary

These options (when used) always require a value.

### DIFF
--- a/src/Tools/Console/Command/DumpSchemaCommand.php
+++ b/src/Tools/Console/Command/DumpSchemaCommand.php
@@ -73,7 +73,7 @@ EOT)
             ->addOption(
                 'line-length',
                 null,
-                InputOption::VALUE_OPTIONAL,
+                InputOption::VALUE_REQUIRED,
                 'Max line length of unformatted lines.',
                 '120',
             );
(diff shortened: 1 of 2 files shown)


greg0ire (approved):

greg0ire:
Thanks @gharlan !

JoniJnm:
I think this change is wrong, when `--all` is specified, the `--range--from` and `--range-to` should not be set, so they should not be required

https://github.com/doctrine/migrations/blob/1b88fcb812f2cd6e77c83d16db60e3cf1e35c66c/src/Tools/Console/Command/VersionCommand.php#L137-L141

greg0ire:
@JoniJnm do you think this just by reading the code or did you experience an actual bug? Because this makes the value required. Options still stay options.

JoniJnm:
After upgrading from 3.9.0 to 3.9.4 my command fails. I run:

```
doctrine:migrations:version --em=app --add --all
```

the error is:

```
The "--range-from" option requires a value.
```

That's why I was checking the code, to know if I did something wrong or it's a breaking change

(I need that command in some integration tests)

gharlan:
That error should only happen if you use `--range-from` without a value. `--range-from` requires a value (`--range-from=...`). But only if you use the option. Options are always optional.

JoniJnm:
ok, sorry, I'm was using the wrapper [avaibooksports/doctrine-migrations-multiple-database-bundle](https://github.com/AvaiBookSports/DoctrineMigrationsMultipleDatabaseBundle)

PR #1516, opened 2025-07-02
Add --single-block option to generate a single nowdoc sql

vinise:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no

#### Summary

Add --singl-block option to generate single nowdoc sql 
`#doctrine:migrations:diff --single-block`

```php
{
    $this->addSql(<<<'SQL'
        CREATE TABLE "user" (
          id SERIAL PRIMARY KEY,
          firstname VARCHAR(100) NOT NULL,
          lastname VARCHAR(100) NOT NULL
        );
        ALTER TABLE emails ADD test_number INT NOT NULL;
        ALTER TABLE platform ADD test_number INT NOT NULL;
    SQL);
}
```

### DIFF
--- a/src/Generator/SqlGenerator.php
+++ b/src/Generator/SqlGenerator.php
@@ -42,6 +42,7 @@ class SqlGenerator
         array $sql,
         bool $formatted = false,
         bool|null $nowdocOutput = null,
+        bool $singleBlock = false,
         int $lineLength = 120,
         bool $checkDbPlatform = true,
     ): string {
@@ -61,7 +62,9 @@ class SqlGenerator
                 $query = $this->formatQuery($query);
             }
 
-            if ($nowdocOutput === true || ($nowdocOutput !== false && $formatted && strlen($query) > $maxLength )) {
+            if ($singleBlock === true) {
+                $code[] = $query;
+            } elseif ($nowdocOutput === true || ($nowdocOutput !== false && $formatted && strlen($query) > $maxLength )) {
                 $code[] = sprintf(
                     "\$this->addSql(<<<'SQL'\n%s\nSQL);",
                     preg_replace('/^/m', str_repeat(' ', 4), $query),
@@ -88,6 +91,13 @@ PHP
                     $currentPlatform,
                 ),
                 '',
+            );
+        }
+
+        if ($singleBlock === true) {
+            return sprintf(
+                "\$this->addSql(<<<'SQL'\n%s;\nSQL);",
+                preg_replace('/^/m', str_repeat(' ', 4), implode(";\n", $code)),
             );
         }
 
--- a/src/Tools/Console/Command/DiffCommand.php
+++ b/src/Tools/Console/Command/DiffCommand.php
@@ -70,6 +70,12 @@ EOT)
                 'Output the generated SQL as a nowdoc string (always active for formatted queries).',
             )
             ->addOption(
+                'single-block',
+                null,
+                InputOption::VALUE_NONE,
+                'Output the generated SQL as a single block nowdoc string.',
+            )
+            ->addOption(
                 'line-length',
                 null,
                 InputOption::VALUE_REQUIRED,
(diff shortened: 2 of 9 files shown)


src/Tools/Console/Command/DiffCommand.php
```diff
@@ -69,6 +69,12 @@ protected function configure(): void
                 InputOption::VALUE_NEGATABLE,
                 'Output the generated SQL as a nowdoc string (always active for formatted queries).',
             )
+            ->addOption(
+                'single-block',
+                null,
+                InputOption::VALUE_NONE,
```
greg0ire, line 75: Maybe it's a good thing to use `VALUE_NEGATABLE` here as well?

tests/Generator/DiffGeneratorTest.php
```diff
@@ -102,7 +102,7 @@ public function testGenerate(): void
             ->with(self::logicalOr(
                 self::equalTo(['UPDATE table SET value = 2']),
                 self::equalTo(['UPDATE table SET value = 1']),
-            ), true, false, 80)
+            ), true, false, false, 80)
```
greg0ire: Maybe we should start using named arguments, it's becoming hard to understand what's going on here.

vinise: I made change for any `DiffGenerator::generate` & `SqlGenerator::generate`.  
I will rethink both class arguments with configuration file like `MigratorConfiguration`, I really think this could improve readability.

src/Tools/Console/Command/DiffCommand.php
vinise: Negatable on this option will do nothing, and nothing to be done.
Will be added on separate MR with default formatting options in config were it'll make more sense

tests/Generator/DiffGeneratorTest.php
```diff
@@ -99,10 +99,16 @@ public function testGenerate(): void
...
-            ->with(self::logicalOr(
-                self::equalTo(['UPDATE table SET value = 2']),
-                self::equalTo(['UPDATE table SET value = 1']),
-            ), true, false, 80)
+            ->with(
+                fqcn: self::logicalOr(
+                    self::equalTo(['UPDATE table SET value = 2']),
+                    self::equalTo(['UPDATE table SET value = 1']),
+                ),
+                formated: true,
```
greg0ire: `formated` should be come `formatted`

tests/Generator/DiffGeneratorTest.php
```diff
@@ -99,10 +99,16 @@ public function testGenerate(): void
 
         $this->migrationSqlGenerator->expects(self::exactly(2))
             ->method('generate')
-            ->with(self::logicalOr(
-                self::equalTo(['UPDATE table SET value = 2']),
-                self::equalTo(['UPDATE table SET value = 1']),
-            ), true, false, 80)
+            ->with(
+                fqcn: self::logicalOr(
```
greg0ire, line 103: Looks like PHPStan won't let you use named arguments with this method.

vinise: What do you prefer then ? Revert this naming ? Baseline PHPStan for these 2 classes?

greg0ire: Ignore it and add a link to https://github.com/sebastianbergmann/phpunit/issues/6163

greg0ire: No need to address this, I fixed it upstream: https://github.com/sebastianbergmann/phpunit/pull/6257

src/Generator/SqlGenerator.php
```diff
@@ -61,7 +62,9 @@ public function generate(
                 $query = $this->formatQuery($query);
             }
 
-            if ($nowdocOutput === true || ($nowdocOutput !== false && $formatted && strlen($query) > $maxLength )) {
+            if ($singleBlock === true) {
```
greg0ire, line 65: `singleBlock` is typed as a boolean, you do not need to compare it with another boolean.
```suggestion
            if ($singleBlock) {
```

greg0ire:
Does `addSql` accept multi-query strings and handle them properly?

PR #1519, opened 2025-07-15
Resolve deprecations introduced with DBAL 4.3

mbabker:
|      Q       |   A
|------------- | -----------
| Type         | improvement
| BC Break     | no
| Fixed issues | N/A

#### Summary

Resolves deprecations reported when running with DBAL 4.3.

### DIFF
--- a/src/Metadata/Storage/TableMetadataStorage.php
+++ b/src/Metadata/Storage/TableMetadataStorage.php
@@ -9,6 +9,9 @@ use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Doctrine\DBAL\Schema\AbstractSchemaManager;
+use Doctrine\DBAL\Schema\ComparatorConfig;
+use Doctrine\DBAL\Schema\Name\UnqualifiedName;
+use Doctrine\DBAL\Schema\PrimaryKeyConstraint;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Schema\TableDiff;
 use Doctrine\DBAL\Types\Types;
@@ -25,6 +28,7 @@ use Doctrine\Migrations\Version\Version;
 use InvalidArgumentException;
 
 use function array_change_key_case;
+use function class_exists;
 use function floatval;
 use function round;
 use function sprintf;
@@ -112,7 +116,7 @@ final class TableMetadataStorage implements MetadataStorage
         $this->connection->executeStatement(
             sprintf(
                 'DELETE FROM %s WHERE 1 = 1',
-                $this->platform->quoteIdentifier($this->configuration->getTableName()),
+                $this->configuration->getTableName(),
             ),
         );
     }
@@ -196,8 +200,14 @@ final class TableMetadataStorage implements MetadataStorage
             return null;
         }
 
+        if (class_exists(ComparatorConfig::class)) {
+            $comparator = $this->schemaManager->createComparator((new ComparatorConfig())->withReportModifiedIndexes(false));
+        } else {
+            $comparator = $this->schemaManager->createComparator();
+        }
+
         $currentTable = $this->schemaManager->introspectTable($this->configuration->getTableName());
-        $diff         = $this->schemaManager->createComparator()->compareTables($currentTable, $expectedTable);
+        $diff         = $comparator->compareTables($currentTable, $expectedTable);
 
         return $diff->isEmpty() ? null : $diff;
     }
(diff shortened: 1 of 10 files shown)


mbabker:
> Call to deprecated method getName() of class Doctrine\DBAL\Schema\Schema.

So that's the one remaining case I haven't quite worked out myself yet because the deprecation implies "stop calling this and the info it gives won't be public API in 5.0", and I'm not quite sure what the way forward for this case is.

greg0ire:
@morozov hi! Can you maybe help with this?

morozov:
I will take a look in the next couple of days.

src/Metadata/Storage/TableMetadataStorage.php
```diff
@@ -109,10 +118,20 @@ public function reset(): void
     {
         $this->checkInitialization();
 
+        $tableName = $this->configuration->getTableName();
+
+        if (str_contains($tableName, '.')) {
+            $parts = array_map($this->platform->quoteSingleIdentifier(...), explode('.', $tableName));
```
morozov: It looks like the meaning of the table name is inconsistent across the project. Here, it's escaped and potentially corrupted (e.g. a value like `migrations."project.v1"`, which is valid for Postgres and SQL Server wouldn't work). In a lot of other cases like this, it would:
 https://github.com/doctrine/migrations/blob/cad2d77234894b5c131416c3b7dd44164b041129/src/Metadata/Storage/TableMetadataStorage.php#L144-L146

I propose to just remove the call to `Connection::quoteIdentifier()` for consistency with the rest of the code.

A proper, longer-term solution would be:
1. Parse the name in the configuration
2. Have `getTableName()` return an `OptionallyQualifiedName`
3. Use `Name::toSQL()` in the code that uses the table name in SQL

src/Metadata/Storage/TableMetadataStorage.php
```diff
@@ -240,7 +265,11 @@ private function getExpectedTable(): Table
         $schemaChangelog->addColumn($this->configuration->getExecutedAtColumnName(), 'datetime', ['notnull' => false]);
         $schemaChangelog->addColumn($this->configuration->getExecutionTimeColumnName(), 'integer', ['notnull' => false]);
 
-        $schemaChangelog->setPrimaryKey([$this->configuration->getVersionColumnName()]);
+        if (class_exists(PrimaryKeyConstraint::class)) {
+            $schemaChangelog->addPrimaryKeyConstraint(new PrimaryKeyConstraint(null, [new UnqualifiedName(Identifier::unquoted($this->configuration->getVersionColumnName()))], true));
```
morozov: The `PrimaryKeyConstraint` constructor is internal. Please use `PrimaryKeyConstraint::editor()` instead. You can also use `UnqualifiedName::unquoted()` instead of the `UnqualifiedName` constructor and `Identifier::unquoted()`.

morozov (commented):
> stop calling this and the info it gives won't be public API in 5.0

This is the right interpretation. It is internal concern of the `Schema` what namespace name to use by default to resolve unqualified names. Unfortunately, Migrations use this value as a workaround for the discrepancy between how the DBAL represents the introspected schema and how the ORM generates the schema that represents entities (see https://github.com/doctrine/migrations/pull/1463).

I think the easiest way forward is to introduce a getter for `Schema::$defaultNamespaceName`. I didn't anticipate it would be used like this when deprecated the original method.

src/Metadata/Storage/TableMetadataStorage.php
mbabker: Quoting removed, the longer term solution can be looked at later.

src/Metadata/Storage/TableMetadataStorage.php
mbabker: Much simpler than where I copied this stuff from, thanks!

mbabker:
> > stop calling this and the info it gives won't be public API in 5.0
> 
> This is the right interpretation. It is internal concern of the `Schema` what namespace name to use by default to resolve unqualified names. Unfortunately, Migrations use this value as a workaround for the discrepancy between how the DBAL represents the introspected schema and how the ORM generates the schema that represents entities (see #1463).
> 
> I think the easiest way forward is to introduce a getter for `Schema::$defaultNamespaceName`. I didn't anticipate it would be used like this when deprecated the original method.

For now I've just gone ahead and added that deprecation to the PHPStan ignores since it seems there's work to be done for this case.

greg0ire (approved):

SenseException (approved):

greg0ire:
@mbabker can you please look into the remaining failure?

mbabker:
> @mbabker can you please look into the remaining failure?

I honestly have no idea what this could be.  At first I thought it might be an issue with the PHPStan stubs, but [it doesn't fail in the playground](https://phpstan.org/r/132c6338-fa57-4502-adee-bc75fbb85b20).

greg0ire:
What's funny is that [if you replace one of the `mixed` by e.g. `a`, the error message also mentions `TKey` and `TValue`](https://phpstan.org/r/bde4e51e-23c2-4a4b-bb4c-c9484d600b00)

greg0ire:
OMG [I can reproduce the issue on the playground when I select level 8](https://phpstan.org/r/2b50516f-8662-46aa-8f03-b35f6abc5dc4)!

greg0ire:
I filed a bug report. Let us ignore this issue.

Issue #1502, opened 2025-05-04
`FrozenDependencies` thrown when trying to run twice migration commands in the same runtime


nikophil:
### BC Break Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 3.8.0

#### Summary

Hello, 

in [zenstruck/foundry](https://github.com/zenstruck/foundry), we reset the database by using migrations. We allow to configure multiple configuration files for the migrations, which will basically result in [a loop where we call `doctrine:migration:migrate`  through `doctrine/doctrine-migration-bundle`](https://github.com/zenstruck/foundry/blob/2.x/src/ORM/ResetDatabase/MigrateDatabaseResetter.php#L62-L64).

Since [this PR](https://github.com/doctrine/migrations/pull/1432), this behavior is impossible because the second call to `doctrine:migration:migrate` will say the dependencies are frozen. (see https://github.com/zenstruck/foundry/issues/889)

Maybe we could "unfroze" the `DependencyFactory` as soon as the migrations are run?

(I know [another issue](https://github.com/doctrine/migrations/issues/1440) exists regarding `FrozenDependencies` exception, but I think the problem is different here)

#### How to reproduce

just call twice the migration mechanism with different configurations in the same runtime

thank you for you help!

nikophil:
hey @greg0ire @VincentLanglet 

would you see any solution for this? 🙏 

I think this issue is the same than [@leefant's comment](https://github.com/doctrine/migrations/issues/1440#issuecomment-2730920046)

VincentLanglet:
> Maybe we could "unfroze" the `DependencyFactory` as soon as the migrations are run?

Indeed, that seems ok to me.

Main issue was to freeze the migration to report any call of `addSql` in postMethod
```
$migration->{'post' . ucfirst($direction)}($toSchema);
```

So it would be ok to unfreeze the migration at the end.

Do you want to try a PR for this ?

nikophil:
Thanks for your answer @VincentLanglet 

> Do you want to try a PR for this ?

yeah I'll do this soon(-ish).

PR #1528, opened 2025-10-03
fix: unfreeze `DependencyFactory` after command

nikophil:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | https://github.com/doctrine/migrations/issues/1502

#### Summary

When a migration command runs, the `DependencyFactory` is frozen, but never "unfrozen", so it cannot be run several times programmatically, like [we would need in Foundry](https://github.com/zenstruck/foundry/blob/2.x/src/ORM/ResetDatabase/MigrateDatabaseResetter.php?rgh-link-date=2025-05-04T19%3A20%3A31.000Z#L62-L64).

I propose to unfreeze at the end of all commands, but I'd be also OK to only do it in `doctrine:migration:migrate`


src/Tools/Console/Command/DumpSchemaCommand.php
```diff
@@ -80,7 +80,7 @@ protected function configure(): void
     }
 
     /** @throws SchemaDumpRequiresNoMigrations */
-    public function execute(
```
nikophil, line 83: not sure why this method was public

not sure this was made on purpose, but this change might be a small BC break in this case, if someone was using directly `DumpSchemaCommand::execute()` from outside. Should I handle this?

greg0ire: When I look at the commit that introduced it, I see a call in the tests. So that might have been for testability purposes. But yes, you should handle this, and in fact, you did that with line 82.-87, didn't you?

nikophil: Yes, I already fixed it, because it has been spotted by the CI as a BC break

nikophil:
I fixed phpstan failures, which are due to a new phpstan version

greg0ire:
It seems there are CI jobs failing. Please take a look at [this guide](https://www.doctrine-project.org/contribute/#dealing-with-checks-and-tools) for more on how to handle those.

Specifically [this part](https://www.doctrine-project.org/contribute/#running-checks-before-pushing)

nikophil:
cs fixed :+1:

greg0ire (approved):

greg0ire:
@goetas @VincentLanglet please review

VincentLanglet (approved):

src/Tools/Console/Command/DumpSchemaCommand.php
```diff
@@ -79,8 +79,14 @@ protected function configure(): void
...
 
+    /** keep BC */
+    public function execute(InputInterface $input, OutputInterface $output): int
+    {
+        return parent::execute($input, $output);
+    }
+
     /** @throws SchemaDumpRequiresNoMigrations */
-    public function execute(
+    public function doExecute(
```
derrabus: `protected`?

src/Tools/Console/Command/DoctrineCommand.php
```diff
@@ -119,6 +119,17 @@ protected function initialize(InputInterface $input, OutputInterface $output): v
...
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        try {
+            return $this->doExecute($input, $output);
+        } finally {
+            $this->dependencyFactory?->unfreeze();
+        }
+    }
+
+    abstract protected function doExecute(InputInterface $input, OutputInterface $output): int;
```
derrabus: Introducing a new abstract function (to a class that is not internal) is a BC break, isn't it?

nikophil: yeah you're right, I acted as this class was internal.

that's strange that `roave-backward-compatibility-check` workflow did not see it 

I've made the method concrete, and it now directly throws an exception. I think this is not a BC anymore

nikophil: hey @derrabus is this solution good enough?

nikophil:
Hey @greg0ire @derrabus any chance to have this PR merged? :pray: 

thanks!

nikophil:
friendly ping to get this PR merged :pray: @greg0ire

greg0ire (approved):

PR #1534, opened 2025-10-25
Setup docs-builder

greg0ire:
It is now possible to generate the docs locally, and they are checked in the CI as well. I had to fix a number of pre-existing issues for that check to pass.

### DIFF
--- a/src/AbstractMigration.php
+++ b/src/AbstractMigration.php
@@ -90,38 +90,22 @@ abstract class AbstractMigration
         }
     }
 
-    /**
-     * @throws MigrationException|DBALException
-     *
-     * @phpstan-ignore throws.unusedType,throws.unusedType
-     */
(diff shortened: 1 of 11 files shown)


src/AbstractMigration.php
```diff
@@ -90,38 +90,22 @@ public function skipIf(bool $condition, string $message = 'Unknown Reason'): voi
         }
     }
 
-    /**
-     * @throws MigrationException|DBALException
-     *
-     * @phpstan-ignore throws.unusedType,throws.unusedType
```
stof, line 96: This should probably be submitted as a separate PR.

stof (approved):

src/AbstractMigration.php
greg0ire: It's a separate commit already, and this is the lowest supported branch

PR #1535, opened 2025-10-26
feat(Console): Add --check option to diff command

ahmed-bhs:
|      Q       |   A
  |------------- | -----------
  | Type         | feature
  | BC Break     | no
  | Fixed issues | N/A

**Summary**

  This pull request introduces a new `--check` option to the `doctrine:migrations:diff` command to improve its usability in CI/CD environments.

  Currently, verifying that no pending migrations exist requires using fragile and complex shell scripts that check for the creation of new migration files. This approach is not ideal as it's not explicit and can be brittle. This PR proposes a cleaner, integrated solution.

  The new `--check` flag allows you to verify the database schema status without generating a migration file.

   * If the database schema is in sync with the mapping, the command will output a success message and exit with a 0 status code.
   * If the database schema is not in sync, the command will output a clear, actionable error message (suggesting the diff command to run) and exit with a 1 status code.

  This makes CI configuration much simpler, more readable, and improves the overall developer experience by providing a robust, self-documenting way to ensure schema consistency in automated workflows.

### DIFF
--- a/src/Generator/DiffGenerator.php
+++ b/src/Generator/DiffGenerator.php
@@ -37,6 +37,38 @@ class DiffGenerator
     ) {
     }
 
+    public function hasChanges(string|null $filterExpression, bool $fromEmptySchema = false): bool
+    {
+        if ($filterExpression !== null) {
+            $this->dbalConfiguration->setSchemaAssetsFilter(
+                static function ($assetName) use ($filterExpression) {
+                    if ($assetName instanceof AbstractAsset) {
+                        $assetName = $assetName->getName();
+                    }
+
+                    return preg_match($filterExpression, $assetName);
+                },
+            );
+        }
+
+        $fromSchema = $fromEmptySchema
+            ? $this->createEmptySchema()
+            : $this->createFromSchema();
+
+        $toSchema = $this->createToSchema();
+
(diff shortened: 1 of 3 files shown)


greg0ire:
> verifying that no pending migrations

In what situations is this needed? I mean, if there are migrations, you probably want to run them at some point, and you probably want to do so inside your CI, right?

thePanz:
@ahmed-bhs is your intent to check if "pending migrations exists"?
> Currently, verifying that no pending migrations exist requires using fragile and complex shell scripts that check for the creation of new migration files.

Why not using the https://github.com/doctrine/migrations/blob/3.9.x/src/Tools/Console/Command/UpToDateCommand.php command? 
It is described as:

> The UpToDateCommand class outputs if your database is up to date or if there are new migrations that need to be executed.

stof:
If the goal is to detect cases where you forgot to generate your migration after changing your mapping, this is already covered by the `doctrine:schema:validate` command.

Thre is no reason to reimplement the same command in doctrine/migrations.

ahmed-bhs:
Thank you @stof and @thePanz for your help — this is already covered by doctrine:schema:validate.

Issue #1551, opened 2025-11-24
DbalExecutor uses executeQuery for write operations, causing writes to replicas with PrimaryReadReplicaConnection


christiaan:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | latest

#### Summary

DbalExecutor in doctrine/migrations uses Connection::executeQuery for executing SQL statements, including write operations. However, PrimaryReadReplicaConnection in doctrine/dbal expects executeQuery to be used only for read operations. This can result in write queries being sent to replica nodes, causing data inconsistency.

#### Current behavior

When running migrations, DbalExecutor calls executeQuery for all SQL statements, including writes (e.g., INSERT, UPDATE, DELETE, DDL). If the connection is configured as PrimaryReadReplicaConnection, these write operations may be executed on a replica instead of the primary.

#### How to reproduce

1. Configure Doctrine DBAL with PrimaryReadReplicaConnection and at least one replica.
2. Run a migration that performs a write operation (e.g., creates a table or inserts data).
3. Observe that the write query is executed using executeQuery, which may target a replica.

#### Expected behavior

Write operations in migrations should use executeStatement to ensure they are always executed on the primary node, as per the contract of PrimaryReadReplicaConnection. Only read operations should use executeQuery.

See https://github.com/doctrine/dbal/blob/05f3985a266bb75c44e621e6a647027016061265/src/Connections/PrimaryReadReplicaConnection.php#L43

PR #1552, opened 2025-11-24
Use executeStatement for migration SQL

christiaan:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1551

#### Summary

DbalExecutor now uses Connection::executeStatement (instead of executeQuery) because executeQuery is intended for read-only operations in DBAL.

See https://github.com/doctrine/dbal/blob/05f3985a266bb75c44e621e6a647027016061265/src/Connections/PrimaryReadReplicaConnection.php#L43

### DIFF
--- a/src/Version/DbalExecutor.php
+++ b/src/Version/DbalExecutor.php
@@ -271,8 +271,8 @@ final class DbalExecutor implements Executor
             $this->outputSqlQuery($query, $configuration);
 
             $stopwatchEvent = $this->stopwatch->start('query');
-            // executeQuery() must be used here because $query might return a result set, for instance REPAIR does
-            $this->connection->executeQuery($query->getStatement(), $query->getParameters(), $query->getTypes());
+            // executeStatement() must be used here because executeQuery() is for READ operations only
+            $this->connection->executeStatement($query->getStatement(), $query->getParameters(), $query->getTypes());
             $stopwatchEvent->stop();
 
             if (! $configuration->getTimeAllQueries()) {
--- a/tests/Version/ExecutorTest.php
+++ b/tests/Version/ExecutorTest.php
@@ -99,6 +99,26 @@ class ExecutorTest extends TestCase
 
     public function testExecuteUp(): void
     {
+        $call = 0;
+        $this->connection
+            ->expects(self::exactly(2))
+            ->method('executeStatement')
+            ->willReturnCallback(function (string $statement, array $parameters = [], array $types = []) use (&$call): int {
+                $call++;
+                if ($call === 1) {
+                    self::assertSame('SELECT 1', $statement);
+                    self::assertSame([1], $parameters);
+                    self::assertSame([3], $types);
+                }
+
+                if ($call === 2) {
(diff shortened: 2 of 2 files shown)


src/Version/DbalExecutor.php
```diff
@@ -271,8 +271,8 @@ private function executeResult(MigratorConfiguration $configuration): void
             $this->outputSqlQuery($query, $configuration);
 
             $stopwatchEvent = $this->stopwatch->start('query');
-            // executeQuery() must be used here because $query might return a result set, for instance REPAIR does
```
greg0ire, line 274: Have you read https://github.com/doctrine/migrations/pull/1071#discussion_r534336117 ?

stof:
Replacing the comment explaining why `executeQuery` is used today is a no-go. This ignores the fact that we tried that in the past, had to revert it due to regressions and added this comment to explain why the code is implemented this way.

-1 for this PR as it will introduce the same regression again.

christiaan:
Ah thanks for pointing that out! I tried searching but could not find the relevant issues/code related to that rather brief comment. Doing a blame on that line also did not really help.

So if I understand it correctly `executeQuery` should be used for READ queries. But running an `OPTIMIZE` or `REPAIR` query using `executeStatement` is actually broken in DBAL and throws errors except when abusing `executeQuery`?

Some context; we had some mutations being accidentally executed on the replica and wanted to guard against that in development. (In production this is fixed by using a read only user for the replica connection.) So we added safeguards to our code that when executeQuery is being used it is only used for READ operations as prescribed by the PrimaryReadReplicaConnection. When building in that safeguard our migrations failed.

stof:
PDO (and some other DB drivers) have 2 different methods depending on the *result* of the query (more than on the permission they need):
- an API returning a result set (which needs to be managed in memory)
- an API returning a number of affected rows

Running a query returning a result set through the API that does not expect it will generally cause the DB driver to fail. On the other hand, running a query that does not have a result set through the API supporting result sets won't crash (it will return an empty result set).
As `doctrine/migrations` exposes a single `addSql` method, it must run those queries in the way that works.

For simple queries, `SELECT` (doing reads most of the time) returns a result set while `UPDATE` and `DELETE` don't.
But this is not a 100% guarantee:
- `SELECT` might require write permissions (when reading the next value of a sequence for instance)
- Postgresql supports using `UPDATE` statements in Common Table Expressions, allowing to have a query that performs some writes but _ends_ with a read query as the main query.
- some special queries require write permissions but have a result set (`REPAIR`) while being useful in DB migrations.

`PrimaryReadReplicaConnection` exposes an API that assumes that `executeQuery` can be used for read queries submitted to the replica, but it also exposes the `ensureConnectedToPrimary` method to allow handling cases where you have to issue a write query expecting a result set.

greg0ire:
> Doing a blame on that line also did not really help.

Blame is a limited tool. You should be using [the git pickaxe](https://git-scm.com/book/en/v2/Git-Tools-Searching) for this. That's how I found about #1071

PR #1564, opened 2026-02-10
Address latest PHPStan findings

greg0ire:


src/SchemaDumper.php
```diff
@@ -86,7 +86,8 @@ public function dump(
             if (method_exists($table, 'getObjectName')) {
                 $tableName = $table->getObjectName()->toSQL($this->platform);
             } else {
-                $tableName = $table->getQuotedName($this->platform);
+                /** @phpstan-ignore method.deprecated */
+                $tableName = $table->getName();
```
greg0ire, line 90: The tests fail unless I do this change, undoing what @derrabus  did in 4e633c76

I must admit it's not clear to me when I should use which method.

src/Generator/DiffGenerator.php
```diff
@@ -132,7 +143,13 @@ private function createToSchema(): Schema
 
         if ($schemaAssetsFilter !== null) {
             foreach ($toSchema->getTables() as $table) {
-                $tableName = $table->getName();
+                /** @phpstan-ignore function.alreadyNarrowedType */
+                if (method_exists($table, 'getObjectName')) {
```
morozov: I'd check for `instanceof NamedObject` instead of `method_exists()` just because it's less prone to typos.

src/Metadata/Storage/TableMetadataStorage.php
```diff
@@ -206,8 +207,15 @@ private function needsUpdate(Table $expectedTable): TableDiff|null
             $comparator = $this->schemaManager->createComparator();
         }
 
-        $currentTable = $this->schemaManager->introspectTable($this->configuration->getTableName());
-        $diff         = $comparator->compareTables($currentTable, $expectedTable);
+        /** @phpstan-ignore function.alreadyNarrowedType */
+        if (method_exists($this->schemaManager, 'introspectTableByUnquotedName')) {
+            $currentTable = $this->schemaManager->introspectTableByUnquotedName($this->configuration->getTableName());
```
morozov, line 212: This implies that the table name in the config is the literal value, not the SQL. E.g. if the previous API was expected to accept quoted names (e.g. `"Users"` on Postgres to retain the upper case), now it won't, because the quotes will be considered part of the name, not the syntax.

If this is relevant, you will need to parse the name and pass the result to `introspectTable()`. However, there's no forward-compatible version of `introspectTable()` in DBAL 4.5, because we're reusing this name in 5.0 with a different signature.

I think, the viable approach here is to detect the DBAL version and:
1. On DBAL 4, keep calling the deprecated method
2. On DBAL 5, parse the name and call the same method, which is no longer deprecated.

Alternatively, we can use `introspectTableByUnquotedName()` (as proposed in the PR) and accept the risk that it won't work with quoted names. Its support in the DBAL is such rudimentary that I believe few developers rely on it.

greg0ire: OK… I think I will go the "accept the risk route" and let people file bug reports if they actually rely on it.

src/Generator/DiffGenerator.php
greg0ire: Addressed.

morozov (approved):

src/Metadata/Storage/TableMetadataStorage.php
luzrain: Hi. There is a problem with this change after update. I use friendsofdoctrine/dbal-clickhouse and their [ClickHouseSchemaManager](https://github.com/FriendsOfDoctrine/dbal-clickhouse/blob/master/src/ClickHouseSchemaManager.php) don't implement `Doctrine\DBAL\Platforms\AbstractPlatform::createMetadataProvider() `method that lead to an exception if you try to call `introspectTableByUnquotedName` method.

May we add try catch block here? I could prepare pull request if you are agree.

greg0ire: I don't understand the issue, but I'm fairly sure we are not going to fix it with a try/catch block. Please open a dedicated issue and explain this in more detail.

PR #1567, opened 2026-02-18
Handle schema names in table metadata

### LINKED ISSUE #1566: Configuring custom database schema name for migration_versions table breaks every command in release 3.9.6
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | yes
| Version     | 3.9.6

#### Summary

Release 3.9.6 breaks configurations where _migration_versions_ table is stored in a non-default schema.

#### Current behavior

When running any migrations command an exception will be thrown if the migrations table is configured to be stored in a non-default schema.

#### How to reproduce

1. Use a database supporting schemas (in my case: SQL Server)
2. Configure the migrations table name to include the schema name (e.g. by using doctrine migrations bundle):

```yaml
doctrine_migrations:
    migrations_paths:
        # namespace is arbitrary but should be different from App\Migrations
        # as migrations classes should NOT be autoloaded
        'DoctrineMigrations': '%kernel.project_dir%/migrations'
    enable_profiler: false

    storage:
        table_storage:
            table_name: 'customschema.migration_versions'
```

3. Run any migrations command (e.g. by using symfony console).

```bash
bin/console doctrine:migrations:status

In TableDoesNotExist.php line 16:
                                                                            
  There is no table with name "customschema.migration_versions" in the schema.  
                                                                            
```
#### Expected behavior

Up until release 3.9.5 every command worked as expected. The _migration_versions_ table was stored, read and written in the defined schema.

greg0ire:
Fixes #1566

In a previous commit, I switched from `introspectTable()`, which handles this with a single argument to `introspectTableByUnquotedName()`, which handles it with 2 arguments.

I could not find a way to test this since we use SQLite and that the corresponding DBAL platform does not support schemas.

### DIFF
--- a/src/Metadata/Storage/TableMetadataStorage.php
+++ b/src/Metadata/Storage/TableMetadataStorage.php
@@ -28,11 +28,14 @@ use Doctrine\Migrations\Version\Version;
 use InvalidArgumentException;
 
 use function array_change_key_case;
+use function assert;
 use function class_exists;
+use function explode;
 use function floatval;
 use function method_exists;
 use function round;
 use function sprintf;
+use function str_contains;
 use function strlen;
 use function strpos;
 use function strtolower;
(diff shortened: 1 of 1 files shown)


SenseException:
It looks like there's no test that covers table names with a dot. Maybe you can cover your new code path with a test like `Doctrine\Migrations\Tests\Metadata\Storage\TableMetadataStorageTest::testTableStructure()`

greg0ire:
@SenseException that test relies on SQLite, which does not support namespaces.

stlrnz:
I can confirm that this PR fixes https://github.com/doctrine/migrations/issues/1566

Thank you!

SenseException:
@greg0ire It introduces a new code path with a new if-block, that I, AFAIK, don't see covered by a test. Please let me know if there's one that covers the new path elsewhere that I missed.

greg0ire:
I don't think it is currently covered because it is about schemas, and we only have SQLite available in this test suite I think, and it does not support schemas, or rather, the corresponding DBAL platform doesn't.

stlrnz:
Do you think this PR can be approved and merged anytime soon?

greg0ire:
I don't see myself setting extra CI jobs just to have an RDBMS that has schemas, and cover this codepath. I will bypass merge rules and merge without approval this time, and deal with the consequences. Sorry :bow:

Issue #1584, opened 2026-04-07
[Bug] TableMetadataStorage::needsUpdate() fails with schema-qualified table names when using doctrine/dbal 4.4.x


rafarian:
### Bug Report

|    Q        |   A
|------------ | ------
| BC Break    | no
| Version     | 3.9.6

#### Summary

`TableMetadataStorage::needsUpdate()` passes the full configured table name
(e.g. `myschema.doctrine_migration_versions`) directly to
`introspectTableByUnquotedName()`, introduced in doctrine/dbal 4.4.x.
Unlike the deprecated `introspectTable()` it replaces, the new method does **not**
parse dot notation — schema and table name must be supplied as separate arguments.
This causes a fatal error when the migrations table is configured with a schema prefix.

#### Current behavior

Running any migration command (e.g. `doctrine:migrations:migrate`) throws:

```
There is no table with name "myschema.doctrine_migration_versions" in the schema.
```

The regression was introduced in 3.9.6 in
[`TableMetadataStorage::needsUpdate()`](https://github.com/doctrine/migrations/blob/3.9.6/src/Metadata/Storage/TableMetadataStorage.php)
by switching from the deprecated `introspectTable($fullName)` to
`introspectTableByUnquotedName($fullName)`:

```php
if (method_exists($this->schemaManager, 'introspectTableByUnquotedName')) {
    // passes "myschema.doctrine_migration_versions" as a single string —
    // introspectTableByUnquotedName() treats the dot as part of the table name
    // and does NOT parse schema.table notation
    $currentTable = $this->schemaManager->introspectTableByUnquotedName(
        $this->configuration->getTableName(),
    );
}
```

The deprecated `introspectTable()` worked because platform-level query builders
(e.g. `PostgreSQLSchemaManager::buildQueryConditions()`) explicitly split
`schema.table` on the dot. The new explicit API has no such fallback.

#### How to reproduce

1. Use PostgreSQL with `doctrine/dbal ^4.4` and `doctrine/migrations 3.9.6`.
2. Configure a schema-qualified migrations table:

```yaml
# config/packages/doctrine_migrations.yaml
doctrine_migrations:
    storage:
        table_storage:
            table_name: 'myschema.doctrine_migration_versions'
```

3. Run `php bin/console doctrine:migrations:status` (or any migration command).

#### Expected behavior

`needsUpdate()` should split a schema-qualified table name before calling
`introspectTableByUnquotedName()`:

```php
[...]

PR #1585, opened 2026-04-07
Add support for schema-qualified table names in TableMetadataStorage

rafarian:
|      Q       |   A
|------------- | -----------
| Type         | bug
| BC Break     | no
| Fixed issues | #1584 

#### Summary

`TableMetadataStorage::needsUpdate()` passes the full configured table name
(e.g. `myschema.doctrine_migration_versions`) directly to
`introspectTableByUnquotedName()`, introduced in doctrine/dbal 4.4.x.
Unlike the deprecated `introspectTable()` it replaces, the new method does **not**
parse dot notation — schema and table name must be supplied as separate arguments.
This causes a fatal error when the migrations table is configured with a schema prefix.

### DIFF
--- a/src/Metadata/Storage/TableMetadataStorage.php
+++ b/src/Metadata/Storage/TableMetadataStorage.php
@@ -209,7 +209,14 @@ final class TableMetadataStorage implements MetadataStorage
 
         /** @phpstan-ignore function.alreadyNarrowedType */
         if (method_exists($this->schemaManager, 'introspectTableByUnquotedName')) {
-            $currentTable = $this->schemaManager->introspectTableByUnquotedName($this->configuration->getTableName());
+            $fullName = $this->configuration->getTableName();
+
+            if (str_contains($fullName, '.')) {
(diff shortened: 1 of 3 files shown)


rafarian:
Its another approach for #1566

src/Metadata/Storage/TableMetadataStorage.php
```diff
@@ -246,6 +249,15 @@ private function checkInitialization(): void
...
     }
 
+    private function getQualifiedTableName(): string
+    {
+        $schema = $this->configuration->getSchemaName();
+
+        return $schema !== null
+            ? $schema . '.' . $this->configuration->getTableName()
+            : $this->configuration->getTableName();
+    }
```
greg0ire, line 259: I think this method should be moved to the configuration object, and `getSchemaName()` should be removed.

greg0ire (commented):
The issue I see with this solution is that it requires configuring the schema name. It might be a good contribution, but maybe to 4.0.x 🤔

